Issue #454, opened 2014-05-15
ColorCodes: trigger words needs some work. 


jdorweiler:
In reference to #442.  Take a closeer look at the trigger words and removing 'code'.

PR #612, opened 2014-08-28
ColorCodes: treat "code" as semantic noise.

mwmiller:
We want to allow the string "code" to be used in the triggering query,
but we want it to have no effect on how the query is understood
or processed.

Addresses issue #454.

cc: @jdorweiler Any other corners in this of which I should be aware?

### DIFF
--- a/lib/DDG/Goodie/ColorCodes.pm
+++ b/lib/DDG/Goodie/ColorCodes.pm
@@ -62,6 +62,9 @@ sub percentify {
     return @out;
 }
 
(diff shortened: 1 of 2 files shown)


amulya349:
Hello, I am a newbie. Duckduckgo seems interesting to me. I want to contribute to this project. 

I searched for `red html` and `red html code`. `red html code` gives me details of the color. Is this issue is related to the problem that `red html` should also give the same result ?

PR #781, opened 2014-11-23
Conversions: Editable fields

### LINKED ISSUE #745: Editable conversions
https://duck.co/forum/thread/11007/editable-instant-answers

------
IA Page: https://duck.co/ia/view/conversions

killerfish:
fixes #745  (eventually).
@chrismorast i need some design pointers for this one, SS below
input
.ddgsi-chev-down isnt working as expected? Went with the flat buttons, they seemed to suit better than those in the styleguide, what do you think how we should update this (color changes etc)? 
Btw, I was of the opinion that we could replicate this with currency etc, so we could use this over there as well.

//cc @moollaza @mwmiller 
@mintsoft  guys could you list me the explicit cases we should cater for this one? (like temperature, rounding digits etc)

P.s. Dont mind the mess, ill clean up when we finalize the design :smile:

### DIFF
--- a/lib/DDG/Goodie/Conversions.pm
+++ b/lib/DDG/Goodie/Conversions.pm
@@ -7,7 +7,7 @@ with 'DDG::GoodieRole::NumberStyler';
 use Math::Round qw/nearest/;
 use bignum;
 use Convert::Pluggable;
-use Data::Dumper;
+
 name                      'Conversions';
 description               'convert between various units of measurement';
 category                  'calculations';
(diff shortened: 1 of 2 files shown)


chrismorast:
I appologize for the delay on this @killerfish - working on this now.

chrismorast:
Here you go @killerfish . Here's what I've got. In looking more at this I think it's better to allow the user to change unit type rather than value (as you've indicated with the carets). The value should be editable as a field.

Let me know your thoughts or concerns, if any.

unit conversion


cc'ing @abeyang for input as well.

russellholt:
Can we wait on implementing this until custom templates are supported?

killerfish:
@chrismorast i think to make both the unit and values editable, we might have to port the conversions module to JS, i can check it out @russellholt do you think i should, or we just halt everything on this right now?

russellholt:
@killerfish  I think you should wait, I don't like replacing the structured answer with HTML and javascript.
We are working on the general case in the coming weeks.

jagtalon:
Closing this for now while waiting. Can always reopen! :)

PR #816, opened 2014-12-06
Words goodie

pavlos256:
## Introduction

Hi! This PR adds a new goodie, "Words".

Gives a list of words, optionally matching a few criteria.

It's a lot more convenient and flexible than going through the search results and repeating the query on each site.

The Enhanced North American Benchmark LExicon Updated Millennial Edition, AKA Enable2K, released in the Public Domain. The [original source](http://personal.riverusers.com/~thegrendel/)  seems to be down but there are plenty of copies around, usually named "enable2k.zip", ["enable2k.txt"](http://www.morewords.com/enable2k.txt) or ["WORD.LST"](http://www.cs.duke.edu/csed/tapestry/data/yawl/). 

Quoting the data source itself: "[...] is the most thoroughly researched, and therefore the most authoritative word list and reference available for Scrabble players [...] has become something of a de facto standard for electronic word lists used by word gamers. It has been adopted by developers and programmers for interactive on-line word game servers and word construction (anagramming) software."

Also, it's free (Public Domain).

None that I could find.

* words starting with duck
* 9-letter words like cro*rd
* 20 words
* random words with 4 letters
* 5 6-letter words ending in ay
* words like cro----rd

Word games (Scrabble, crosswords,etc) players.

Yes, I found this instant answer idea [here](https://duck.co/ideas/idea/448/wildcard-word-lookup-for-scrabble-ish-games), waiting for a developer.

None.

No problems yet. It's the first time I'm writing Perl though, so I'd appreciate a look for mistakes and inefficiencies.

IIRC, on Hacker News.

words-screenshot

## Checklist

```
[X] Added metadata and attribution information (https://duck.co/duckduckhack/metadata)
[X] Wrote test file and added to t/ directory (https://duck.co/duckduckhack/goodie_tests)
[] Verified that instant answer adheres to design guidelines (https://duck.co/duckduckhack/design_styleguide)
[X] Verified that instant answer adheres to code styleguide (https://duck.co/duckduckhack/code_styleguide)
[] Tested cross-browser compatibility
```


nilnilnil:
@pavlos256 sorry for the delay. Have you considered trying to make this a fathead? That might be a better place to start. I would scope that out first, see if it's possible / better, and get back to us here. Does that sound good?

pavlos256:
@nilnilnil I suppose I could look into it. I have one question, though, to help me better understand what you want. What is it that you consider problematic with the current approach and how it would be better with the suggested approach? Thanks!

nilnilnil:
@jdorweiler looked a bit closer here and determined that it's not reasonable to make this a Fathead either. I think the problem we're trying to solve now is how to reduce the memory footprint. The way that pops out at me is to have a separate datastore that can do the same operations the hashes do.

nilnilnil:
I like the idea of seeing how far we can push this Goodie. It's uncharted territory. I don't think we can move forward with 37MB per process though. CDB is a constant DB. It should be able to replace the hashes. Those hashes look like they're constant structures. Is that right? We could also experiment with SQLite or something like that; but I'm not sure how large that will make the distribution (which would make it a PITA to install due to size). We also have an internal DB that contains a lot of words and permutations; but I'd like to keep this opensource if at all possible.

mintsoft:
@nilnilnil looking over it, I don't see any reason why a cdb per index couldn't be used; it should be trivial to create something in `/share` which can create the cdb's from the flat file (essentially moving the indexing into this script)

pavlos256:
I think I see where the confusion is. Yes, the indexes can be moved to CDB, but the indexes are not the problem. Of all the ~40MB this uses, the indexes take less than 500KB. Here's [their contents](https://gist.github.com/pavlos256/e1334a272aab47b21cf9).

The problem is Perl. From what I read while researching the issue, Perl stores a lot of data for each value (scalar) instead of just storing its binary representation. This is why reading the whole 1.7MB word list as **one string** takes about 1.7MB, but reading it as an array of ~175K strings (words) takes so much memory.

nilnilnil:
number of words: 173528 at components/skunkworks/caine.pl line 28.
Array size: 11.165824 MB at components/skunkworks/caine.pl line 30.
Hash size: 18.07657 MB at components/skunkworks/caine.pl line 32.

The hash there is only a lookup table. It has no real values set.

mintsoft:
@nilnilnil @pavlos256 Ah! I see; can't we just forgo the array entirely and store the word data directly within the cdb instead?

pavlos256:
We could but I think that would hurt performance. Say we want 20 10-letter words. We'd have to ask CDB to give us the list of 10-letter words (about 20K words), bring them all in as a Perl array, work with them (pick at random and/or do pattern matching) and then return the result. This would have to be repeated for every query. Right?

mintsoft:
True, however I don't think it's possible to do this without a performance hit, anything being moved from RAM and onto disk will be slower. The question is how slow would that actually be? Is it actually acceptable performance & memory wise to do it that way?

pavlos256:
I guess we'll have to try it and see. I can do it when I have the time if nobody beats me to it.

jagtalon:
I wonder if we could use an API too like https://wordnik.com/ or http://www.twinword.com/ -- both of which we already use.

jagtalon:
@pavlos256 close this for now then?

pavlos256:
@jagtalon This does work. Have we concluded that it is unacceptable as it is?

jdorweiler:
@pavlos256 yeah the memory use is too high the way it is now.  Did you try out CDB?

W25:
@pavlos256 @nilnilnil Here is my version: https://github.com/duckduckgo/zeroclickinfo-goodies/pull/943
It needs around 1.3 MB of RAM for the dictionary.

pavlos256:
I haven't tried CDB yet but I will get to it; I just haven't had the time yet. Feel free to close it in the meantime if you want.

@W25 That looks interesting. How's the performance per query? Do you think it's possible to return random words by getting more matches and shuffling them?

pavlos256:
I just tried it with CDB.

* Both this and the CDB version tests where done as standalone programs. These standalone programs start at about 4 MB of memory.

* The pure Perl version adds 40MB to that [1], while the CDB version adds about 11MB.

* The CDB version is slower. Depending on the query, it may take the same time or 15 times as much.

I don't think CDB was worth it. It's pretty amazing that it runs this fast but it still takes a significant amount of memory. Maybe I should rewrite the whole thing in C. :wink: 

Whatever happens, I'm very interested in seeing if this feature will be implemented and how.

[1]: When running with duckpan inside the VM, the Perl words goodie version adds ~60MB (from 145MB -> 203 MB).

moollaza:
@pavlos256 as mentioned in the other Words PR, https://github.com/duckduckgo/zeroclickinfo-goodies/pull/943#issuecomment-71529980, we're requiring that all Goodies should be able to return within 50ms and should use minimal amounts memory.

Are you still interested in trying to refine this IA to that point? Would you like me to close this PR in the mean time?

pavlos256:
> Are you still interested in trying to refine this IA to that point? Would you like me to close this PR in the mean time?

It seems it's not possible using DDG's software stack of choice at the moment, so yeah, let's close it.

nilnilnil:
I'd like to stress that this is a cool idea. I'm sorry it didn't work out. I hope it doesn't deter you from working on other things!

PR #877, opened 2014-12-24
Update Travis to use faster container-based (Docker) infrastructure

moollaza:
Travis offers a [new, faster infrastructure](http://docs.travis-ci.com/user/workers/container-based-infrastructure/) that uses Docker. The only downside is that it doesn't allow for `sudo` access. 

This PR cleans up the dist.ini (we have duplicate commands in the dist and travis.yml file) and updates travis.yml to install our only system dependency, GNU MPFR, via a shell script, to avoid the use of `sudo`

Travis also allows for caching of folders and packages, we could potentially cache the CPAN packages but I'm not sure about that

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 ---
+sudo: false
 before_install:
   - export HARNESS_OPTIONS=j10:c HARNESS_TIMER=1
   - rm .travis.yml
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ before_install:
   - rm .travis.yml
   - git config --global user.name "Dist Zilla Plugin TravisCI"
   - git config --global user.email $HOSTNAME":not-for-mail@travis-ci.org"
+  - chmod 755 install_mpfr.sh
   - ./install_mpfr.sh
 install:
   - cpanm  --quiet  --notest --skip-installed Dist::Zilla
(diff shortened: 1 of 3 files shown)


jagtalon:
@moollaza @mintsoft 

> Goodies repo that contains the binaries of the system packages

Will get big when we add binaries for multiple platforms too!

> Actually there's a few other single-goodie repos

I think my main problem with separate repos (now that we're getting a lot) is that people don't see them. Too fragmented.

I don't think we should move the IAs that need MPFR and Guitar Chords somewhere else just because Travis Docker can't use them. I think as long as we can install it on TravisCI + Codio we shouldn't do anything.

mintsoft:
@jagtalon I know what you mean, it feels like a difficult balance to get right. 

If I (for example) created a new goodie that had a requirement on a system module etc I presume there are implications for the live environment as well as Travis and dev. Out of curiosity, how is that handled? Is there something we can adapt for duckpan/Travis?

jagtalon:
@mintsoft currently we have our own repo using Pinto although @malbin might have more insight into this problem

jagtalon:
@moollaza We might be limited to what TravisCI can do similar to https://github.com/duckduckgo/zeroclickinfo-spice/pull/1456--we can't just edit .travis.yml :/

moollaza:
Closing for now. Hopefully we fix this soon!

Issue #124, opened 2012-09-26
crontab says "at this freq." and then gives two points, not a frequency


nospampleasemam:
http://duckduckgo.com/?q=crontab+0+0+*+*+*+%2Fusr%2Fbin%2Fnice+-n+15+%2Fusr%2Flocal%2Fddg%2Fsysadmin%2Flog%2Flogrotate_ubuntu.sh

mintsoft:
Fixed by @W25 in https://github.com/duckduckgo/zeroclickinfo-goodies/pull/885

PR #882, opened 2014-12-27
Fix crontab parsing

heytitle:
REF #124


heytitle:
@mintsoft since I have been away, The #885 seems to solve this bug already. I'll close it then.

Issue #888, opened 2015-01-02
Leap year test plan returns last year


TheAdamGalloway:
Hey,
Travis CI is giving build errors because t/LeapYear.t returns 
```perl
'is it a leap year' => test_zci('2014 CE is not a leap year'),
```
So the expected result is inconsistent with the actual year that is returned.

MrChrisW:
Thanks @TheAdamGalloway

Not sure but we might want to remove the test query as we'll be changing it every year :laughing: @mintsoft will know the best course of action! [L19](https://github.com/duckduckgo/zeroclickinfo-goodies/blob/master/t/LeapYear.t#L19

On another note, this Goodie doesn't look like it's live at the moment: https://duckduckgo.com/?q=is+it+a+leap+year :frowning:

TheAdamGalloway:
I have merged the fix for this error into my pull request so hopefully it gets accepted soon.
The goodie is probably not functioning due to the error in Travis CI

TheAdamGalloway:
This has been solved.

mintsoft:
@TheAdamGalloway It's solved in that the tests aren't failing anymore, however I'd like to keep it open to encourage reviewing of https://github.com/duckduckgo/zeroclickinfo-goodies/issues/889 until it's merged if that's OK with you? If not, just close it again and I'll open a second issue.

Thanks,

TheAdamGalloway:
No that's fine, just thought it'd make sense to close it since another
issue has been raised. But I don't mind.

-Adam Galloway <http://www.adamgalloway.me/> (PGP KEY)
<http://pgp.mit.edu:11371/pks/lookup?op=get&search=0xE798AB9347B8CD1C>

<http://adamgalloway.me/>

PR #889, opened 2015-01-02
LeapYear: Adding MockTime to output

mintsoft:
Fixes to address reduced test coverage caused by: https://github.com/duckduckgo/zeroclickinfo-goodies/issues/888

@MrChrisW this is what's normally done for testing temporal things (see https://github.com/duckduckgo/zeroclickinfo-goodies/blob/master/t/CalendarToday.t#L87 for example)

However, I've just noticed that this technique absolutely hasn't worked here as the current date and time isn't actually used when a query is made, only the time when the goodie was loaded : https://github.com/duckduckgo/zeroclickinfo-goodies/blob/master/lib/DDG/Goodie/LeapYear.pm#L34

### DIFF
--- a/lib/DDG/Goodie/LeapYear.pm
+++ b/lib/DDG/Goodie/LeapYear.pm
@@ -75,8 +75,7 @@ sub format_year {
 }
 handle remainder => sub {
     
-    my ($second, $minute, $hour, $dayOfMonth, $month, $partyear, $dayOfWeek, $dayOfYear, $daylightSavings) = localtime();
-    my $year = $partyear + 1900;
+    my $year = (localtime)[5] + 1900;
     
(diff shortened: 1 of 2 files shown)


mintsoft:
@jagtalon @moollaza what are the implications of only loading this once for the live goodie (if it was live). I presume that it would depend on the next reloading/recycle of the applications until the date and time is correct again?

mintsoft:
Well these have passed now. I've just tested moving the updating of `$year` etc into the `handle` so that the datetime is always updated when the goodie is used which appears to work

mintsoft:
Also harmonised the indentation on spaced

jagtalon:
@mintsoft nice! :+1:

PR #912, opened 2015-01-09
Updated DuckDuckGo Goodie

zekiel:
adding to the ddg_goodie
should've done --amend instead of continuous commiting.


zekiel:
//cc @nilnilnil

MrChrisW:
@zekiel This is failing as using `\` is not a valid way to escape single quotes in YAML (I know it's crazy) 

So you'll need to update [line 69 of responses.yml](https://github.com/duckduckgo/zeroclickinfo-goodies/blob/zac/privatesearchtrigger/share/goodie/duck_duck_go/responses.yml#L69) 

From: ` base_format: '[DuckDuckGo. The search engine that doesn\'t track you]'`

To: ` base_format: '[DuckDuckGo. The search engine that doesn''t track you]'`

zekiel:
Thanks @MrChrisW  --very much. 

updated!

jagtalon:
:+1:

PR #932, opened 2015-01-19
A basic LICENSE file

### LINKED ISSUE #931: Add a LICENSE file
Similar to https://github.com/duckduckgo/zeroclickinfo-spice/pull/1424 in Spice

th4t:
With a slightly modified APPENDIX section - no mentions of placeholder-field replacement.  Fixes #931, low-hangig fruit indeed.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -178,13 +178,15 @@
    APPENDIX: How to apply the Apache License to your work.
 
       To apply the Apache License to your work, attach the following
-      boilerplate notice. The text should be enclosed in the appropriate
+      boilerplate notice, with the fields enclosed by brackets "{}"
+      replaced with your own identifying information. (Don't include
+      the brackets!)  The text should be enclosed in the appropriate
       comment syntax for the file format. We also recommend that a
       file or class name and description of purpose be included on the
       same "printed page" as the copyright notice for easier
       identification within third-party archives.
 
-   Copyright 2015 DuckDuckGo, Inc.
+   Copyright {yyyy} {name of copyright owner}
 
    Licensed under the Apache License, Version 2.0 (the "License");
    you may not use this file except in compliance with the License.
@@ -197,4 +199,3 @@
    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
    See the License for the specific language governing permissions and
    limitations under the License.
-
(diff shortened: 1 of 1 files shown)


moollaza:
@th4t thanks a lot! I'm not actually sure you're supposed to modify the appendix though, hard-coding the year especially might not be good if we were to start adding the boilerplate to our documents.

Let's leave the License text as it comes and I'll be happy to merge this in :)

th4t:
Alrighty, vanilla content restored.

jagtalon:
@th4t @moollaza not sure if I understand correctly--why not put DuckDuckGo and 2015 in like https://github.com/duckduckgo/zeroclickinfo-spice/blob/master/LICENSE#L189?

moollaza:
@jagtalon I didn't see it was modified in the Spice License, that should be corrected then.

Modifying that text is actually modifying the License which I don't think you're supposed to do. The Apache foundation says to copy the license, which includes that boilerplate section at the bottom and then if you want, you're supposed to use the boilerplate as a header in all your source files, which we should investigate further and consider.

moollaza:
Fixed the Spice repo's LICENSE.

moollaza:
@th4t looks good now, thanks!

th4t:
What would make most sense as far as I see, would be to have the LICENSE file contain everything up to *END OF TERMS AND CONDITIONS*. Everything after this line can be included into source files, if needed.

I quite like how it is applied in [less.js](https://github.com/less/less.js). They have the [content of the license](https://github.com/less/less.js/blob/master/LICENSE) and a mention in the [README file](https://github.com/less/less.js#license).

jagtalon:
:+1:

moollaza:
@th4t I'm still unsure. It seems everyone applies the license either the way I've done it, or the way you're suggesting. GitHub's choosealicense.com suggest the current LICENSE file is correct: http://choosealicense.com/licenses/apache-2.0/

But looking at Twitter and Facebook's GH repos they both have the Apache license appear BOTH ways in different repos. Most of the Twitter repo's cut the License short where you've suggested as well.

As mentioned the Apache website says to "copy the file" and links to the fulltext so that's why I think it's supposed to include everything...

th4t:
@moollaza thanks for clarifying! :+1:  You are right, there seem to be at least two common ways, both are achieving their goal and in the end it depends on preference :)

PR #935, opened 2015-01-20
Fix returning error as structured content in ParseCron, e.g.

yegg:
Crontab a * * * * *


yegg:
cc @russellholt

nilnilnil:
:+1: nice.

PR #936, opened 2015-01-20
ParseCron.t: Let tests pass.

jagtalon:
HOTFIX

We can go back to it next time to show invalids. For now, the tests with the invalids return immediately.


russellholt:
:+1: 

Need to reconsider the `die()` short circuit within the `try` later.

Issue #770, opened 2014-11-21
Sunset / Sunrise doesn't work on IE 8


jagtalon:
Probably SVG issues.

------
IA Page: https://duck.co/ia/view/sun_info

iambibhas:
Sunset/sunrise seems to have stopped working altogether. It's triggering the weather IA in some cases.

moollaza:
@iambibhas and I talked, but for everyone else. I've made an internal task about the weather IA triggering. The SunInfo Goodie currently gives the user's sunrise/sunset based on the Geo::IP location. We're going to move this to use the Time&Date API so it supports other countries as well.

moollaza:
Closing. We stopped supporting IE 8 for Instant Answers a long time ago

Issue #626, opened 2014-09-09
Open Source DDG Constants answers


moollaza:
Currently DDG has some internal hardcoded answers for the following mathematical constants. We should put these in an open source Goodie so they can be expanded on and improved with some additional trigger words:

## constants
- 'stefan-boltzmann constant' => 'the Stefan Boltzmann constant = 5.6704 10^-8 kg s^-3 K^-4',
- 'boltzmann constant' => 'the Stefan Boltzmann constant = 5.6704 10^-8 kg s^-3 K^-4',
- 'speed of sound' => 'speed of sound at sea level = 340.29 m / s',
- 'speed of light' => 'the speed of light = 299 792 458 m / s',
- 'c' => 'the speed of light = 299 792 458 m / s',
- 'rydberg constant' => 'Rydberg constant = 10 973 731.6 m^-1',
- 'radius of mercury' => 'radius of Mercury = 2 439.7 kilometers',
- 'radius of venus' => 'radius of Venus = 6 051.8 kilometers',
- 'radius of earth' => 'radius of Earth = 6 378.1 kilometers',
- 'radius of the earth' => 'radius of Earth = 6 378.1 kilometers',
- 'radius mars' => 'radius of Mars = 3397 kilometers',
- 'radius of jupiter' => 'radius of Jupiter = 71 492 kilometers',
- 'radius of saturn' => 'radius of Saturn = 60 268 kilometers',
- 'radius of neptune' => 'radius of Neptune = 24 764 kilometers',
- 'radius of pluto' => 'radius of Pluto = 1180 kilometers',
- 'radius of the sun' => 'radius of the sun = 695 500 kilometers',
- 'proton mass' => 'proton mass = 1.67262158 10^-27 kilograms',
- 'mass of a proton' => 'proton mass = 1.67262158 10^-27 kilograms',
- 'planck\'s constant' => 'Planck\'s constant = 6.626068 10^-34 m^2 kg / s',
- 'plancks constant' => 'Planck\'s constant = 6.626068 10^-34 m^2 kg / s',
- 'h' => 'Planck\'s constant = 6.626068 10^-34 m^2 kg / s',
- 'permeability of free space' => 'the magnetic constant = 1.25663706 10^-6 m kg s^-2 A^-2',
- 'molar gas constant' => 'the molar gas constant = 8.314472 m^2 kg s^-2 K^-1 mol^-1',
[...]

PR #939, opened 2015-01-20
Solar System

MrChrisW:
# Solar System IA :rocket: :new_moon: 

Return various attributes of objects in the solar system

NASA: http://solarsystem.nasa.gov/planets/index.cfm

##### Supported Objects
*earth, jupiter, mars, mercury, neptun, saturn, uranus, venus, sun*

##### Supported Attributes
* radius, volume, mass, surface area
* size => radius
* area => surface area

##### Example Triggers
* `<attribute> of <planet>`
* `what is the <attribute> of <planet>`

Astronomers :smile:

** New screenshots see [below](https://github.com/duckduckgo/zeroclickinfo-goodies/pull/939#issuecomment-73118150)*

## Checklist

```
[X] Added metadata and attribution information (https://duck.co/duckduckhack/metadata)
[X] Wrote test file and added to t/ directory (https://duck.co/duckduckhack/goodie_tests)
[X] Verified that instant answer adheres to design guidelines (https://duck.co/duckduckhack/design_styleguide)
[X] Verified that instant answer adheres to code styleguide (https://duck.co/duckduckhack/code_styleguide)
[X] Tested cross-browser compatibility

    - Windows 8
        [X] Google Chrome
        [X] Firefox

    - Windows 7
        [X] Google Chrome
        [X] Firefox
        [X] IE 8

    - Windows XP
        [X] IE 8

	- Linux
        [X] Chromium
        [X] Firefox
```

------
> IA Page https://duck.co/ia/view/solar_system


lib/DDG/Goodie/Planets.pm
```diff
@@ -0,0 +1,109 @@
...
+  s/$attributesString//g; # Remove attributes
+  s/^\s+|\s+$//g; # Trim
+
+  return unless $_; # Return if empty query
+  $planetObj = $planets->{$_}; # Get planet data
+  return unless $planetObj; # Return if we don't have a valid planet
+
+  # Switch to imperial for non-metric countries
+  # https://en.wikipedia.org/wiki/Metrication
+  if ($loc->country_code =~ m/UK|US|MM|LR/i) {
```
MrChrisW: I'm going to get in trouble for this conditional statement :laughing: The only country I'm not sure about is the UK (they still have road signs in imperial)

Ultimately it may be best to show both like **3,928 miles (6,321 kilometers)** and swap the primary number based on the location.

share/goodie/planets/planets.yml
```diff
@@ -0,0 +1,154 @@
+---
+sun:
+  radius: 695,508 km
+  radius_imperial: 432,168.6 miles
```
MrChrisW: There's some incontinences with naming the units. Should `miles` be `mi` or should `km` be kilometers` ?

lib/DDG/Goodie/Planets.pm
```diff
@@ -0,0 +1,109 @@
...
+
+my %planetImages = (
+    earth => goodie_img_tag({filename => 'images/Earth.svg',height => 48, width => 48,}),
+    jupiter => goodie_img_tag({filename => 'images/Jupiter.svg',height => 48, width => 48,}),
+    mars => goodie_img_tag({filename => 'images/Mars.svg',height => 48, width => 48,}),
+    mercury => goodie_img_tag({filename => 'images/Mercury.svg',height => 48, width => 48,}),
+    neptune => goodie_img_tag({filename => 'images/Neptune.svg',height => 48, width => 48,}),
+    saturn => goodie_img_tag({filename => 'images/Saturn.svg',height => 48, width => 48,}),
+    uranus => goodie_img_tag({filename => 'images/Uranus.svg',height => 48, width => 48,}),
+    venus => goodie_img_tag({filename => 'images/Venus.svg',height => 48, width => 48,})
```
MrChrisW: These icons require [attribution](http://thenounproject.com/kevinwhitexD/) so we may want to change them for an internal design or completely remove images. I'll wait and see what happens when @abeyang gets his hands on this. :laughing:

MrChrisW:
IE8 can only do Base64 URI images up to 32kb in size. I'm unsure about other versions of IE? 

selection_075

[SunInfo](https://github.com/duckduckgo/zeroclickinfo-goodies/blob/master/lib/DDG/Goodie/SunInfo.pm) is also broken on IE8.

lib/DDG/Goodie/Planets.pm
moollaza: @MrChrisW actually we now have a Setting that users can set to signify their Units of Measure preference. We should switch it on the frontend according to that, like we do for Forecast

moollaza: https://github.com/duckduckgo/zeroclickinfo-spice/blob/master/share/spice/forecast/forecast.js#L377

MrChrisW: @moollaza Can I access this preference in the back-end (perl)? It's a little difficult to add front end JS to a Goodie, however it is possible.

moollaza:
@MrChrisW this is looking pretty good so far!

Regarding SunInfo there's some discussion about how to fix it. Really the problem is the Goodie::ImageRole implementation. We have a better way of handling SVG's on the frontend via JS/CSS Classes. See https://github.com/duckduckgo/zeroclickinfo-goodies/issues/770#issuecomment-67340979

MrChrisW:
Thanks @moollaza, hopefully the design can be improved upon. 

I'll use the CSS hack to workaround the SVG issues in IE, provided we still include images!

share/goodie/planets/planets.yml
```diff
@@ -0,0 +1,154 @@
+---
+sun:
```
MrChrisW: Remove the Sun? No image available, its also a star not planet.

t/Planets.t
```diff
@@ -0,0 +1,63 @@
...
+
+   	DDG::Request->new(query_raw => "size of jupiter", location => test_location("au")) => test_zci("Jupiter, Radius is 69,911 km",  html => qr{<div class="zci--planets"><span class="planets--planetImage"><img src=".*" height="48" width="48"/></span><span class="planets--info"><span class="text--primary planets--planetAttribute">69,911 km</span><span class="text--secondary planets--planetName">Jupiter, Radius</span></span></div>}),
+
+   	DDG::Request->new(query_raw => "size of planet jupiter", location => test_location("au")) => test_zci("Jupiter, Radius is 69,911 km",  html => qr{<div class="zci--planets"><span class="planets--planetImage"><img src=".*" height="48" width="48"/></span><span class="planets--info"><span class="text--primary planets--planetAttribute">69,911 km</span><span class="text--secondary planets--planetName">Jupiter, Radius</span></span></div>}),
```
MrChrisW: Fix indentation.

share/goodie/planets/planets.yml
```diff
@@ -0,0 +1,154 @@
...
+  density: 0.687 g/cm3
+uranus:
+  radius: 25,362 km
+  radius_imperial: 15,759.2 miles
+  radius_scientific: 2.5362 x 104 km
+  surface_area: 8,083,079,690 km2
+  surface_area_imperial: 3,120,894,516 mi2
+  surface_area_scientific: 8.0831 x 109 km2
+  volume: 68,334,355,695,584 km3
+  volume_imperial: #not found
```
MrChrisW: Fix this.

share/goodie/planets/planets.yml
```diff
@@ -0,0 +1,154 @@
...
+  mass: 1,989,100,000,000,000,000,000,000,000,000 kg
+  mass_imperial: 4,385,214,857,119,400,000,000,000,000,000 lbs
+  mass_scientific: 1.989 x 1030 kg
+  effective_temperature: 5504 °C
+  effective_temperature_imperial: 9939 °F
+  effective_temperature_scientific: 5777 K
+  density: 1.409 g/cm3
+mercury:
+  radius: 2,439.7 km
+  radius_imperial: 1,516.0 miles
```
MrChrisW: Round or remove decimal point in all values?

lib/DDG/Goodie/Planets.pm
mintsoft: For the record; the UK is metric first and imperial second

share/goodie/planets/planets.yml
mintsoft: Personally I'd keep everything SI (i.e. Metres)

share/goodie/planets/planets.yml
mintsoft: I'd say leave it in as it adds value; I'd "expect" sun results as well as planets. "SolarBodies.yml" ?

share/goodie/planets/planets.yml
mintsoft: You could keep them in if they're significant; i.e. shows the level of measurement accuracy

lib/DDG/Goodie/Planets.pm
MrChrisW: @mintsoft Thanks for clearing that up, I found it quite confusing that the country has adopted a partial metrication policy. In Australia we used British Imperial units back in the 70s (before my time) :laughing: 

So I'd like to clarify the best way to proceed:
 1. Switching units based on users preferences (not sure this is accessible without using JS)

 2. Switching units based on location (currently in use)

 3. Not switching units, all units as SI (modern metric)

Let me know your thoughts.

share/goodie/planets/planets.yml
MrChrisW: I agree :+1:

MrChrisW:
@mintsoft As my other response got buried In an outdated diff I'll re-paste here.

So I'd like to clarify the best way to proceed:

 1. Switching units based on users preferences (not sure this is accessible without using JS)

 2. Switching units based on location (currently in use)

 3. Not switching units, all units as SI (modern metric)

lib/DDG/Goodie/Planets.pm
mintsoft: I'd say SI and sod the Americans; however I'm sure they have other opinions :)

MrChrisW: @mintsoft I'm inclined to agree good sir, :tea: The fact the NASA website also lists the metric values first further justifies it. http://solarsystem.nasa.gov/planets/profile.cfm?Object=Mercury&Display=Facts

I'll await feedback :smile:

jagtalon:
@MrChrisW holy crap this looks pretty.

lib/DDG/Goodie/SolarSystem.pm
```diff
@@ -0,0 +1,113 @@
...
+
+# Load object data 
+my $objects = Load(scalar share('objects.yml')->slurp);
+
+# Handle statement
+handle query_lc => sub {
+  # Declare vars
+  my ($attribute, $attributesString, $result, $objectObj);
+  
+  s/what is (the)|(of)|(object)//g; # Remove common words, strip question marks
```
jagtalon: Should there be a `^` at the beginning of `what`?

lib/DDG/Goodie/SolarSystem.pm
```diff
@@ -0,0 +1,113 @@
...
+#Build HTML output
+sub pretty_output {
+  my ($result, $operation, $image) = @_;
+  my $html = "<div class=\"zci--objects\">";
+  $html .= "<span class=\"objects--objectImage\">";
+  $html .= $image;
+  $html .= "</span>";
+  $html .= "<span class=\"objects--info\">";
+  $html .= "<span class=\"text--primary objects--objectAttribute\">";
+  $html .= $result;
```
jagtalon: Encode these?

jagtalon:
@abeyang I wonder how we can do the attribution with this one.

https://ddh1.duckduckgo.com/?q=what+is+the+size+of+venus&ia=answer
https://ddh1.duckduckgo.com/?q=size+of+earth&ia=answer

jagtalon:
With attribution I mean the person who made the images.

MrChrisW:
@jagtalon I was hoping @abeyang could make me some new sexy looking planets? :laughing:

lib/DDG/Goodie/SolarSystem.pm
MrChrisW: Unfortunately that would result in an issue with using the `<sup></sup>` tags in [line 83](https://github.com/MrChrisW/zeroclickinfo-goodies/blob/060a2af098e4c000271813aadb16cb5783093247/lib/DDG/Goodie/SolarSystem.pm#L83) all values are pulled directly from the YML data, It's doubtful there's any XSS risk here.

jagtalon:
@MrChrisW oh that's another route.

abeyang:
Hi @MrChrisW , apologize for my silence... having some internal discussions re: planet images. Hope to get back to you soon!

jagtalon:
@MrChrisW I'm not sure how important it is, but I think it would be good if we could specify if its either mean or equatorial radius that we're displaying.

MrChrisW:
@jagtalon Sure, I think mean radius is assumed but it's probably best to include it.

abeyang:
Hey @MrChrisW, just wanna give you an update. We're thinking of possibly creating the icons ourselves, but we don't really get to it til sometime next week. Hope you're OK waiting until then...

MrChrisW:
Hey @abeyang no problem, looking forward to seeing your grand design concept  :smile:

abeyang:
Hey @MrChrisW , sorry for the delay. We had to make the rounds with @thm for these planet designs. We *may* need to update Saturn, but aside from that, the rest are ready to go!

In the end, it should look something like this:

planets-v1 1-earth

Note that Saturn has to push the text out a bit:

planets-v1 1-saturn

And here are the assets:
http://cl.ly/2s2r3O2o3n1r

Can't wait to see this up and running!

mintsoft:
@abeyang @thm Those look awesome!

MrChrisW:
@abeyang Fantastic, thanks! Will implement shortly.

P.S Thanks @thm

MrChrisW:
Updated as per @abeyang feedback, let me know if the styles need adjusting. 

I think the `margin-right` on `objectImage` needs adjusting when the image is Saturn as it pushes the text to far to the right. http://duckpan1.mrchris.me/?q=size+of+saturn
 
Viewable here: http://duckpan1.mrchris.me/?q=size+of+earth

Still a few more issues I'm looking to clear up. 

+ Decide if this [conditianl statment](https://github.com/duckduckgo/zeroclickinfo-goodies/pull/939/files#diff-cde95278264cf233c2de28ccf5ffa736R56) used to switch units from metric to imperial is required. Otherwise just display in metric. 
+ Fix [volume_imperial](https://github.com/duckduckgo/zeroclickinfo-goodies/pull/939/files#diff-cdfd7be09ec7cf342db5db5fd676ddfeR129) for uranus
+ Double check all planet values are correct. 
+ Resolve style issue with Saturn.
+ Include Pluto data? as we have an image for it. (poort pluto :cry: )

MrChrisW:
Looks like some internal triggers will need to be removed if this is released #626

MrChrisW:
The CSS loading after the answer has been displayed really annoys me, is this only on DuckPan? 

out

moollaza:
@MrChrisW 

>Looks like some internal triggers will need to be removed if this is released #626

Yup but that's a quick fix ;)

>The CSS loading after the answer has been displayed really annoys me, is this only on DuckPan? 

I noticed that too. I'm not entirely sure, but I think we recently discovered a Goodie-Spice race condition that we're working to fix. I see the CSS only loads after the 'Images' and 'Videos' tabs appear.

------

Nasa has a TON of awesome data on these planets, http://solarsystem.nasa.gov/planets/profile.cfm?Object=Uranus&Display=Facts&System=Metric has the missing Uranus data I believe

>Include Pluto data?

Yes! It's technically a "dwarf planet" so I would argue that belongs to the "Planets" Goodie :)

MrChrisW:
@moollaza 

>I noticed that too. I'm not entirely sure, but I think we recently discovered a Goodie-Spice race condition that we're working to fix. I see the CSS only loads after the 'Images' and 'Videos' tabs appear.

Great, looking forward to seeing a fix in the future. 

>Nasa has a TON of awesome data on these planets,

That's my source! Shhh don't tell anyone.

>Yes! It's technically a "dwarf planet" so I would argue that belongs to the "Planets" Goodie :)

I made the choice to change the name to SolarSystem Goodie as techically the sun isn't a planet! :laughing: I'll add pluto pronto :+1:

MrChrisW:
This should be ready for a design re-review from @abeyang 

+ View online: [Earth](http://duckpan1.mrchris.me/?q=size+of+earth) | [Saturn](http://duckpan1.mrchris.me/?q=size+of+saturn)
+ [CSS](https://github.com/MrChrisW/zeroclickinfo-goodies/blob/planets/share/goodie/solar_system/solar_system.css)

@moollaza @jagtalon Still need an answer regarding the display of metric vs non-metric? See [comment](https://github.com/duckduckgo/zeroclickinfo-goodies/pull/939#issuecomment-70767197)

mintsoft:
@MrChrisW Metric! Sod the yanks!

MrChrisW:
@mintsoft Amen to that! :laughing:

abeyang:
Sorry @MrChrisW -- had to deliberate on how to better show the planets. We finalized with this:

planets-v2-earth

planets-v2-saturn

The background cover is 60px, and each planet/sun/moon is 40px (all except Saturn). Here are the updated assets: http://cl.ly/0d2b1M0Q0E0i

Finally, can we try to vertically-center the text with the cover? Seems a bit misaligned right now. You can try to use the classnames proposed under **Images** in our [styleguide](https://duckduckgo.com/styleguide):

```html
<div class="zci__main">
	<div class="zci__img-wrap">
		<img class="zci__img">
	</div>
	<div class="zci__body">Text Goes Here</div>
</div>
```

Thanks! Can't wait to see this go live!

MrChrisW:
@abeyang Thanks for getting back with an update on this :smile: I'll update and push shortly!

moollaza:
@MrChrisW just checking in -- is this ready for us or are there more changes coming?

MrChrisW:
@moollaza This is still awaiting some small changes from **me** :smile: I'll ping when ready :+1:

MrChrisW:
@abeyang This has been updated :+1: 

+ View online: [Earth](http://duckpan1.mrchris.me/?q=size+of+earth) | [Saturn](http://duckpan1.mrchris.me/?q=size+of+saturn)

I had a go at using the CSS classes recommended in the styleguide (which looks amazing btw :astonished: ) however after a while I decided to fallback on the classes I'd already created as I couldn't get it looking quite right :frowning: 

Text should be vertically centred now:

menu_295
menu_296

mintsoft:
@MrChrisW Was just going to tell you about that ;)

MrChrisW:
@mintsoft You're totally the reason I added that edit! :laughing:

MrChrisW:
Any suggestions on dealing overflow of large numbers on mobile? We could show the scientific notation :grey_question: 

selection_300

mintsoft:
@MrChrisW Definitely scientific notation. Or alternatively do it in zetta-lbs?

MrChrisW:
@mintsoft Thanks! I'm leaning towards scientific notation, I feel like we need to choose one way or the other not a mix of both. I'd appreciate any feedback from the design team or other staff when available :+1:

mintsoft:
Yeah, makes sense @abeyang or @chrismorast any opinion or thoughts?

abeyang:
Can we do something like this:
planets-v2-saturn-mass

So, I know we also mentioned having the western notation, but thinking back to my high school days of unit conversions (way back in the day), I don't ever recall using pounds -- whereas miles (vs km) still seems to make some sense here.

In any case, the more important factor is showing the exponential factor; multiplication sign is `&times;` I believe.

MrChrisW:
@abeyang thanks. The scientific notations provided by NASA only include metric, I'm guessing I'll need to convert the values. See [example data](https://solarsystem.nasa.gov/planets/profile.cfm?Object=Mars&Display=Facts)

My main question is, should we display all values as scientific notation or only the few that are too long to be displayed in decimal notation?

abeyang:
@MrChrisW I don't think we need to use scientific notation for everything. Maybe a good basis (for now) for when to use scientific notation is when the value goes above one million...

MrChrisW:
@abeyang Updated again to reflect the required changes:

Examples:
http://duckpan1.mrchris.me/?q=mass+moon
http://duckpan1.mrchris.me/?q=volume+neptune
http://duckpan1.mrchris.me/?q=mass+of+earth

Note the differing number of decimal places on Neptune vs Moon, I'm unsure if we should round all these values to 3 decimal places. @mintsoft

mintsoft:
@MrChrisW I'd say round it for consistency; it looks a little weird having different sig-figs

MrChrisW:
selection_304

IE 8 planet vertical align is broken.

MrChrisW:
Fixed IE8 issue by removing the use of CSS3 `transform` and reverting to table-cell and vertical-align :frowning: rounded the scientific notations.

mintsoft:
@MrChrisW If the images were changed to be `div`s with a background image of the image that you are looking for then you could switch them out with media queries?

moollaza:
@MrChrisW @mintsoft we should be able to handle the img loading in the template, if you're interested in updating this to use the awesome new Goodie Templates?!

We'd need to use a custom template here, but that shouldn't cause any problems. Here's an example to follow @MrChrisW: https://github.com/duckduckgo/zeroclickinfo-goodies/blob/master/lib/DDG/Goodie/BPMToMs.pm#L64

I swear actual documentation is coming soon :smile:

MrChrisW:
Thanks for the suggestions @mintsoft @moollaza 
I'll update to use a template, looks simple enough :wink: 
>I swear actual documentation is coming soon :smile:

I'll believe that when I see it ! :laughing:

mintsoft:
@moollaza Heard it all before ;)

MrChrisW:
@moollaza I've begun the process of converting the IA to use the new templates. 

The issue I'm running into when using the `base` template, it seems to want to return multiple results as tiles, see an example of my [structured_answer](https://gist.github.com/MrChrisW/b8313af838ae7331a3f8) and the result 
selection_306

It's probably something silly I'm overlooking! Thanks

moollaza:
@MrChrisW it looks like you disabled the detail template? I believe that's causing the problem?

MrChrisW:
@moollaza :+1:

MrChrisW:
@moollaza Updated to use the Goodies template. [Working well](http://duckpan1.mrchris.me/?q=size+of+mars&ia=answer)

>We should be able to handle the img loading in the template

Further on this comment, I take it we simply want to send the planet name to the template and load the image using an `img` tag (easy!). 

I'm still confused as to how this will enable us to detect mobile and change the image accordingly?

share/goodie/solar_system/solar_system.css
```diff
@@ -0,0 +1,29 @@
+.zci--solar_system .objects--objectImage {
+    height: 60px;
+    width: 60px;
+    display: table-cell;
+    vertical-align: middle;
+    background-color: #333;
+    border-radius: 5%;
```
MrChrisW: Won't work with IE8 :+1:

MrChrisW:
@moollaza I've added a new mobile template. Using the `detail_mobile` option. See 
[mobile.handlebars](https://github.com/MrChrisW/zeroclickinfo-goodies/blob/planets/share/goodie/solar_system/mobile.handlebars), [content.handlebars](https://github.com/MrChrisW/zeroclickinfo-goodies/blob/planets/share/goodie/solar_system/content.handlebars)

I've moved the width & height values inline in order to set a differing height width if the image is Saturn. 

@abeyang I've noticed on mobile the `.zci` selector padding is modified to 0px. Can I set the padding here as it makes things look a little strange (see below)
```css
.zci {
    padding-left: 0px;
    padding-right: 0px;
}
```

selection_313

share/goodie/solar_system/solar_system.css
mintsoft: Should be fine; it'll degrade gracefully

mintsoft:
@MrChrisW Looks like there's a `html_enc` problem here:
http://duckpan1.mrchris.me/?q=mass+of+earth&ia=answer

MrChrisW:
@mintsoft Nice catch, that's what I get for not including full tests of returned `structured_data` :frowning: 

Fixed :tada:

mintsoft:
@MrChrisW Not sure what i've done here, I was just resizing the window a little and it's got stuck like this:
image

moollaza:
@MrChrisW I spoke with @chrismorast and he says we should be trying to use SVG wherever possible (which means we'll need to get SVG assets from the design team wherever we're currently using PNG), which would really simplify image handling (because we won't need to check for retina devices.

For Goodies, if we have an SVG we can slurp the content directly into the HTML.

> am I linking to the images correctly? 

@MrChrisW when we link to images in production, the path are versioned, so you'll need to include the current Goodie version number. For now, you can get the current version number in Perl with `{{ENV{DDG_GoodieBundle_OpenSourceDuckDuckGo}}}`:

```handlebars
<img src="/share/goodie/{{goodie_version}}/{{img}}">
```

moollaza:
Once we have SVG you could slurp the SVG straight into the HTML. We could even Base64 encode them!

mintsoft:
@moollaza does the push towards SVG's mean we're sidelining IE8? :grinning:

moollaza:
@mintsoft damnit, that's why we're using PNG's. I couldn't remember the reason. Everywhere else on the site we fallback to PNG when SVG won't work (we use Modernizr, so `Modernizr.svg` tells us when we can use them). I'm not sure if we have a simple solution for that via the templates yet so I guess we'll need to stick to PNG for now.

MrChrisW:
@mintsoft Confirmed broken in Chrome! Thanks for testing :+1: 

@moollaza The IA was originally using SVG images and slurping them using `DDG::GoodieRole::ImageLoader` then the design team switched to PNG images :confused: I'll leave it as is for now...

Regarding the image paths, I'm assuming the correct path is like this ?
`<img src="/share/goodie/{{goodie_version}}/solar_system/img/{{imageName}}.png`

I'll leave this until the last second (commit) as this will break testing on DuckPAN :+1:

MrChrisW:
@mintsoft I think I've fixed the Chrome newline issue, can you [confirm](http://duckpan1.mrchris.me/?q=mass+of+earth&ia=answer)? Thanks :+1:

mintsoft:
@MrChrisW Yup, all fixed! 
image

MrChrisW:
Suggestion: Add the ability to switch units using a trigger like: 
+ Search Default = Location specific
+ Search `mass sun kg` = Metric
+ Search `mass sun lbs` = United States customary units

mintsoft:
That sounds cool @MrChrisW although I'd probably also want `mass sun metric` or `mass sun imperial`

moollaza:
> I'll leave this until the last second (commit) as this will break testing on DuckPAN :+1: 

What error are you getting from DuckPAN? I thought I fixed the subdirectory problem...

MrChrisW:
@moollaza 
>when we link to images in production, the path are versioned, so you'll need to include the current Goodie version number.

Including the versioned path in a DuckPAN environment won't work as the path is only versioned in prod? Right?

Secondly I cannot seem to use the environment var like this 
```perl
{{ENV{DDG_GoodieBundle_OpenSourceDuckDuckGo}}}
``` 

as it results in 
```perl 
Bareword "DDG_GoodieBundle_OpenSourceDuckDuckGo" not allowed while "strict subs"
```

MrChrisW:
@mintsoft Those types of queries should work now :+1: There's probably improvements that can be made, see a81199d

+ http://duckpan1.mrchris.me/?q=size+of+mars+metric&ia=answer
+ http://duckpan1.mrchris.me/?q=mass+sun+kg&ia=answer

mintsoft:
@MrChrisW an improvement that springs to mind is supporting `mass of mercury in kg` or `size of venus in metric`

MrChrisW:
@mintsoft :+1: these queries work now.

moollaza:
>Including the versioned path in a DuckPAN environment won't work as the path is only versioned in prod. Right?

@MrChrisW  DuckPAN will ignore the version number so versioned paths should work fine :)

>Secondly I cannot seem to use the environment var like this 

My bad, I totally wasn't thinking straight. We export that variable in Production so it won't be available in DuckPAN. The installed Goodie Package will have a version though that you can access locally:

```perl
my $version = $DDG::GoodieBundle::OpenSourceDuckDuckGo::VERSION // 0;
```

This will give you `0` when using DuckPAN (because it loads the files from the repo's local `/lib` instead of your `/perl ` install dir, but it should work for us in Production. I'll make a task to expose the Goodie version to the JS so you can access it more easily from the templates (we do this already for `spice_version` but haven't needed it yet for Goodies).

MrChrisW:
@moollaza No problem! Thanks.

>DuckPAN will ignore the version number so versioned paths should work fine

Doesn't look like that's the case as accessing `/share/goodie/0/solar_system/img/mars.png` fails with `ERROR: File not found`

Example: 
+ http://duckpan1.mrchris.me/share/goodie/solar_system/img/mars.png 
+ http://duckpan1.mrchris.me/share/goodie/0/solar_system/img/mars.png

Looks like it may have something to do with commit [021f3a2bc7792b83eef5512d17860510ba6c1b08](https://github.com/duckduckgo/p5-app-duckpan/commit/021f3a2bc7792b83eef5512d17860510ba6c1b08#diff-ebc5e1818c64eceac5960f3997f6aca9R98)

abeyang:
@MrChrisW looking good! Can we make these minor font adjustments:

``` css
.zci--solar_system .objects--objectAttribute {
  font-size: 1.319em;
}

.zci--solar_system .objects--objectName {
  font-size: 0.972em;
}
```

Almost there!

moollaza:
>Doesn't look like that's the case as accessing /share/goodie/0/solar_system/img/mars.png fails with ERROR: File not found

@MrChrisW mind changing `0` to `000` or `999` for now? It's a hack but it'll get things moving ahead for you. I've made a task for better image loading for Goodies. DuckPAN only expects `\d{3,4}` because that's where we're at for Spice versions. I feel like there was a reason why I didn't just generalize it to `\d+` but I can't think of it right now.

MrChrisW:
@moollaza I did see that line `\d{3,4}` and tried accessing using 3-4 numbers however this still fails with the same result:

Test: http://duckpan1.mrchris.me/share/goodie/000/solar_system/img/mars.png

Thanks

MrChrisW:
@abeyang Hello! Thanks will update with this adjustments, can you comment on the `padding` value on mobile? Thanks

moollaza:
@MrChrisW I'm an idiot -- made a typo in the share path structure!

http://duckpan1.mrchris.me/share/goodie/solar_system/000/img/mars.png

The version goes after the IA name.

MrChrisW:
@moollaza Awesome! I'll update shortly. Glad we finally got to the bottom of that :laughing:

mintsoft:
@MrChrisW are you just waiting on feedback from @abeyang now?

MrChrisW:
@mintsoft That's correct. 

In the meantime feel free to nitpick at any of the code :smile:

lib/DDG/Goodie/SolarSystem.pm
```diff
@@ -0,0 +1,133 @@
...
+
+my @unitTriggers = ( 'kg', 'km2', 'km3', 'km', 'mi2', 'mi3', 'mi', 'lbs', 'metric', 'imperial');
+
+triggers any => @triggers;
+
+# Load object data 
+my $objects = Load(scalar share('objects.yml')->slurp);
+
+# Handle statement
+handle query_lc => sub {
```
mintsoft: @MrChrisW Speaking of nit-picking; this seems to be indented to 2 spaces when it should be 4!

lib/DDG/Goodie/SolarSystem.pm
```diff
@@ -0,0 +1,133 @@
...
+    my $number = $1; my $exponent = $2;
+    $result =~ s/$number$exponent/$number<sup>$exponent<\/sup>/;
+    $result =~ s/x/&times;/;
+  }
+
+  #$saturn var is provided to handlebars template to set size of image
+  $saturn = ($objectName eq "saturn") ? 1 : 0;
+
+  #Return result and html
+      return $operation." is ".$result,
```
mintsoft: Also odd indentation here

lib/DDG/Goodie/SolarSystem.pm
```diff
@@ -0,0 +1,133 @@
...
+  
+  # Convert flag surface_area = Surface Area
+  if($attribute =~ /_/ ) { $attribute = join ' ', map ucfirst lc, split /[_]+/, $attribute; }
+
+  # Human friendly object name + attribute
+  my $operation = ucfirst($_)." - ".ucfirst($attribute);
+
+  # Superscript for km3, mi3, km2 or mi2 
+  if($result =~ m/(km|mi)(\d)/) {
+    my $symbol = $1; my $superscript = $2;
```
mintsoft: You could do
```perl
my ($symbol, $superscript) = ($1, $2);
```
if you wanted?

share/goodie/solar_system/content.handlebars
```diff
@@ -0,0 +1,19 @@
+<div class="objects--body">
+    <span class="objects--objectImage">
+        {{#if saturn}}    
+            <img width="45px" height="28px" src="share/goodie/solar_system//{{goodie_version}}/img/{{imageName}}.png">
```
mintsoft: Hmm I think if the width and height aren't set then it'll default to the resolution of the png? So we can bomb the if and the special "is saturn" flag too?

share/goodie/solar_system/mobile.handlebars
```diff
@@ -0,0 +1,19 @@
+<div class="objects--body">
+    <span class="objects--objectImage">
+        {{#if saturn}}    
+            <img width="45px" height="28px" src="share/goodie/solar_system//{{goodie_version}}/img/{{imageName}}@2x.png">
```
mintsoft: Same again, although I'm guessing the sizes are here for increased screen pixels? Can we set them to 50% to get the same effect instead?

MrChrisW:
@mintsoft Thanks for nitpicking! :smile:

share/goodie/solar_system/mobile.handlebars
MrChrisW: Regarding the inline styles, we can remove the `width` and `height` values for the main `content.handlebars` (desktop) and the images will look normal. 

However with the `mobile.handlebars` as the images are double the size `width` and `height` need to be set, unfortunately setting these values to `50%` will only set them image to half the size of it's [surrounding container](https://github.com/MrChrisW/zeroclickinfo-goodies/blob/planets/share/goodie/solar_system/solar_system.css#L2-L3) (60px, 60px)

abeyang:
@MrChrisW looks good! One **final** nitpick -- can we have the surface area return in scientific notation as well? Any reason why it's not?

screen shot 2015-04-13 at 12 28 44 pm

chrismorast:
Are you sure @abeyang ? I much prefer to see the number written out over the scientific notation

MrChrisW:
@abeyang Any numbers that overflow on a screen 350px> are scientific notation. Sun Surface Area does not :smile: 
selection_345

@chrismorast @abeyang Are you able to recommend a [fix for the `padding-left`](https://github.com/duckduckgo/zeroclickinfo-goodies/pull/939#issuecomment-89159354) on mobile?

abeyang:
Hey @MrChrisW , I'm sorry, but my dev server decided to stop working for me. Can you throw it onto yours for me? Thanks so much!

abeyang:
For the mobile, let's try this:

``` css
@media only screen and (max-width: 44em) {
  .zci {
    padding-left: 1.2em;
  }
}
```

@sdougbrown @moollaza Please LMK if you have better suggestions...

MrChrisW:
@abeyang Thanks, looks like my server decided to kill DuckPAN too :laughing: [Fixed now](http://duckpan1.mrchris.me/?q=mass+of+the+sun&ia=answer)

I'll update with the new CSS once @moollaza approves :smile: I wonder if there's any other IAs that push the padding-left on mobile?

MrChrisW:
Ping @moollaza

moollaza:
@MrChrisW sorry, that particular CSS won't work because it'll affect any/all other IA's when this Solar System IA CSS loads -- I'll take a peak today to figure it out!

moollaza:
@MrChrisW looks like the `<div class="object--body">` should have the class `zci__body` -- that'll add the correct side padding:

2015-04-17_11h17_59

That may be a missing component of the template though -- I'll double check.

moollaza:
The problem is that the mobile_detail isn't getting wrapped in our `base_detail` template. If you specify `wrap_detail => 'base_detail', it fixes the mobile view, but double-wraps the normal, detail view, which is a bug. I'll make a task to fix this.

/cc @sdougbrown

MrChrisW:
@moollaza Thanks for looking into that.

Should I proceed to update, specifying `wrap_detail` and await the internal fix or leave as is?

moollaza:
@MrChrisW I think the best way forward is to add `zci--body` to your mobile template for now. We'll need to coordinate the internal fix with removing that to prevent double padding though

MrChrisW:
@moollaza Sure, I've added `zci__body` to the mobile template.

http://duckpan1.mrchris.me/?q=size+of+mars+metric&ia=answer

moollaza:
@MrChrisW :+1: 

@abeyang I believe this is ready for a final design review? The mobile padding has been fixed. If that's all that was needed please merge :+1:

abeyang:
:boom:

mintsoft:
Awesome; can't wait to see this live @MrChrisW !

MrChrisW:
Thanks for the great feedback everyone :+1:  @mintsoft ! @moollaza @abeyang @jagtalon @chrismorast 
_____________________________________

jagtalon:
01n4njx

moollaza:
@MrChrisW sorry for the long delay, I hadn't realized this was merged and ready for release. Will set it up to go live with the next set of IAs!

MrChrisW:
@moollaza I was starting to wonder.. :wink: Thanks for the update!

moollaza:
@MrChrisW IT'S LIVE!

https://duckduckgo.com/?q=mass+of+earth&t=ffab&ia=answer

Sorry about the long delay. Thanks for another awesome IA :+1:

MrChrisW:
@moollaza AMAZING!! Thanks

PR #943, opened 2015-01-21
Words goodie using DAWGs

W25:
A version of https://github.com/duckduckgo/zeroclickinfo-goodies/pull/816 using DAWGs (lower memory usage).

The data structure is described in the following articles:
* Appel, Jacobson. The world's fastest Scrabble program. Communications of the ACM. Volume 31 Issue 5, May 1988.
* http://www.strchr.com/ternary_dags
* http://www.strchr.com/dawg_predictive
* http://stevehanov.ca/blog/index.php?id=115

I tried both ternary DAWG and a DAWG with sorted array of edges for the provided dictionary (enable2k.txt). The latter was 20% smaller, so I chose this data structure.

Problems:
* The returned words are not random, because a DAWG implies sorting. It would be slower (but still possible if you wish) to return random words.
* The existing goodie NLetterWords.pm processes some of the queries (e.g., '8 letter words'), but Pavlos' goodie can do more, so we should remove or disable NLetterWords.pm, because the two goodies may conflict.
* We use the 'any' trigger for 'words', but the query '8-letter words' does not trigger this goodie. It may be a conflict with NLetterWords.pm or some bug in the IA plugin infrastructure.


nilnilnil:
@W25 this is pretty cool. Did you evaluate any of the methods discussed in #816 like CDB, SQLite etc? What compelled you to take this approach? Just trying to understand the thought process.

W25:
@nilnilnil Thank you. I studied [SQLite documentation](https://sqlite.org/optoverview.html) and found that the LIKE optimization takes into account only the initial sequence of non-wildcard characters. SQLite still has to check a lot of candidate words in our case. CDB is a simple key-value storage, which is not enough for this task. Previously, I did [a similar job for Mozilla](http://www.strchr.com/dawg_predictive) using DAWGs.

Pavlos @pavlos256 asked about returning random words. We can do this (using randomized BFS), but the matching function will be large, complicated, and slow. Would you prefer to make it faster and return sorted results, or to make it slower and return random results?

W25:
@nilnilnil Is it worthwhile to work out a more complex and slow algorithm for returning random words?

nilnilnil:
I'd like @pavlos256 to weigh in first.

pavlos256:
In short, this is very promising and uses a tiny amount of memory. It always returns the same words though, so I personally consider it not good enough for inclusion.

It is also considerably slower when it has to match wildcards but I think this could be overlooked if the other problem was fixed.

Here's a detailed comparison: https://i.imgur.com/LbYaOdd.png

Each query was matched 1001 times, loading time & parsing was excluded.

mintsoft:
@pavlos256 what unit is the duration in on that spreadsheet, I'm guessing seconds?

pavlos256:
@mintsoft Yes, you're right. It was measured using the Benchmark module and the results were so consistent that I found 3x1000 iterations enough.

nilnilnil:
I don't think those are acceptable numbers for a Goodie. I'm sorry about that. I think as an absolute maximum it should be 50ms. That even makes me uncomfortable. That is like reading 2MB+ sequentially from a spinning disk. These operations won't happened out of band, and we don't defer computation since we've determined a good IA fit based on query space.

W25:
@nilnilnil Thank you, I will try to find a faster algorithm. If I understand correctly, Pavlos made 1000 iterations, so the maximum time is around 5 ms for my version and 3 ms for his version.

moollaza:
@W25 just checking in here. Are you still interested in finishing this one up then? Shall I close the PR for now?

W25:
@moollaza @nilnilnil  I found no silver bullet, but there is an engineering compromise. We can generate random words using DAWGs, but the code is slower. If you need more speed, I can re-implement the algorithm in C as a Perl extension (XS). What do you think about this?

nilnilnil:
I'd like to hold off. @W25 we're already working with you on other things! :smile:

moollaza:
I'll mark this one as "On Hold" and close it for now then.

Thanks again @W25 !

PR #950, opened 2015-01-23
Convert Week Goodie to Structured Template, update tests

### LINKED ISSUE #712: Reviving the "Week" Goodie
This Goodie was merged a long time ago (originally submitted by @gsquire), but for some reason is not deployed to the site (perhaps taken down during the redesign and forgot to be re-deployed?)

Here's a picture of it:

what_is_the_current_week_at_duckduckgo

//cc @jagtalon @mwmiller 

@chrismorast any design suggestions? Shall we add some CSS to bump the font size to 1.5em?

@gsquire I noticed the trigger seems to require "what is", for example "23rd week of 1991" wouldn't trigger, perhaps the regex could be relaxed?

@mwmiller is this IA a candidate for the Goodie Dates Role?

moollaza:
Fixes #712.

My only thought is that I've kinda "abused" the `operation` to provide a subtitle. I tried a few permutation where I properly indicated the operation and input but in every case it looked repetitive and ugly.

//cc @jagtalon @mintsoft @gsquire @javathunderman

2015-01-22_19h01_30

---

The triggering could be relaxed a bit, but I think that's an issue for another day. Someone from the community is welcome to tackle that ;)

### DIFF
--- a/lib/DDG/Goodie/Week.pm
+++ b/lib/DDG/Goodie/Week.pm
@@ -63,23 +63,32 @@ handle query_raw => sub {
 
     return if $week =~ s/(nd|th|rd|st)$// and $week > 52;
 
-    my $dt = DateTime->now(time_zone => $loc->time_zone)
-        if ($week eq 'current' or $year eq 'current');
+    my $dt = DateTime->now(time_zone => $loc->time_zone) if ($week eq 'current' or $year eq 'current');
+
+    my ($response, $operation);
 
     if ($week eq 'current' and $year eq 'current') {
-        return "We are in currently in the " . ordinate($dt->week_number) .
-                ' week of ' . $dt->year . '.',
-            html => 'We are in currently in the ' . $dt->week_number
-                    . '<sup>' . ordsuf($dt->week_number) . '</sup>'
-                    . ' week of ' . $dt->year . '.';
+        my ($dt_week_num, $dt_year) = (ordinate($dt->week_number), $dt->year);
+        $response = "We are currently in the $dt_week_num week of $dt_year.";
+        $operation = "Current week";
     } elsif ($year eq 'current') {
         $year = $dt->year();
     }
-    my (undef, $month, $day) = Monday_of_Week($week, $year);
-    return "The " . ordinate($week) . " week of $year began on " .
-        "$months[--$month] " . ordinate($day) . '.',
-        html =>"The $week<sup>" . ordsuf($week) . "</sup> week of $year began on " .
-            "$months[$month] $day<sup>" . ordsuf($day) . '</sup>.';
+
+    unless ($response){
+        my (undef, $month, $day) = Monday_of_Week($week, $year);
+        my ($week_num, $day_num, $out_month) = (ordinate($week), ordinate($day), $months[--$month]);
+
+        $response = "The $week_num week of $year began on $out_month $day_num.";
+        $operation = "Find the $week_num week of $year";
+    }
+
+    return $response,
+      structured_answer => {
+        input     => [],
+        operation => "Assuming the week starts on Monday",
+        result    => $response
+      };
 };
 
 1;
--- a/lib/DDG/Goodie/Week.pm
+++ b/lib/DDG/Goodie/Week.pm
@@ -42,45 +42,53 @@ my @months = qw/
     December
 /;
 
-handle query_raw => sub {
+handle query_lc => sub {
     return unless /^\s*
-        what(?:'?s|\sis|\swas)?\s+
-        (?:the\s+)?
-        (?:(current|(\d{1,2})(?:nd|th|rd|st)?)\s+)?
+        what(?:'?s|\sis|\swas)?\s
+        (?:the\s)?
+        (?:(current|(?<week>\d{1,2})(?:nd|th|rd|st)?)\s)?
         week
         (
-            \s+of\s+
-            (?:(?:the|this)\s+)?
-            (year|\d{4})
+            \sof\s
+            (?:(?:the(?:\scurrent)?|this)\s)?
+            (?<year>year|\d{4})
             |
-            \s+is\s+this
+            \sis\sthis
         )?\??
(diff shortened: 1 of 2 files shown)


MrChrisW:
@moollaza Will be good to see this finally released into the wild :cactus:  

One thing I've noticed is the use of the *simple past tense* **began** for dates in the future, this seems incorrect. Can a conditional statement to change wording be added?

**Current**
selection_105


**Suggestions:**
selection_107
selection_108
selection_109

lib/DDG/Goodie/Week.pm
```diff
@@ -63,23 +63,32 @@ handle query_raw => sub {
...
+        my ($week_num, $day_num, $out_month) = (ordinate($week), ordinate($day), $months[--$month]);
+
+        $response = "The $week_num week of $year began on $out_month $day_num.";
+        $operation = "Find the $week_num week of $year";
+    }
+
+    return $response,
+      structured_answer => {
+        input     => [],
+        operation => "Assuming the week starts on Monday",
```
mintsoft: We could support this as a mechanism if we just change structured answers to have "subtitle" instead of "operation" ?

javathunderman:
@moollaza @MrChrisW Nice to see this moving. Just a question, what is the status on the other "release goodie" issues? Can they be put up?

lib/DDG/Goodie/Week.pm
moollaza: @mintsoft I agree, but on a very awesome note, we already have a working demo of Goodies passing along data to the frontend (in an almost identical fashion to Spice's passing an object to `Spice.add()`)

Once that's available I suspect we'll be able to port Goodies to use that approach where we'd have "title", "subtitle" etc.

moollaza:
@MrChrisW really good find regarding the trigger/word tense. I'll see if I can fix that up

moollaza:
@javathunderman I'll run through them to see if any need updating. We should be able to launch the ones that don't.

moollaza:
Should be fixed now!

javathunderman:
@moollaza Great!

lib/DDG/Goodie/Week.pm
mintsoft: Awesome!

mintsoft:
@moollaza the tests are failing across the board on this PR, I'm guessing it's something else that's done it though?

MrChrisW:
@mintsoft Travis is failing to include Goodie.pm so all tests are currently failing. I was going to lodge an issue but I think @moollaza is aware of it.

javathunderman:
@MrChrisW Yikes. When is that going to be fixed?

moollaza:
@MrChrisW @mintsoft not sure why that error is happening again, same thing happens for Spice on Travis as well. But all the tests for week are passing locally. Feel free to verify and merge :)

moollaza:
Should be fixed now, also added "when" to triggers so we can handle grammatically correct questions for future dates

gsquire:
@moollaza Thanks for seeing this through. I have gotten pretty busy with school. Great job! :)

MrChrisW:
@moollaza Tested and working! :+1:
 One last small thing ..... Should it be displaying dates in the future `5th week of 2018` with a query like "what **was** the" ?

jagtalon:
@moollaza In the case where we're in the current week of the year, it would still say it in the past tense:

screen shot 2015-01-30 at 3 59 24 am

CC @abeyang what do you think of the look? Pretty standard though. :)

jagtalon:
I wonder if we should add the end day too? Probably not important--I'm not the target audience of this goodie. :) Works pretty well! Fails when I put in "what was the 5th week of 0000", though:

screen shot 2015-01-30 at 4 04 49 am

jagtalon:
Also fails when I put in "what was the 0 week of 2011":

screen shot 2015-01-30 at 4 06 20 am

jagtalon:
Another one with "what was the 99th week of 2011"--sorry for all the comments!

MrChrisW:
@jagtalon Damn nice catch! image

The `$year `needs to validated (between1900 to 2099) before feeding it to [`Monday_of_Week`](https://github.com/duckduckgo/zeroclickinfo-goodies/pull/950/files#diff-5104ecc1d9a187d3513009341a0e7e8aR86) and ensure the week is between a valid range also!

jagtalon:
@MrChrisW thanks! How do you know that we need to limit it to 1900 and 2099? It looks like it's working for the year 1100 for me (and correct too)

MrChrisW:
@jagtalon I was just trying to think of a reasonable/useful range :smile: 
Turns out Date::Calc supports 1 to the largest positive integer representable on your system (at least 32767)!!!

moollaza:
@jagtalon good catch! Will fix. Verifying years sounds like another good helper function. Maybe we should add one to the dates role?
/cc @mintsoft

javathunderman:
Seems like I have called attention to this. 

Mission accomplished. I'm going to work on some other goodies that have to be released.

mintsoft:
@moollaza IIRC there's already a regex for the dates within the role; it probably just needs wrapping with a getter

javathunderman:
@moollaza Seems like this has taken a back seat- any update?

moollaza:
Sorry guys, I've been busy dealing with other new IA's and bugs -- I'll get to this eventually. If anyone else wants to finish it up, you're welcome to do so!

mintsoft:
@moollaza I'll see if I can slot it in this weeked; I'll probably supplant some year validation into the DatesRole and then use it here

javathunderman:
ping @mintsoft

mintsoft:
@javathunderman Apologies; I've been moving house so I've not really had the time! Hopefully, now I'm all done I can dig something up.

javathunderman:
@mintsoft No problem!

mintsoft:
Highjacked in: https://github.com/duckduckgo/zeroclickinfo-goodies/pull/1348

PR #849, opened 2014-12-17
Added goodie to convert alphabetic phone number to numeric phone number.

stevelippert:
# Goodie Pull Request Template

Using this template will help us better understand your Instant Answer and assist you when necessary.  Simply copy and paste the markdown below into the description of your GitHub pull request and complete as appropriate.

**What does your Instant Answer do?**
Converts alphabetic phone numbers to numeric phone numbers. (E.g., Query: '1-800-fun-hack to phone number' returns 'Phone Number: 1-800-386-4225'

**What problem does your Instant Answer solve (Why is it better than organic links)?**
Right now organic results don't convert the alphabetic phone numbers into easy to dial numeric phone numbers.

**What is the data source for your Instant Answer? (Provide a link if possible)**
Mapped phone keys between numbers and the alphabet.

**Why did you choose this data source?**
Its the definitive source

**Are there any other alternative (better) data sources?**
None that I'm aware of.

**What are some example queries that trigger this Instant Answer?**
1-800-LAWYR-UP to digits
1-800-fun-hack to phone number

**Which communities will this Instant Answer be especially useful for? (gamers, book lovers, etc)**
Users in the United States that don't want to think about converting phone numbers with words / letters into an actual phone number.

**Is this Instant Answer connected to a DuckDuckHack [Instant Answer idea](https://duck.co/ideas)?**
Yes, https://duck.co/ideas/idea/4718/letters-to-keypad-numbers-and-vice-versa

**Which existing Instant Answers will this one supercede/overlap with?**

**Are you having any problems? Do you need our help with anything?**
I've only completed the first part, going from alphabetic phone numbers to numeric phone numbers, not going from numeric phone numbers to words.

**Where did you hear about DuckDuckHack? (For first time contributors)**
Blog entry, http://blog.ackx.net/duckduckhack-accepted.html.

**What does the Instant Answer look like? (Provide a screenshot for new or updated Instant Answers)**
image
http://i.imgur.com/dbRbKDo.png

## Checklist
Please place an 'X' where appropriate.

```
[X] Added metadata and attribution information (https://duck.co/duckduckhack/metadata)
[X] Wrote test file and added to t/ directory (https://duck.co/duckduckhack/goodie_tests)
[X] Verified that instant answer adheres to design guidelines (https://duck.co/duckduckhack/design_styleguide)
[X] Verified that instant answer adheres to code styleguide (https://duck.co/duckduckhack/code_styleguide)
[X] Tested cross-browser compatibility

[...]

PR #964, opened 2015-01-29
Phonewords to Phone Numbers

moollaza:
Continuation of #849 

/cc @jagtalon @MrChrisW @stevelippert 

2015-01-29_15h37_36

### DIFF
--- a/lib/DDG/Goodie/PhoneAlphabet.pm
+++ b/lib/DDG/Goodie/PhoneAlphabet.pm
@@ -1,0 +1,63 @@
+package DDG::Goodie::PhoneAlphabet;
+# ABSTRACT: Taking a phone number with letters in it and returning the phone number
+
+use DDG::Goodie;
+
+zci answer_type => 'phone_alphabet';
+zci is_cached   => 0;
+
+name "PhoneAlphabet";
+description "Returns the phone number from a word phone number";
+primary_example_queries "1-800-FUN-HACK to digits", "1-800-LAWYR-UP to phone number";
+category 'reference';
+topics 'special_interest';
+code_url "https://github.com/duckduckgo/zeroclickinfo-goodies/blob/master/lib/DDG/Goodie/PhoneAlphabet.pm";
+attribution github => ["https://github.com/stevelippert", "Steve Lippert"],
+            twitter => ["https://twitter.com/stevelippert", "stevelippert"];
+
+# Triggers
+triggers any => 'to digit', 'to digits', 'to phone', 'to phone number';
+
+sub components {
+    my %phoneAlphabet = (
+      a => "2",
+      b => "2",
+      c => "2",
+      d => "3",
+      e => "3",
+      f => "3",
+      g => "4",
+      h => "4",
+      i => "4",
+      j => "5",
+      k => "5",
+      l => "5",
+      m => "6",
+      n => "6",
+      o => "6",
+      p => "7",
+      q => "7",
+      r => "7",
+      s => "7",
+      t => "8",
+      u => "8",
+      v => "8",
+      w => "9",
+      x => "9",
+      y => "9",
+      z => "9"
+    );
+
+    my @components = map { defined $phoneAlphabet{$_} ? ($phoneAlphabet{$_}) : ($_) } split //, $_;
+    return join("", @components);
+}
+
+handle remainder => sub {
+    return unless $_;
+    $_ = lc;
+    my @words = split(//, $_);
+    my @numbers = map { components($_) } @words;
+    return "Phone Number: " . join("", @numbers);
+};
+
+1;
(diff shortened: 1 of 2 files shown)


moollaza:
My only gripe about this is that it doesn't properly format the output number. It preserver the original formatting. I attempted to normalize the numbers, but it was going to be a decent amount of work (lots of edge cases and intl. formats) so I've skipped it for now. Can ask the community to tackle that. This could possible evolve into a "phone number formatter" IA if we want.

lib/DDG/Goodie/PhoneAlphabet.pm
```diff
@@ -0,0 +1,40 @@
...
+            twitter => ["https://twitter.com/stevelippert", "stevelippert"];
+
+triggers any => 'to digit', 'to digits', 'to phone', 'to phone number';
+
+handle remainder => sub {
+    # Return unless it looks like a phone number
+    return unless ($_ =~ /[-0-9A-Za-z]{6,15}$/);
+    # Return if it's a Base64 number
+    return if ($_ =~ /^0x\d+$/);
+    return unless $_;
```
mintsoft: I think this is unnecessary? doesn't L23 handle this?

lib/DDG/Goodie/PhoneAlphabet.pm
```diff
@@ -0,0 +1,40 @@
...
+    # Lower case everything.
+    my $num = lc $_;
+    # Use a regex to replace each letter with the corresponding number from the phone key pad.
+    $num =~ tr/abcdefghijklmnopqrstuvwxyz/22233344455566677778889999/;
+
+    return "Phone Number: $num",
+    structured_answer => {
+        input     => ["$_"],
```
mintsoft: Is this html_enc'd by duckpan?

lib/DDG/Goodie/PhoneAlphabet.pm
```diff
@@ -0,0 +1,40 @@
...
+code_url "https://github.com/duckduckgo/zeroclickinfo-goodies/blob/master/lib/DDG/Goodie/PhoneAlphabet.pm";
+attribution github => ["https://github.com/stevelippert", "Steve Lippert"],
+            twitter => ["https://twitter.com/stevelippert", "stevelippert"];
+
+triggers any => 'to digit', 'to digits', 'to phone', 'to phone number';
+
+handle remainder => sub {
+    # Return unless it looks like a phone number
+    return unless ($_ =~ /[-0-9A-Za-z]{6,15}$/);
+    # Return if it's a Base64 number
```
mintsoft: Hex number surely?

mintsoft:
@moollaza I was going to pick this up this weekend but you've beaten me to the punch!

lib/DDG/Goodie/PhoneAlphabet.pm
moollaza: Haha, definitely didn't sleep enough last night. You're right!

lib/DDG/Goodie/PhoneAlphabet.pm
moollaza: Good catch, I'm not actually sure, but I think it is

moollaza:
@mintsoft if you want to take it further you're welcome to. I'll fixup the stuff you mentioned, but you can pick it up after if you'd like.

lib/DDG/Goodie/PhoneAlphabet.pm
moollaza: Turns out it isn't. I've added a commit to encode it.

moollaza: I wonder if we should add a param to force encoding:

```perl
structured_answer => {
    encode_output => 1,
    input => ["..."],
    operation => "...",
    answer => "..."
}
```

jagtalon:
@moollaza nice! @abeyang https://ddh1.duckduckgo.com/?q=1-800-GOT-PUNK+to+digits&ia=answer

abeyang:
@moollaza @jagtalon looks good. Remind me -- we can't just get rid of the "Phone number:" in the byline?

MrChrisW:
@abeyang We can remove the `operation` value that returns "Phone number" quite easily I think.

jagtalon:
@abeyang so you'd rather have them removed?

moollaza:
@abeyang I like having an operation so it's clear how we interpreted your query and what it was we did. Would a different `operation` make more sense? Maybe "Digitize" or "Convert"?

chrismorast:
I think it should stay in the byline.

moollaza:
Alright then, looks like this is good to merge. Thanks @mintsoft @jagtalon @abeyang @chrismorast @MrChrisW

lib/DDG/Goodie/PhoneAlphabet.pm
mintsoft: I think we probably should do something similiar, however we need to balance XSS against HTML output for each goodie somehow

moollaza:
@MrChrisW @stevelippert congrats! It's live: https://duckduckgo.com/?q=1-800-GOT-JUNK+to+digits&ia=answer

Issue #969, opened 2015-02-01
WorkdaysBetween: Tests failing 


MrChrisW:
The following test is currently failing in [t/WorkdaysBetween.t](https://github.com/duckduckgo/zeroclickinfo-goodies/blob/master/t/WorkdaysBetween.t)

Query: **business days between january and february**
````shell
Failed test 'Regexp: answer'
'There are 231 Workdays between 01 Feb 2015 and 01 Jan 2016.'
doesn't match '(?^:There are [1-9][0-9] Workdays between 01 Jan [0-9]{4} and 01 Feb [0-9]{4}\.)'
````

Looks like it's switching Jan and Feb around in the result.

mintsoft:
@MrChrisW I think because it's now february "january" is treated as 2016 (`'There are 231 Workdays between 01 Feb 2015 and 01 Jan 2016.'`) so there are 3 digits in the number of workdays and it doesn't match.

The test is easily fixable (will do shortly); however the decision was made to treat naked month names as the next occurrence unless it's the current month; is this counter intuitive in general or does it just not work as expected in this case?

MrChrisW:
@mintsoft That makes sense, thanks for fixing. 

>however the decision was made to treat naked month names as the next occurrence unless it's the current month; is this counter intuitive in general or does it just not work as expected in this case? 

That's a good question. I think it may be a little counter intuitive or against common sense. 

Personally I'd expect the query `business days between january february` to result in `20 - Workdays between 01 Jan 2015 ‐ 01 Feb 2015`, provided the current year is 2015. 

It's confusing that it results in `between 01 Feb 2015 ‐ 01 Jan 2016` as I believe I'm searching for **between january to february** not **between february to january** 

Anyway I think this might be tolerated for now :+1:

mintsoft:
@MrChrisW I absolutely agree with you in this context.

I think I'll examine how the role is handling month names this evening. I think more generally, if the name of the month you're on is the same as the month that you're asking for then the current month is what should be used. Apparently that's not what it's doing!

moollaza:
> It's confusing that it results in between 01 Feb 2015 ‐ 01 Jan 2016 as I believe I'm searching for between january to february not between february to january

Totally agree on that! I think if the year is ambiguous, we should assume they mean for the current year, if the months given are in order of the calendar.

[...]

PR #970, opened 2015-02-01
WorkdaysBetween: Adding MockTime

mintsoft:
'Fixing' https://github.com/duckduckgo/zeroclickinfo-goodies/issues/969 with MockTime; there is still the question as to whether this is counter intuitive


moollaza:
LGTM :+1:

moollaza:
@mintsoft should we make a new issue so we remember to use a proper fix?

Issue #971, opened 2015-02-01
DatesRole has suspicious behaviour with naked months


mintsoft:
It appears that "February" when in February is being parsed as next year rather than the current year

https://github.com/duckduckgo/zeroclickinfo-goodies/issues/969
https://github.com/duckduckgo/zeroclickinfo-goodies/pull/970#issuecomment-72366718

mintsoft:
I've just looked through the DatesRole and it's exactly "to spec" as it were. I.e. it behaves as was agreed at the time and this does make sense. However, I'm thinking that the answer to the issue raised by the tests failing here: https://github.com/duckduckgo/zeroclickinfo-goodies/issues/969 might be to permit the consumer of the role to provide a relative date for ambiguous matches. 

I.e. there are 2 datestrings in the query, if the first datestring is parsed as-is currently and then the resulting DateTime object is used as the base for the second vague_datestring parsing then the result will be more consistent and inline with user expectations I think. Thoughts? @moollaza @MrChrisW

mintsoft:
Should be resolved by: https://github.com/duckduckgo/zeroclickinfo-goodies/pull/972

mintsoft:
@moollaza @MrChrisW closing this as it should be fixed by #972

PR #972, opened 2015-02-01
Relative ambiguous date handling improvements

mintsoft:
Tweaking behaviour around parse_all_datestrings_to_date so that dates are relative to the first one:

https://github.com/duckduckgo/zeroclickinfo-goodies/issues/969
https://github.com/duckduckgo/zeroclickinfo-goodies/issues/971

cc @MrChrisW @moollaza


MrChrisW:
Sorry for the delay @mintsoft, This looks good.

Just so I have an understanding of what's going on, can you explain the conditional statement in [line 400](https://github.com/duckduckgo/zeroclickinfo-goodies/blob/mintsoft/datesrole-months/lib/DDG/GoodieRole/Dates.pm#L400-L403) It looks like you're now passing a `base_time` across to `parse_descriptive_datestring_to_date` and using that if defined over `DateTime->now` 

As I haven't played around much with the roles I'd just like to start to get an understanding :smile:

mintsoft:
@MrChrisW Sure; this is essentially to handle descriptive string queries like "tomorrow" or "january". Originally; it used $now as the reference (or base_time), so that if you say "january" when it's actually february then it thinks you mean next year; when we (@mwmiller and myself) were writing it previously it seemed to make the most sense.

Now it cascades the relative point, so `qw(january february)` would get handled as february relative to january which is itself relative to `$now`. The actual logic exists here in parse_descriptive_datestring_to_date: https://github.com/duckduckgo/zeroclickinfo-goodies/pull/972/files#diff-6b357409d5744d2f8714cc73e5e834a3R441 however I've essentially allowed the consumer of the role to inject a base value

MrChrisW:
@mintsoft I appreciate the long explanation actual, pretty sure I have an better understanding now :wink: 

>Now it cascades the relative point, so `qw(january february)` would get handled as february relative to january which is itself relative to $now  The actual logic exists here in parse_descriptive_datestring_to_date

That makes sense. Use $now as the reference/base time unless `$base` was passed to `parse_datestring_to_date($date, $dates_to_return[0])` then use that. As you said in your second edit.

>Edit: That's a little longer than I was planning for a short description, let me know if my waffle is unclear

Your waffle is clear :laughing: I'd prefer a longer description any day if it helps me understand the function/code. :+1:

mintsoft:
@MrChrisW It's just a shame that it wasn't clear from the code; I might swing back and see if we can clean it up a little in another PR

moollaza:
@mintsoft @MrChrisW LGTM :+1:

PR #981, opened 2015-02-04
Frequency to Wavelength instant answer

nebulous:

Calculates the wavelength in meters/cm/mm for a given frequency expressed in Hz/kHz/MHz/Ghz/THz and velocity factor as a ratio of the speed of light.
Answers and/or tests could potentially be rounded to account for differences in architecture float precision

Wavelength can be directly calculated, so any organic link would be unnecessary indirection.

    wavelength 144.39Mhz
    λ 2.4GHz

Amateur radio enthusiasts
RF engineers
RTL-SDR or SatNogs hackers

No. Created to scratch my own itch for this feature.

**Which existing Instant Answers will this one supersede/overlap with?**

None. Though with a bit of pressure, could potentially be forced into conversions.

No.

**What are the terms of use for the API? Will DuckDuckGo need specific authorization (e.g. an API key)? Are there any costs associated with API usage?**

Calculation. free to all.

I use DDG, heart perl, and wanted to see how I could add a feature I wanted.

Standard structured text answer.
<a href="http://imgur.com/Ve3lBtJ"></a>

## Checklist

```
[X] Added metadata and attribution information
[X] Wrote test file and added to t/ directory
[X] Verified that instant answer adheres to design guidelines (https://duck.co/duckduckhack/design_styleguide)
[] Verified that instant answer adheres to code styleguide (https://duck.co/duckduckhack/code_styleguide)
[X] Tested cross-browser compatibility

    - Windows 8
        [] Google Chrome
        [] Firefox
        [] Opera
        [] IE 10

    - Windows 7
        [] Google Chrome
        [] Firefox
        [] Opera
        [] IE 8
        [] IE 9
        [] IE 10

    - Windows XP
        [] IE 8

    - Mac OSX
        [X] Google Chrome
        [X] Firefox
        [] Opera
        [X] Safari

    - iOS (iPhone)
        [] Safari Mobile
        [] Google Chrome
        [] Opera

    - iOS (iPad)
        [] Safari Mobile
        [] Google Chrome
        [] Opera

    - Android
        [] Firefox
        [] Native Browser
        [] Google Chrome
        [] Opera
```

### DIFF
--- a/lib/DDG/Goodie/Wavelength.pm
+++ b/lib/DDG/Goodie/Wavelength.pm
@@ -26,7 +26,7 @@ zci is_cached   => 1;
 # Metadata.  See https://duck.co/duckduckhack/metadata for help in filling out this section.
 name "Wavelength";
 description "Frequency to Wavelength translation";
-primary_example_queries "λ 2.4GHz", "144.39 MHz wavelength","lambda 1500kHz";
+primary_example_queries "λ 2.4GHz", "144.39 MHz wavelength","lambda 1500kHz", "1Mhz lambda VF=0.85";
 category "physical_properties";
 topics "math", "science", "special_interest";
 code_url "https://github.com/nebulous/zeroclickinfo-goodies/blob/master/lib/DDG/Goodie/Wavelength.pm";
@@ -39,11 +39,15 @@ triggers any => "λ", "wavelength", "lambda";
 
 # Handle statement
 handle remainder => sub {
-    my ($freq,$units) = m/([\d\.]+)\s*((k|M|G|T)?hz)/i;
+    my ($query) = @_;
+    my ($freq,$units) = $query =~ m/([\d\.]+)\s*((k|M|G|T)?hz)/i;
     return unless $freq and $units;
 
+    my ($vf) = $query =~ m/\bvf[ =]([\d\.]+)/i;
+    $vf = 1 if (!$vf or $vf>1 or $vf<0);
+
     my $mul     = MULTIPLIER->{lc($units)};
-    my $hz_freq = $freq * $mul;
+    my $hz_freq = $freq * $mul * (1/$vf);
 
     my $output_value = (SPEED_OF_LIGHT / $hz_freq);
     my $output_units = 'Meters';
@@ -64,7 +68,7 @@ handle remainder => sub {
     return $output_text,
         structured_answer => {
             input     => [$freq, FORMAT_UNITS->{lc($units)}],
-            operation => "Wavelength",
+            operation => "Wavelength ( Velocity Factor $vf × Speed of light in a vacuum )",
             result    => $output_text,
         };
 };
--- a/lib/DDG/Goodie/Wavelength.pm
+++ b/lib/DDG/Goodie/Wavelength.pm
@@ -1,0 +1,58 @@
+package DDG::Goodie::Wavelength;
+use utf8;
+# ABSTRACT: Frequency to wavelength
+
+use DDG::Goodie;
+
+use constant SPEED_OF_LIGHT => 299792458; #meters/second
+use constant MULTIPLIER => {
+     hz => 1,
+    khz => 10**3,
+    mhz => 10**6,
+    ghz => 10**9,
+    thz => 10**12
+};
+
+zci answer_type => "wavelength";
+zci is_cached   => 1;
+
+# Metadata.  See https://duck.co/duckduckhack/metadata for help in filling out this section.
+name "Wavelength";
+description "Frequency to Wavelength translation";
+primary_example_queries "λ 2.4GHz", "144.39 MHz wavelength","lambda 1500kHz";
+category "physical_properties";
+topics "math", "science", "special_interest";
+code_url "https://github.com/nebulous/zeroclickinfo-goodies/blob/master/lib/DDG/Goodie/Wavelength.pm";
+attribution  github => ["nebulous", "John Lifsey"],
+               cpan => "nebulous",
+            twitter => "aresweet";
+
+# Triggers
+triggers any => "λ", "wavelength", "lambda";
+
+# Handle statement
+handle remainder => sub {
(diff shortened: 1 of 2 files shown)


lib/DDG/Goodie/Wavelength.pm
```diff
@@ -0,0 +1,72 @@
+package DDG::Goodie::Wavelength;
+use utf8;
+# ABSTRACT: Frequency to wavelength
+
+use DDG::Goodie;
+
+use constant SPEED_OF_LIGHT => 299792458; #meters/second
+use constant MULTIPLIER => {
```
mintsoft: Is it worth extending this out for mHz, uHz, nHz ?

mintsoft:
@nebulous Thanks for this! It looks good! 

The only thing that jumps out is that I think the output should explicitly say that it's for light in a vacuum rather than any oscillator, what'd you think?

lib/DDG/Goodie/Wavelength.pm
nebulous: Initially I was going to add all of (https://en.wikipedia.org/wiki/Hertz#SI_multiples) but thought it more likely that users would query mhz or mHz for common 10^6 Hz queries than 10^-3 queries. Visions of features/conversions/heterodyne calculations then danced in my head, but thought simplicity in the most common set might be preferable. Open to suggestions of how to avoid ambiguity there.

nebulous:
@mintsoft sure, I think *operation* could be "Wavelength based on speed of light in a vacuum" instead of "wavelength" but it's possible that could add confusion if we don't have additional options for velocity factor, at which point things are looking like antenna calculators.... which brings me to my next idea for DDG :)

Happy to patch with the above, or with some more clever wording/placement as directed.

lib/DDG/Goodie/Wavelength.pm
mintsoft: Hmm this is true, if the expected use case is always related to radio bands then there's limited use of anything with f<1; even ELF and SLF are in the single Hz ranges; so it's probably not worth it.
In general, as long as there's a unambiguous representation of how the input was parsed in the output then the user has less of the "wtf that's not what I meant!" reaction.

mintsoft:
@nebulous maybe the option of allowing the user to supply the velocity could work well, but default to 299792458, and provide in the output "using velocity of $v" ?

nebulous:
@mintsoft Added Velocity Factor with default of 1x speed-of-light in 76e5d1d

user may optionally specify VF between 0 and 1, with out of bounds queries forced to 1

mintsoft:
@nebulous Nice, looks good. @jagtalon @moollaza any input?

moollaza:
Apologies for the delay! Will take a looks soon.

moollaza:
I know it's a little late to ask but @nebulous, can you please fill out the [pull request template](https://github.com/duckduckgo/zeroclickinfo-spice/blob/master/pull_request_template_spice.md) and edit your initial comment.

It's standard protocol for IA development so we can all get a better idea about this IA, how it works and what it looks like

Thanks!

nebulous:
@moollaza: np, updated.

jagtalon:
I think it looks pretty good. @abeyang https://ddh9.duckduckgo.com/?q=144.39+MHz+wavelength&ia=answer

abeyang:
Can we simplify the byline? Something more like:

> Wavelength (in meters): 144.39 MHz

jagtalon:
@abeyang the Velocity Factor could be changed (apologies for not pointing that out first): https://ddh9.duckduckgo.com/?q=1Mhz+lambda+VF%3D0.85&ia=answer so the velocity factor can be important. @nebulous @mintsoft I wonder if it could be simplified as "in a vacuum" though, but I can't say because I'm not familiar with how people will actually use this.

nebulous:
The intended use-case was for RF, which tends to be either in a vacuum or in some kind of medium/waveguide. For instance, a common case might be "wavelength 2.4GHz vf=0.8" to approximate an ISM-band wave traveling down RG6 coaxial cable with a polyethylene dielectric(about 10cm at vf 0.8, but over 12cm in a vacuum).

abeyang:
@nebulous @jagtalon Would this suffice then:

> Wavelength for 2.4GHz (Velocity Factor of 0.5)

nebulous:
@abeyang @jagtalon Maybe a combination of both of your suggestions could work:
The most common case will be default VF=1 in which case we can just say:

    Wavelength for 2.4GHz (in a vacuum)

but if vf is specified say:

    Wavelength for 2.4GHz (Velocity Factor of 0.5)

abeyang:
I'm cool with that! :+1:

mintsoft:
@nebulous seems sensible; however I do feel it should state "light" or "em radiation" somewhere; especially with a trigger like `wavelength`

jagtalon:
@nebulous seems good to me too. @mintsoft @nebulous @abeyang would it make sense if we showed `Wavelength for 2.4GHz (Speed of light in a vacuum)`?

mintsoft:
@jagtalon looks good, what about when a VF is set? Something like: `Wavelength for 2.4GHz (0.75  x Speed of light in a vacuum)`  ?

nebulous:
Ok, I'll see if I can round up some tuits this evening to output:

where VF=1

    Wavelength for 123MHz (speed of light in a vacuum)

where VF=0.87

    Wavelength for 123Mhz (0.87 × speed of light in a vacuum)

That should allow eventual enhancement later on to include other waves (sound pressure in standard atmosphere of ~ 330 meters/second etc) unambiguously in the future if we so choose.

abeyang:
@mintsoft @nebulous It's still a little verbose for my taste, but I suppose it's necessary. :+1:

moollaza:
@nebulous congrats! It's live: https://duckduckgo.com/?q=wavelength+144.39Mhz&ia=answer

PR #985, opened 2015-02-06
More checklist kind of feel.

hemanth:
More checklist like feel.

//cc @MrChrisW


moollaza:
@hemanth thanks. We're actually going to roll out a revised/updated checklist soon but this doesn't hurt.

hemanth:
:+1:

PR #989, opened 2015-02-07
Improve Table CSS for mobile displays

moollaza:
/cc @abeyang @mintsoft @stefolof

Updated the CSS for slightly improved viewing on mobile screens.

**\< 400px**

Before:
less than 400px before

After:
less than 400px after

------

**\> 400px**

Before:

more than 400px before

After:

more than 400px after


jagtalon:
@moollaza looks good :+1:

Issue #990, opened 2015-02-07
PhoneAlphabet: Should also trigger on "to numbers"


moollaza:
https://github.com/duckduckgo/zeroclickinfo-goodies/blob/master/lib/DDG/Goodie/PhoneAlphabet.pm#L19

Works: https://duckduckgo.com/?q=1-800-GOT-JUNK+to+digits&ia=answer
Should work: https://duckduckgo.com/?q=1-800-GOT-JUNK+to+numbers&ia=answer

MrChrisW:
Fixed by @samph in #991

PR #991, opened 2015-02-07
PhoneAlphabet: Also triggers on to numbers

samph:
Addresses PhoneAlphabet: Should also trigger on "to numbers" #990


moollaza:
@samph nicely done! Thanks a lot :+1:

PR #846, opened 2014-12-16
First Goodie returns pi

jmvbxx:
**What does your Instant Answer do?**

This is my first Goodie and I wanted it to return 'pi' to user-specified number of decimal places.

**What problem does your Instant Answer solve (Why is it better than organic links)?**

I wrote an Instant Answer for this earlier today.

**What is the data source for your Instant Answer? (Provide a link if possible)**

For now pi comes from Math::Trig which only provides limited number of decimal places.

**Why did you choose this data source?**

For simplicity.

**Are there any other alternative (better) data sources?**

We could calculate pi right in the code.

**What are some example queries that trigger this Instant Answer?**

pi 2 => 3.14
pi 5 => 3.14159

**Which communities will this Instant Answer be especially useful for? (gamers, book lovers, etc)**

Students

**Is this Instant Answer connected to a DuckDuckHack [Instant Answer idea](https://duck.co/ideas)?**

Yes, I wrote an Instant Answer earlier today.

**Which existing Instant Answers will this one supercede/overlap with?**

N/A

**Are you having any problems? Do you need our help with anything?**

The code is extremely basic.  I have not yet done the pi.t and I have to put checks into pi.pm.

**Where did you hear about DuckDuckHack? (For first time contributors)**

N/A

**What does the Instant Answer look like? (Provide a screenshot for new or updated Instant Answers)**

img-20141216-wa0003

## Checklist
Please place an 'X' where appropriate.

```
[] Added metadata and attribution information (https://duck.co/duckduckhack/metadata)
[] Wrote test file and added to t/ directory (https://duck.co/duckduckhack/goodie_tests)
[] Verified that instant answer adheres to design guidelines (https://duck.co/duckduckhack/design_styleguide)
[] Verified that instant answer adheres to code styleguide (https://duck.co/duckduckhack/code_styleguide)
[] Tested cross-browser compatibility

    Please let us know which browsers/devices you've tested on:
    - Windows 8
        [] Google Chrome
        [] Firefox
        [] Opera
        [] IE 10

    - Windows 7
        [] Google Chrome
        [] Firefox
        [] Opera
        [] IE 8
        [] IE 9
        [] IE 10

    - Windows XP
        [] IE 7
        [] IE 8

    - Mac OSX
        [] Google Chrome
        [] Firefox
        [] Opera
        [] Safari

    - iOS (iPhone)
        [] Safari Mobile
        [] Google Chrome
        [] Opera

    - iOS (iPad)
        [] Safari Mobile
[...]

PR #997, opened 2015-02-10
Continue Pi

puskin94:
Hi @mintsoft and @moollaza 
I'm trying to continue the [Pi IA](https://github.com/duckduckgo/zeroclickinfo-goodies/pull/846).
Now the situation is this:
situa
The layout breaks the number... How can I fix it?


mintsoft:
@puskin94 Hmm, nothing jumps out immediately except that `$PI1` etc should probably explicitly be strings? (Wrap in `"` or `'`)

puskin94:
@mintsoft ops, my fault. Now it should works!

puskin94:
Added a simple Regex to be secure that it will only trigger numbers ;)

puskin94:
@mintsoft Ok I fixed all the Bugs.... The output now is switched to a structured Answer, avoiding the broke string and beautifying the answer. Updated the test file too.
Just tell me for future improvements ;)

mintsoft:
@puskin94 Cool looks good, how about expanding the query space somewhat with handling for queries like `23 digits of pi` and `pi to 88 digits`?

lib/DDG/Goodie/Pi.pm
```diff
@@ -0,0 +1,60 @@
...
+
+    my $PI1 = '3.1415926535897932384626433832795028841971693993751058209749445923078164062862089986280348253421170679';
+    my $PI2 = '8214808651328230664709384460955058223172535940812848111745028410270193852110555964462294895493038196';
+    my $PI3 = '4428810975665933446128475648233786783165271201909145648566923460348610454326648213393607260249141273';
+    my $PI4 = '724587006606315588174881520920962829254091715364367892590360011330530548820466521384146951941511609';
+
+    my $answer;
+    my $num;
+
+    if (($decimal  > 0) && ($decimal < 400)) {
```
mintsoft: You could reduce the nesting here by doing this instead:
```perl
return unless $decimal  > 0 && $decimal < 400;
```

lib/DDG/Goodie/Pi.pm
```diff
@@ -0,0 +1,60 @@
...
+# Handle statement
+handle remainder => sub {
+
+    my $decimal = $_;
+
+    return unless $decimal && $decimal =~ /^(\d+)$/;
+
+    # $PI has 399 decimals
+
+    my $PI1 = '3.1415926535897932384626433832795028841971693993751058209749445923078164062862089986280348253421170679';
```
mintsoft: I'd probably move these outside of the handle so they don't need to be re-declared on each execution; also is there any reason why it's in 4 variables rather than one? It'd simplify the code below too if it's in one string

lib/DDG/Goodie/Pi.pm
```diff
@@ -0,0 +1,60 @@
+package DDG::Goodie::Pi;
+# ABSTRACT: This Goodie returns 'pi' to a user-specified number of decimal places
+
+use DDG::Goodie;
+use Math::Trig;
```
mintsoft: I presume `Math::Trig` isn't actually need as this IA has it's own definition of `pi`?

puskin94:
@mintsoft Thanks for the feedback! I hope I fixed all the stuff ;)

lib/DDG/Goodie/Pi.pm
```diff
@@ -0,0 +1,46 @@
...
+
+zci answer_type => "pi";
+zci is_cached   => 1;
+
+name "Pi";
+primary_example_queries "pi 7";
+description "Ex. returns 3.1415926";
+
+code_url "https://github.com/duckduckgo/zeroclickinfo-goodies/blob/master/lib/DDG/Goodie/Pi.pm";
+attribution github => ["https://github.com/jmvbxx", "jmvbxx"];
```
mintsoft: You should add yourself here!

lib/DDG/Goodie/Pi.pm
```diff
@@ -0,0 +1,46 @@
...
+        '8214808651328230664709384460955058223172535940812848111745028410270193852110555964462294895493038196'.
+        '4428810975665933446128475648233786783165271201909145648566923460348610454326648213393607260249141273'.
+        '724587006606315588174881520920962829254091715364367892590360011330530548820466521384146951941511609';
+
+# Handle statement
+handle remainder => sub {
+
+    my $decimal = $_;
+
+    return unless $decimal && $decimal =~ /^(\d+)$/ && $decimal > 0 && $decimal < 400;
```
mintsoft: You could make this more dynamic by having a variable under `$PI` which is the number of characters and compare it to that? Means that if we add more digits then we don't need to update the `unless`

mintsoft:
@moollaza @jagtalon Think looks about ready

puskin94:
@mintsoft ahah have I the privilege to change the name ?? ;)
Done !

lib/DDG/Goodie/Pi.pm
```diff
@@ -0,0 +1,46 @@
...
+        '8214808651328230664709384460955058223172535940812848111745028410270193852110555964462294895493038196'.
+        '4428810975665933446128475648233786783165271201909145648566923460348610454326648213393607260249141273'.
+        '724587006606315588174881520920962829254091715364367892590360011330530548820466521384146951941511609';
+
+# Handle statement
+handle remainder => sub {
+
+    my $decimal = $_;
+
+    return unless $decimal && $decimal =~ /^(\d+)$/ && $decimal > 0 && $decimal < length($PI);
```
mintsoft: I'd probably define `length($PI)`  as $PI_max_digits on L25 and then used that; that way `$PI` is only iterated over once to count the number of characters

lib/DDG/Goodie/Pi.pm
```diff
@@ -0,0 +1,46 @@
...
+
+zci answer_type => "pi";
+zci is_cached   => 1;
+
+name "Pi";
+primary_example_queries "pi 7";
+description "Ex. returns 3.1415926";
+
+code_url "https://github.com/duckduckgo/zeroclickinfo-goodies/blob/master/lib/DDG/Goodie/Pi.pm";
+attribution github => ["https://github.com/puskin94", "puskin94"];
```
mintsoft: Can you add "jmvbxx" back in, you can add two github's like:
https://github.com/duckduckgo/zeroclickinfo-goodies/blob/master/lib/DDG/Goodie/Conversions.pm#L18

puskin94:
@mintsoft Thanks for the help with this code ;)

mintsoft:
@puskin94 No problem; it's what we're here for

puskin94:
@mintsoft @moollaza any news guys ??

moollaza:
@puskin94 sorry about the delay! Will review this ASAP.

puskin94:
@moollaza no problem man!! Thanks for your time ;)

lib/DDG/Goodie/Pi.pm
```diff
@@ -0,0 +1,49 @@
...
+
+    return $answer,
+    structured_answer => {
+        input     => [$decimal." digits of Pi"],
+        operation => "Calculate",
+        result    => $answer
+    };
+
+
+    return;
```
moollaza: No need for this return :) It will never be reached

lib/DDG/Goodie/Pi.pm
```diff
@@ -0,0 +1,49 @@
...
+code_url "https://github.com/duckduckgo/zeroclickinfo-goodies/blob/master/lib/DDG/Goodie/Pi.pm";
+attribution github => ["https://github.com/puskin94", "puskin94"],
+            github => ["https://github.com/jmvbxx", "jmvbxx"];
+
+
+triggers startend => "pi", "digits of pi";
+
+# $PI has 399 decimals
+
+my $PI = '3.1415926535897932384626433832795028841971693993751058209749445923078164062862089986280348253421170679'.
```
moollaza: Once we have Goodie templates, I think it would be cool to use a much larger number of pi digits (http://www.eveandersson.com/pi/digits/) like 1 million! And then we can easily supress the ouput from overflowing the page

puskin94:
@moollaza thanks for the review!! I think that tomorrow evening you will receive the final PR ready to be merged :)

moollaza:
@puskin94 all that needs to be done is removing that return statement. I can do it if you'd like after merge. Its' not going to break anything :+1: 

@abeyang http://moollaza.duckduckgo.com:5000/?q=pi+150

Standard Goodie structured answer template :)

moollaza:
@puskin94 all that needs to be done is removing that return statement. I can do it if you'd like after merge. Its' not going to break anything :+1: 

@abeyang http://moollaza.duckduckgo.com:5000/?q=pi+150

Standard Goodie structured answer template :)

moollaza:
@puskin94 ok @abeyang would like a couple more updates for the UI:

- Change the `operation` to "First ### Digits of Pi", remove the `input`
- Reduce the font size of the numbers down to `1.32em`, you'll need to make a CSS file

puskin94:
@moollaza hei, I can't do anything until this night! There wold be no problem for the changes to the code, but this is the first time that I add a css file to a Goodie... I hope It will be good :)

mintsoft:
@puskin94 Check the docs if you need any help: https://duck.co/duckduckhack/goodie_advanced_handle_functions#advanced-handle-functions

Check the "Using Files in the Share Directory"

puskin94:
@mintsoft thanks man! :+1:

mintsoft:
@puskin94 Oops, sorry I meant: https://duck.co/duckduckhack/goodie_advanced_handle_functions#stylesheets

puskin94:
@mintsoft @moollaza I think I fixed all the stuff ;) Easier than expected ;)

moollaza:
@puskin94 sorry I think you misunderstood me, I mean we should do the million digits later, once we have templates for Goodies, so we could allow someone to request `1000 digits of pi` but it wouldn't fill up the page like it now will. Imagine if they asked for all 1,000,000 -- you'll never see the results on the Search Page!

Also, putting a massive number like that directly into the file isn't the best way to handle it. We should have it in a separate file so that editing the Goodie itself doesn't require scroll through so many rows.

Sorry about the confusion, but for now please revert the change to 1 million digits.

puskin94:
@moollaza I've done the revert! Now $PI has 399 digits ;)

moollaza:
@puskin94 awesome LGTM :+1: 

Design team agrees -- thanks @abeyang !

moollaza:
@puskin94 it's live! https://duckduckgo.com/?q=24+digits+of+pi&ia=answer

moollaza:
@puskin94 we now have full template support! Let's convert this one to use the `Info` template so that we can allow for larger amounts of Pi. The template will prevent the number for filling up the entire page though :)

puskin94:
@moollaza sure! You should see a new PR in these days ;)

PR #1001, opened 2015-02-12
add metadata to repo

jdorweiler:
get metadata from community platform and add to `share/goodie/meta`
@russellholt


russellholt:
:+1:

PR #1003, opened 2015-02-12
Issue#958  Lorem Ipsum generator 

jee1mr:


lib/DDG/Goodie/LoremIpsum.pm
```diff
@@ -0,0 +1,26 @@
...
+code_url "https://github.com/duckduckgo/zeroclickinfo-goodies/blob/master/lib/DDG/Goodie/LoremIpsum.pm";
+attribution github => ["https://github.com/jee1mr", "Jeevan M R"],
+            twitter => ["https://twitter.com/jee1mr","Jeevan M R"];
+
+
+triggers any => "lorem ipsum", "lipsum";
+
+handle remainder => sub {
+    
+    my $text = Text::Lorem->new();
```
mintsoft: Can you move this out of the `handle remainder` so it's only executed once?

lib/DDG/Goodie/LoremIpsum.pm
```diff
@@ -0,0 +1,26 @@
...
+    return $text->paragraphs($_) if $_ && $_ =~ /^\d{10}$/;
```
mintsoft: `9999999999` seems like a little extreme as a max number of paragraphs for this! Maybe 10 is more appropriate?

t/LoremIpsum.t
```diff
@@ -0,0 +1,15 @@
...
+use warnings;
+use Test::More;
+use DDG::Test::Goodie;
+
+zci answer_type => "lorem_ipsum";
+zci is_cached   => 1;
+
+ddg_goodie_test(
+    [qw( DDG::Goodie::LoremIpsum )]
+);
```
mintsoft: Can you add some tests for this like: https://duck.co/duckduckhack/goodie_tests

As the output from this is presumably random; you'll have to use regex on the strings rather than exact hits

mintsoft:
@jee1mr Thanks for this; looks like it'll be good replacement for the closed source version! A couple of things:

* Can you add the completed PR template into the PR description
* There appears to be conflicts preventing a merge of this, can you rebase your fork?

PR #1010, opened 2015-02-14
Conversions: Tweaks to tests around rounding and accuracy

mintsoft:
The upgrade has broken the tests for Conversions due to small changes in rounding, this fixes it! 

Unfortunately currently there isn't a good workflow that can be used for this as it's quite awkward to test the accuracy of Conversions based on an unreleased module and it's difficult to coordinate the module being updated and the tests at the same time.

cc @moollaza @MrChrisW

### DIFF
--- a/t/Conversions.t
+++ b/t/Conversions.t
@@ -191,11 +191,11 @@ ddg_goodie_test(
         }
     ),
     '500 miles in metres' => test_zci(
-        '500 miles = 804,672.249 meters',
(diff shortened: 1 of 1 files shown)


mintsoft:
@abeyang do you have some sort of trigger that's adding `in progress` to everything?

moollaza:
LGTM :+1:

moollaza:
thanks a lot @mintsoft

PR #1018, opened 2015-02-22
Whacking use strict into all goodies

mintsoft:
This should force linguist to detect them as Perl 5 rather than 6 which will also fix the highlighting.

ref: https://github.com/github/linguist/issues/2149

cc @moollaza @jagtalon


moollaza:
LGTM :+1:

PR #789, opened 2014-11-26
Simple encryption service Goodie

diegojuan:
**What does your Instant Answer do?**
Provides a simple crypto service to encrypt and decrypt a string using the RC4 algorithm.

**What problem does your Instant Answer solve (Why is it better than organic links)?**
Provides way to pass sensitive info from one user to another just using their browser and a preshared key.

**What are some example queries that trigger this Instant Answer?**

crypt encrypt key string
crypt decrypt key string
crypt en key string
crypt de key string

**Which communities will this Instant Answer be especially useful for? (gamers, book lovers, etc)**
Mostly sysadmins and people that need to pass sensitive information around, like passwords or credit card numbers.

**Is this Instant Answer connected to a DuckDuckHack [Instant Answer idea](https://duck.co/ideas)?**
Yes, https://duck.co/ideas/idea/1463/simple-encryption-service

**Which existing Instant Answers will this one supercede/overlap with?**
None that I know.

**Are you having any problems? Do you need our help with anything?**
Not so far, maybe after more testing problems can arise :)

**Where did you hear about DuckDuckHack? (For first time contributors)**
On the news, about 2 years ago.

**What does the Instant Answer look like? (Provide a screenshot for new or updated Instant Answers)**

![screen shot 2014-11-26 at 6 51 29 p m]
(https://cloud.githubusercontent.com/assets/4934646/5210590/b5b3c5b4-759d-11e4-9166-4c49932f5adf.png)

## Checklist
Please place an 'X' where appropriate.

```
[X] Added metadata and attribution information (https://duck.co/duckduckhack/metadata)
[X] Wrote test file and added to t/ directory (https://duck.co/duckduckhack/goodie_tests)
[X] Verified that instant answer adheres to design guidelines (https://duck.co/duckduckhack/design_styleguide)
[X] Verified that instant answer adheres to code styleguide (https://duck.co/duckduckhack/code_styleguide)
[X] Tested cross-browser compatibility

    Please let us know which browsers/devices you've tested on:
    - Windows 8
        [] Google Chrome
        [] Firefox
        [] Opera
        [] IE 10

    - Windows 7
        [] Google Chrome
        [] Firefox
        [] Opera
        [] IE 8
        [] IE 9
        [] IE 10

    - Windows XP
        [] IE 7
        [] IE 8

    - Mac OSX
        [X] Google Chrome
        [X] Firefox
        [] Opera
        [X] Safari

    - iOS (iPhone)
        [] Safari Mobile
        [] Google Chrome
[...]

PR #1021, opened 2015-02-23
Continue Rc4 (En|De)cryption

puskin94:
Hei, I'm continuing this IA: https://github.com/duckduckgo/zeroclickinfo-goodies/pull/789

I think I fixed and committed all the suggestions. 
@moollaza I didn't take this advice:
moo
I preferred to use the 'rc4' trigger and check for the presence of $operation, $key && $plaintext in the return unless.


lib/DDG/Goodie/Rc4.pm
```diff
@@ -0,0 +1,66 @@
...
+                result    => $result
+            };
+
+    } elsif ($type eq "decrypt" || $type eq "de" || $type eq "dec") {
+            #To decrypt we do the reverse process, we take the plaintext, transform it using decode_base64()
+            my $decoded = decode_base64($plaintext);
+            # Then we pass it to the RC4 funcion to be decrypted.
+            my $result = RC4($key, $decoded);
+            # No need to encode again, this result is show as is.
+            return $result,
```
moollaza: Let's simplify the `if/else` and just set the variables as needed (will need to set something like `my $operation`). Then below we can have a single `return` which means less duplicate code, and easier maintainability.

puskin94:
@moollaza Done... I was already changing my code for the "duplicate return" but I forgotten to upload the code :-1:

lib/DDG/Goodie/Rc4.pm
```diff
@@ -0,0 +1,65 @@
...
+
+handle remainder => sub { 
+
+    (my $type, my $key, my $plaintext) = split / /;
+    my $operation;
+    my $result;
+
+    return unless $type && $key && $plaintext;
+
+    if ($type eq "encrypt" || $type eq "en" || $type eq "enc") {
```
moollaza: Perl works very well with regular expressions -- we can do this check in a much more concise manner:

```perl
if ($type =~ m/en(c|crypt)?/) {
    ...
} elsif ($type =~ m/de(c|crypt)?/) {
    ...
} else {
    return;
}
```

puskin94:
@moollaza You are right, I handled all the $type(s). Added the dist.ini too

lib/DDG/Goodie/Rc4.pm
```diff
@@ -0,0 +1,66 @@
+package DDG::Goodie::Rc4;
+# ABSTRACT: This goddie provides a simple encription/decryption service
+# using RC4 algorithm and a key provided by the user.
+
+use DDG::Goodie;
+use strict;
+use warnings;
+use Crypt::RC4;
```
moollaza: This should be added to the `dist.ini` if it doesn't already exist there. Make sure to check what the currently released version is and specify that in the dist.ini as other deps have done

lib/DDG/Goodie/Rc4.pm
```diff
@@ -0,0 +1,66 @@
...
+use MIME::Base64;
```
moollaza: Same here :)

puskin94:
@moollaza I think i understand dist.ini now... But, if I'm right, there's no trace of dist.ini in the DDH docs... Future update?

moollaza:
@puskin94 we very briefly mention the dist.ini in a few places: https://github.com/duckduckgo/duckduckgo-documentation/blob/c269d5aed3cfbd24b909022b54867697badc2d17/duckduckhack/testing/testing_triggers.md

We could definitely do a better job of describing it though.

This is a great article about the "dist.ini" file: http://blog.urth.org/2010/06/walking-through-a-real-distini.html

puskin94:
@moollaza Ok ;) Thanks for the news ! :+1: I think that this IA is now ready!

lib/DDG/Goodie/Rc4.pm
```diff
@@ -0,0 +1,66 @@
...
+
+zci answer_type => "rc4";
+zci is_cached   => 1;
+
+name "DDG::Goodie::Rc4.pm";
```
moollaza: This should just be "RC4" or similar, its the name we use for displaying metadata about the IA

lib/DDG/Goodie/Rc4.pm
```diff
@@ -0,0 +1,66 @@
...
+
+handle remainder => sub { 
+
+    (my $type, my $key, my $plaintext) = split / /;
+    my $operation;
+    my $result;
+
+    return unless $type && $key && $plaintext;
+
+    if ($type =~ m/en(c|crypt)?/) {
```
moollaza: I just noticed I made a small mistake here. We need to anchor these regexes: `m/^en(c|crypt)?$/` to avoid partial matches.

lib/DDG/Goodie/Rc4.pm
```diff
@@ -0,0 +1,66 @@
...
+            $result = RC4($key, $decoded);
+            # No need to encode again, this result is show as is.
+            $operation = "Rc4 Decryption";
+    } else {
+        return;
+    }
+
+    return $result,
+        structured_answer => {
+        input     => [],
```
moollaza: We typically reiterate the input so it's clear what we understood from the query.

We should probably change the operation to `Encrypt` or `Decrypt` (remove "ion") and specify the input, and key:

So we end up with something like:
```perl
return "$operation: $plaintext, with key: $key is $result",
structured_answer => {
    operation => $operation,
    input => html_enc($plaintext) . ", Key: $key",
    result => $result
}
```

puskin94:
@moollaza Thanks man! A new PR is waiting your approval!

lib/DDG/Goodie/Rc4.pm
```diff
@@ -0,0 +1,66 @@
...
+use DDG::Goodie;
+use strict;
+use warnings;
+use Crypt::RC4;
+use MIME::Base64;
+
+zci answer_type => "rc4";
+zci is_cached   => 1;
+
+name "DDG::Goodie::RC4.pm";
```
moollaza: Sorry I mean it should just be RC4 like so: `name "RC4"`. You don't need to repeat the filename.

jagtalon:
@puskin94 checking it out!

lib/DDG/Goodie/Rc4.pm
```diff
@@ -0,0 +1,66 @@
...
+use Crypt::RC4;
+use MIME::Base64;
+
+zci answer_type => "rc4";
+zci is_cached   => 1;
+
+name "RC4.pm";
+description "Encrypt or decrypt a text using a key provided by the user";
+primary_example_queries "crypto encrypt key string", "crypto decrypt key string";
+secondary_example_queries "crypto en key string", "crypto de key string";
```
jagtalon: These example queries don't work.

jagtalon:
I feel that `rc4 de duck yWrJniG/nNg=` syntax is pretty weird. If it was `encrypt duck` it would be okay, but this just has too many fields.

@moollaza @abeyang I was thinking something like this, but we don't have any template that's close to that at all. CC @russellholt 

fullsizerender

lib/DDG/Goodie/Rc4.pm
```diff
@@ -0,0 +1,66 @@
...
+            chomp $result;
+            $operation = "Rc4 Encrypt";
+
+    } elsif ($type =~ m/^de(c|crypt)?$/) {
+            #To decrypt we do the reverse process, we take the plaintext, transform it using decode_base64()
+            my $decoded = decode_base64($plaintext);
+            # Then we pass it to the RC4 funcion to be decrypted.
+            $result = RC4($key, $decoded);
+            # No need to encode again, this result is show as is.
+            $operation = "Rc4 Decrypt";
```
jagtalon: I'd like it styled as "RC4" instead of "Rc4"

jagtalon:
@puskin94 need to update tests too to match https://github.com/duckduckgo/zeroclickinfo-goodies/commit/049707802817c8f94cc81246aef155166b767417

jagtalon:
@moollaza @abeyang https://ddh9.duckduckgo.com/?q=rc4+en+mysecretkey+hello&ia=answer

I don't think we should show the key at the bottom. What do you think?

puskin94:
@jagtalon Ops, it is my fault ;) Just updated now ;)

moollaza:
@jagtalon I had actually suggested we show it, but I'm happy either way. Figured it just makes things a little more explicit with regards to how we interpreted the query.

puskin94:
@jagtalon @moollaza should I add a js interface?? Could it looks better?

moollaza:
>should I add a js interface?? Could it looks better?

@puskin94 @jagtalon @abeyang I think we can leave the UI as-is for now and re-visit this one when we have the Goodie templates available? I don't want to complicate this one by returning HTML and/or JS

puskin94:
@moollaza No problem man!

mintsoft:
I was thinking that this would be much better if converted to the templates when available etc. That way we can actually do the encryption in JS/clientside?

moollaza:
>I was thinking that this would be much better if converted to the templates when available etc. That way we can actually do the encryption in JS/clientside?

@mintsoft I agree -- I'm seeing more and more crypto/encoding IA's and I think a single Goodie that uses JS to encode/decode/, encrypt/decrypt would be awesome. Any thoughts on a which JS crypto library we can use?

puskin94:
@moollaza @mintsoft https://code.google.com/p/crypto-js/

mintsoft:
@moollaza aside from the conversation about the future of the IA is this ready to go?

moollaza:
LGTM :+1:

moollaza:
@puskin94 congrats! It's live: https://duckduckgo.com/?q=rc4+en+mysecretkey+hello&ia=answer

Issue #1019, opened 2015-02-22
Square root of 25 answer is confusing


mintsoft:
It's not very clear! 

![image](https://cloud.githubusercontent.com/assets/978048/6320932/ca5de5c2-bae3-11e4-8abd-d1ca87e90f26.png)

Raised from: https://duck.co/forum/thread/12316/square-root-and-other-math-questions

https://duck.co/ia/view/calc_roots

moollaza:
I'm not sure why we have the extra stuff in parens, but I think we should remove it for now. It may also be simplest to use words until we have better support for displaying math (perhaps via MathJax?)

i.e. "square root of 25 = 5", "4th root of 25 = 2.23606797749979"

moollaza:
Looks like this one also needs to be converted to use a structured answer. That way the font will be larger and the parsed query can be re-iterated below as we do for the Calulator IA

kfloey:
@moollaza @mintsoft
>I'm not sure why we have the extra stuff in parens

The stuff in the parens is the square root in simplest form.  You might want to leave it for more complicated answers like the one below, and remove it for exact answers like "square root of 25."

![screenshot from 2015-02-24 14 21 54](https://cloud.githubusercontent.com/assets/9901144/6357199/8fd638b6-bc30-11e4-9f5a-90ddf7f2e99b.png)

kfloey:
I can fix this one, just need to know what you thought about my earlier statement above ^^^

moollaza:
@kfloey I think for now we can remove the simplification, because it's not very clear what it means.

We're currently investigating Goodies that are candidates for the new Goodie Templates, so I'll mention this one to the team. Hopefully we can find a template/design that makes things more obvious here.

@kfloey if you wouldn't mind switching this to a structured answer while you're at it, that would be awesome.

mintsoft:
@kfloey @moollaza I like the simplest form reduction personall, what if it was changed to something like: `root(40) = 2 * root(10) ~ 6.324555`

kfloey:
@moollaza I will get working on that
@mintsoft I agree. I like the reduction and there should be a better way to format it.

Also, I might have found another bug.  If the query is "negative square root of 9" it returns the inverse of the result. 3 becomes 1/3.  I'm not sure that would be the desired or intended answer.

[...]

PR #1024, opened 2015-02-26
Updated Root Calculator to structure answer and fixed other bugs

kfloey:
This commit fixes the bug described in issue #1019

The IA now uses the structured answer format and removes the simplified answer in the parentheses for whole number answers
screenshot from 0000-00-00 00 00 00

but not for answers that are not whole numbers
screenshot from 0000-00-00 00 00 00

- [x] Tested cross-browser compatibility


kfloey:
@mintsoft @moollaza I still need to fix the negative square root bug, but this is my progress so far

mintsoft:
@kfloey I like that, looks like the best of both worlds to me

kfloey:
@moollaza Everything is working as it should; I fixed the bug for negative root.

screenshot from 2015-02-28 10 47 04

moollaza:
@kfloey looks great!

I noticed a couple other problems which are separate from this. You're welcome to fix them in this PR, otherwise I'll make a new issue for it:

2015-02-28_11h34_04
2015-02-28_11h33_18

kfloey:
@moollaza for each of the above issues is the desired output the `sqrt(-100+12)` and `sqrt(100-34)`?  Or are you looking for `sqrt(-100)+12` and `sqrt(100)-34`.  

For bugs, I also noticed that the regex used to control the trigger some work. For example:

screenshot from 2015-02-28 12 11 16

Also, try adding any word in front of square root of "number" (Try: `foo square root of 9`) and duckpan returns an error. I don't think that will show an error to the users, but I doubt DDG want buggy code...

mintsoft:
@kfloey I think @moollaza expects `sqrt(100-34)` for the second one. 
Personally for `negative 100 +12` its exactly what i'd expect, `negative (100+12)`. Maybe @moollaza expected `sqrt(12-100)` i.e. `sqrt(-88)`?

kfloey:
@moollaza I was wrong about the regex needing work.  The reason that those bugs are occurring is because the `str2nbr` function does not support those inputs...

kfloey:
@moollaza @mintsoft Is there a reason this goodie has not been merged with the Calculator goodie.  I would think that some of the things we would want done in the CalcRoots.pm would overlap with the Calculator.pm a little bit.

mintsoft:
@kfloey Good point, Calculator hasn't got any support for handling of radicals or rationals etc currently, it's only doing simplistic decimal calculations. I don't think there'd be any objection to the idea of extending it to support such things if you wanted to etc, it's just never been done!

moollaza:
> I don't think there'd be any objection to the idea of extending it to support such things if you wanted to etc, it's just never been done!

100% agreed! I think there's lots of room to improve the calculator.pm module.

------

Something I pointed out to the team that we briefly discussed though was moving the calculator to the frontend using a lib like [Math.js](http://mathjs.org/) which comes with more functionality than we currently offer, plus it's extensible so we could easily add additional modules to it. 

Any thoughts on that? One big benefit is that we could then easily allow for an interactive calculator

moollaza:
@kfloey @mintsoft sorry I missed the above comments. I was assuming `sqrt(-88)` and `sqrt(100-43)` but it's totally fair that the specified order is ambiguous. I think in either case we parsed this wrong?

On the topic of the calculator IA (and this one) I recently discovered Mathjs.org, which provides a relatively comprehensive (and extensible) library for handling Math equations.

I'll be speaking with the team about moving the calculator (and conversions, which Mathjs supports) to the frontend, so we should probably pause on this until we figure out where we're headed.

Any thoughts or opinions on that?

kfloey:
>Any thoughts on that? One big benefit is that we could then easily allow for an interactive calculator

@moollaza I think that using Math.js would be a great idea.  It seems to have a really well documented site and lots of features.  I've remember seeing at least a few requests for an interactive calculator on the community platform, and I personally believe a better front end Calculator UI would be a lot more efficient and easier to use for DDG users compared to the current calculator IA.


>I'll be speaking with the team about moving the calculator (and conversions, which Mathjs supports) to the frontend, so we should probably pause on this until we figure out where we're headed.

Unless you wanted to merge this so the structured answer for CalcRoots would be live until more progress is made, I see no reason to continue working on this for now, especially if you'll be using Math.js in the future.

moollaza:
This LGTM :+1: 

Not sure how soon Math.js will be coming along.

Issue #1028, opened 2015-02-27
Uppercase: Shouldn't return a result when the input is entirely numerical


moollaza:
Example: https://duckduckgo.com/?q=uppercase+123&ia=answer

>https://duck.co/ia/view/uppercase

MrChrisW:
Fixed by @mintsoft in #1029

PR #1029, opened 2015-02-27
Uppercase: Should not return if the input isn't upper-case-able

mintsoft:
Fixes: https://github.com/duckduckgo/zeroclickinfo-goodies/issues/1028

### DIFF
--- a/lib/DDG/Goodie/Uppercase.pm
+++ b/lib/DDG/Goodie/Uppercase.pm
@@ -28,6 +28,7 @@ handle remainder => sub {
     my $input = shift;
 
     return unless $input;
+    return if $input eq uc($input);
 
     my $upper = uc $input;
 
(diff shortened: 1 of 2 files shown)


lib/DDG/Goodie/Uppercase.pm
```diff
@@ -28,6 +28,7 @@ handle remainder => sub {
     my $input = shift;
 
     return unless $input;
+    return if $input eq uc($input);
```
moollaza: Ah great catch! Better way of handling redundant answers. Nicely done @mintsoft

moollaza:
LGTM :+1:

lib/DDG/Goodie/Uppercase.pm
```diff
@@ -28,6 +28,7 @@ handle remainder => sub {
...
 
     my $upper = uc $input;
```
zachthompson: Move this line to 31 and compare $input to $upper so we don't need to call uc twice.

Issue #827, opened 2014-12-10
Conversions: Show °C and °F instead of "degrees celsius" and "degrees fahrenheit"


jagtalon:
Current:
![screen shot 2014-12-10 at 9 16 25 am](https://cloud.githubusercontent.com/assets/81969/5377207/543d423c-804d-11e4-9240-68b3e1d53dbc.png)

Proposal:
![screen shot 2014-11-05 at 11 05 53 am](https://cloud.githubusercontent.com/assets/81969/5377210/635f17cc-804d-11e4-92ff-d870f8fc32e2.png)

https://duck.co/ia/view/conversions

jagtalon:
This could probably be done while https://github.com/duckduckgo/zeroclickinfo-goodies/pull/781 is being worked on.

greensleeves01:
Excellent new feature!

jagtalon:
Fixed by @ilv in #1030

PR #1030, opened 2015-02-28
Fix Conversions to show °C and °F

ilv:
Fix to #827. I hope it is what you were looking for. 

I also modified the tests file, although there are some tests not related to temperatures that do not pass. Two of them don't pass because of the words mi and sq instead of miles and square. The rest fail because of precision.

### DIFF
--- a/lib/DDG/Goodie/Conversions.pm
+++ b/lib/DDG/Goodie/Conversions.pm
@@ -8,6 +8,7 @@ with 'DDG::GoodieRole::NumberStyler';
 use Math::Round qw/nearest/;
 use bignum;
 use Convert::Pluggable;
+use utf8;
 
 name                      'Conversions';
 description               'convert between various units of measurement';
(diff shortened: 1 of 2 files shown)


MrChrisW:
@ilv Thanks for your contribution, after a quick look this looks great :+1: Will await sign off from @mintsoft & @jagtalon

mintsoft:
@ilv can you add tests to ensure that it still handles queries such as "40 degrees celsius in fahrenheit" as we want to support both :+1: 

The output format can be like you've made it, it's just the inputs that need testing.

mintsoft:
@ilv I think you may be working on an old version of the C::P module too judging bby your comments, have you tried `cpanm Convert::Pluggible` ?

jagtalon:
@ilv nice! Agree with @mintsoft would like to see if "40 degrees celsius in fahrenheit" is working.

jagtalon:
Oh hang on I misread it--I thought it accepts queries such as "40 degrees celsius in °F". @mintsoft i guess it doesn't change what it accepts as input so should be good! :+1:

mintsoft:
@jagtalon As did I actually! Re-reading it's not changed the input behaviour at all

ilv:
Indeed, I did not change the input behaviour, I just implemented what was proposed in #827  :)

@mintsoft Yeah, I was using the version installed on the Virtual Machine from 2014-12-23. I updated Convert::Pluggable and now all test passes! :+1:

mintsoft:
@ilv Cool! Good to hear. Thanks for the work on this

Issue #951, opened 2015-01-24
FrequencySpectrum: Missing diagram labels?


moollaza:
Seems like all the labels in the SVG diagram are missing? Looks like there's some text in the Markup but I'm not seeing it display correctly in FF or Chrome.

![2015-01-23_23h37_03](https://cloud.githubusercontent.com/assets/873785/5886129/f3b83db6-a358-11e4-9ea9-ebb0ff81b1e7.png)

https://duck.co/ia/view/frequency_spectrum

mintsoft:
Fixed in https://github.com/duckduckgo/zeroclickinfo-goodies/pull/1033 by @SoerenWeber

PR #1033, opened 2015-03-01
Fix for Issue #951: missing diagram labels on the y-axis.

SoerenWeber:
This should show the frequency-range labels on the y-axis (#951). The corresponding tspan elements are now children of text elements instead of being their siblings in the SVG.

PR #1035, opened 2015-03-02
Show marker label at the top in FrequencySpectrum

SoerenWeber:
This is a follow-up fix for #951. This bug didn't show up in duckpan because empty tags weren't recognized and rewritten. It's the same issue as for the y-axis labels.

### DIFF
--- a/lib/DDG/Goodie/FrequencySpectrum.pm
+++ b/lib/DDG/Goodie/FrequencySpectrum.pm
@@ -616,7 +616,7 @@ sub add_marker {
         'text-anchor' => 'middle',
         class => 'marker_label'
     );
-    $markerLabel->tag('tspan', id => 'marker_label', -cdata => ucfirst($freq_formatted));
+    $markerLabelText->tag('tspan', id => 'marker_label', -cdata => ucfirst($freq_formatted));
 
     #Add marker line
     $plot->{svg}->group(


moollaza:
@SoerenWeber it looks like the MarkerLabelText needs to be coloured Black (#000) and set to `display: block` or similar?

When testing this with your DuckPAN PR I'm not seeing the actual text display because it has a default `visibility: hidden` and also FrequencySpectrum.css sets the colour to `#FFF` which makes it invisible against the white background.

---

One more thing: I see there's a `<rect id="marker tag">` on that page that's also hidden by default. Showing it reveals a small red line. I'm not sure what it was intended to do but it seems pointless. I think we should remove it.

SoerenWeber:
@moollaza What you're describing seems to be a product of the JavaScript code at the beginning of the script. Your browser probably runs into the catch-block that hides those elements.
It's working in my browser (Chromium 40, 64-bit). In this case the label is readable because the `<rect id="marker tag">` becomes a red rectangle in the background of it.

Nonetheless, I'm not sure if the JavaScript code is indeed necessary.

mintsoft:
Just for reference: 
image
It took me a few seconds to figure out what this is solving

SoerenWeber:
That's not completely correct, @mintsoft. Sorry if I wasn't precise enough. It's about this marker:
marker

#1033 fixed the missing Y-axis labels already - at least I hope so. ;)

mintsoft:
Ah @SoerenWeber Apologies, thanks for clearing that up

SoerenWeber:
No worries @mintsoft, I should have clarified that.
___
I'll try to have a look at the JS code next to find out how we could make it more compatible.

moollaza:
LGTM :+1:

moollaza:
@SoerenWeber we'll be launching templated Goodies that are able to execute JavaScript **very** soon, so hopefully we can cleanup the JS for this one shortly :smiley:

PR #1036, opened 2015-03-03
First DDG Game! 2048

puskin94:
With this IA it is  possible to play 2048 online
play 2048, 2048 game

gamers!

## Checklist

```
- [] Added metadata and attribution information (https://duck.co/duckduckhack/metadata)
- [] Wrote test file and added to t/ directory (https://duck.co/duckduckhack/goodie_tests)
```

Tomorrow I will add all the metadatas, test files and all the missing code ;)

-----
IA Page: https://duck.co/ia/view/game_2048


moollaza:
@puskin94 thanks a lot for this! Unfortunately given all the JS I think this should really be a Spice.

Please take a look at similar "pure JS" Spices like the Timer and Stopwatch to see how they're implemented.

moollaza:
@puskin94 actually hold on with developing this further. It looks like this could be a great candidate for the new type of Templated Goodies we're moving towards, which will also be able to run JavaScript.

puskin94:
@moollaza so what are you suggesting?? Finish importing this in Spices or waiting for the new Goodie Template and finish it here?

mintsoft:
@puskin94 up to you, I think if you want it merged and live soon (i.e. within a week or so) then switch to a spice, if you don't mind waiting a little longer until the templates are ready then use a Goodie!

mintsoft:
@puskin94 Btw it'd be awesome if you could do like `play 1024` or play `4096`!

puskin94:
@mintsoft Just tell me if I should wait for the new Goodie Template or not... you are the masters ;) I can do what DDG asks me ;)
Talking about the possibility to play 1024 or 4096 too, I definitively will develop this feature :+1:

moollaza:
@puskin94 sorry. Don't move it to Spice. We'd like to work with you to make this into a new special type of Goodie. Feel free to finish up the tests, metadata, etc and we'll get back to you soon when we're ready to move forward.

puskin94:
@moollaza no problem! I contine fixing bugs and I'm so excited to work with DDG for this type of Goodie!

puskin94:
There are a lot of bugs... I'm working on it locally. Soon ( I hope :+1: ) I will release it

puskin94:
@moollaza @mintsoft I only have a little problem with test file. Which 'object' should it have??
I'm passing to it only the `html =>` one, but i retrive this error:
`The new() method for DDG::ZeroClickInfo expects a hash reference or a key/value list. You passed an odd number of arguments`. How can I fix it?? 
P.S. the test file online isn't the last one because I changed the html that this IA returns... but there aren't other differences.

mintsoft:
@puskin94 you'll need to output something for the text output:
```perl
return "", html => "$play $html";
```

puskin94:
@mintsoft Hei, I updated the `.pm` file and the test file too... But I get the same error!

puskin94:
@moollaza @mintsoft as you can see in those days I have made a massive Update:
1) Style improved
2) code cleanup
3) can play from `128` to `8192` ( can be modified )
4) can play in a dynamic area size ( from 4x4 to 10x10 ( upper limit is for a better layout )
5) added the points feature

I need only a help with the test file...

puskin94:
@moollaza The only way I found that can fix the tests is a very generic regex... Any advice?

mintsoft:
@puskin94 that's fine; we have a few other IA's whose output is too complex to test

puskin94:
@mintsoft you made me a happy man!

t/2048.t
```diff
@@ -0,0 +1,24 @@
...
+
+zci answer_type => 2048;
+zci is_cached   => 1;
+
+
+ddg_goodie_test(
+    [qw( DDG::Goodie::2048 )],
+
+    'play 2048' => test_zci(
+		'Play 2048', html => qr/(.+)/
```
moollaza: I forgot, you can also put `html => '-ANY-'` -- which does basically the same thing...

puskin94:
@moollaza Thanks for the tip! I updated the test file! Now it looks better!

puskin94:
@moollaza I have a problem... I looked at your last pull request and I was ispired by it. However I'm unable to load the `.handlebars` file... Js is loaded properly but `content => 'DDH.2048.content'` not. How can I solve?

puskin94:
@moollaza Nope, I studied more the results and the only problem is that I cant push on page source the `$play` variable. I know it is currently in beta but there are problems loading the IA template like
2048-test7

moollaza:
@puskin94 is the newest code pushed? if not please push it here and I'll take a look

puskin94:
@moollaza Magically the template loads good this morning!
2048-test8
But the problem is  still there: I can't push `$play` out and, as you can see, js gets angry

moollaza:
I'll try to prepare an example for you today/tomorrow. The JS loading is ready :)

puskin94:
@moollaza Have you any news for this IA?

moollaza:
@puskin94 sorry for the delays.

You can create a JS file in your /share/ dir named game_2048.js.

It's content should look like this:

```js
DDH.game_2048 = DDH.game_2048 || {};

DDH.game_2048.build = function(ops) {

    console.log("Init!");

    // Handlebars Helpers and priv functions here

    return {
         onShow: function() {
            console.log("Look here!");
         }
    };
};
```

**Note:** I think you'll need to change the name of the Perl Package because the `id` being numbers, is going to cause problems (or you'll have some pretty ugly JS)

puskin94:
@moollaza Hei, as you can see here
```
            function stopDefAction(evt) {
                evt.stopPropagation();
                evt.preventDefault();
            }

            var duckbar = document.addEventListener(
            'keydown', stopDefAction, false);
```
I tried to stop the keypress handling but without success. Obv this is an example code to study how the 2 js functions works... If you exec this IA something works about the keypress event: You can't press `enter` in the duckpan search field but this isn't my quest ;) Any suggestion?

moollaza:
@puskin94 your keydown listener needs to look like this in order to work:

```js
            $('html').keydown(function(e){

                e.stopPropagation();
                e.preventDefault();
```

Please make sure you read our JavaScript Styleguide and your code adheres to it. As well, I strongly suggest you use jQuery for all your DOM manipulation and searching. Your code will be much cleaner and more succinct.

For example:

```js
Document.getElementById('container'); => $('#container');
Document.getElementsByClass('points'); => $('.points');
```

Also, please move your private function declarations outside of `onShow()`. That function (`onShow()`) fires every time the tab is opened. Which means you're redeclaring all that code, and running all of it again. You'll want to declare private variables and function before your `return` call and you'll likely want some way to manage state in case someone click to another tab and comes back.

This code will be seen and used by many, many people, so let's make sure it's excellent ;)

puskin94:
@moollaza ok man! Thanks for the help! I will change the code shortly and I will ping you when done ;) :+1:

puskin94:
@moollaza Hei master. I converted all the IA to jQuery... I hope it is good since this is my first approach to this. My only doubt is the test file... taking as example this https://github.com/duckduckgo/zeroclickinfo-goodies/blob/5d2805bc901bdd10ab3c1157306630773f4939eb/t/BPMToMs.t I note you don't specify `data`. Just let me know ;)

share/goodie/game2048/content.handlebars
```diff
@@ -0,0 +1,7 @@
+<br><i><b>Play <span id="game">{{inputNum}}</span> <br><br>
```
moollaza: Let's get rid of the `<br>` in favour of CSS. Much cleaner.

share/goodie/game2048/game2048.css
```diff
@@ -0,0 +1,27 @@
+.area {
```
moollaza: Looks like your indents have mixed tabs and spaces. Please switch to 4 space indentation.

share/goodie/game2048/game2048.css
```diff
@@ -0,0 +1,27 @@
...
+	table-layout: fixed;
+	border: 3px solid #776E65;
+	border-collapse: collapse;
+	margin-left: auto; 
+    margin-right: auto;
+    box-shadow: 2px 2px 2px #999;
+    font-style: italic;
+}
+
+.area td, tr {
```
moollaza: All these CSS rules need to be namespaced so that they don't affect other elements on the page. We auto wrap your Goodie in a `<div class="zci--game_2048">` Please prepend `.zci--game_2048` to all your CSS rules.

share/goodie/game2048/game2048.js
```diff
@@ -0,0 +1,292 @@
...
+    // moves the numbers
+
+    function mov(dir, area) {
+        var i,
+            points = 0,
+            moves = 0,
+            flag = false,
+            exit;
+
+        if (dir == 'a' || dir == 'd') {
```
moollaza: Please use `===` instead of `==`, similarly `!=` should be `!==`

share/goodie/game2048/game2048.js
```diff
@@ -0,0 +1,292 @@
...
+            area = transpose(area);
+        }
+        if (dir == 's' || dir == 'd') {
+            area = swapRows(area);
+        }
+
+        for (var c = 0; c<SIZE; c++) {
```
moollaza: There's a lot of nesting here and it's hard to follow. Any chance this can be simplified or there's a better (non brute-force approach to this?

It may make the code more legible if you switch `c` with `col` and `r` with `row` for example. I think there's definitely room for improvement here. jQuery gives really good helper classes like `siblings()`, `prev()`, `next()` -- I feel like some of those could be used to make this function more efficient.

share/goodie/game2048/game2048.js
```diff
@@ -0,0 +1,292 @@
...
+        }
+        // This check is mandatory in order to avoid the appearance of a new
+        // value in the area if no moves has been made
+        return flag;
+    }
+
+    // Updates the 'points' div
+
+    function upPoints(points) {
+        var spanPoints = $('.points');
```
moollaza: - `increasePoints` seems like a better name?
- I imagine this function be called more than once? If so, do we need to re-declare `var spanPoints` every time, and search through the DOM? If not it's best to create you jQuery variables once so the DOM lookups are all cached and use some type of `$container` variable to make the lookups faster:

```
// init, run once
var $gameContainer = (`.zci--game_2048 .container`),
    $points = $gameContainer.find('.points'),
    ...;
```

share/goodie/game2048/game2048.js
```diff
@@ -0,0 +1,292 @@
...
+        var current = parseInt(spanPoints.text());
+        spanPoints.text(current + points);
+    }
+
+    // After every little table's change, the area is updated.
+    // This function changes the cell background too
+
+    function printArea(area) {
+        var val;
+        for (var r = 0; r < SIZE; r++) {
```
moollaza: Looking like you're looping over all td's here? best to use the `.each()` method:

```js
$('td').each(function(){ ... });
```

share/goodie/game2048/game2048.js
```diff
@@ -0,0 +1,292 @@
...
+            for (var c = 0; c < SIZE; c++) {
+                val = area[r][c];
+                $(tempArea).find('tr').eq(r).find('td').eq(c).html(val);
+                $(tempArea).find('tr').eq(r).find('td').eq(c).css("background-color",color[val]);
```
moollaza: I think it would be more efficient to modify or add/remove a CSS class rather than set the BG colour on-the-fly.

Your stylsheet can describe the colors that the elements should have when they have the corresponding class:

```css
td .points-2 {
    background-color: #aaa;
}

td .points-4 {
    background-color: #aaa;
}

...
```

I just took a peak at https://gabrielecirulli.github.io/2048/ and he handles it the same way which I believe is best. Let's leave the appearance logic and definitions to the CSS. The JS should just update the classes.

share/goodie/game2048/game2048.js
```diff
@@ -0,0 +1,292 @@
...
+        }
+    }
+
+    // 'area' initialization
+
+    function getArea(area) {
+        var sub = [];
+        for (var r = 0; r<SIZE; r++) {
+            for (var c = 0; c<SIZE; c++) {
+                sub[c] = $(area).eq(r).find('td').eq(c).text();
```
moollaza: Same here. you can just each jQuery's iterators, `.each()`

share/goodie/game2048/game2048.js
```diff
@@ -0,0 +1,292 @@
...
+    function start() {
+        getArea(area);
+        getRand(area);
+    }
+
+
+    return {
+        onShow: function() {
+
+            container = $('#2048-area');
```
moollaza: You still need to maintain state. Click the images tab and then click your game tab -- all of this will run again and the game-board doubles in height I believe. (That's what was happening last time I tried it..)

puskin94:
@moollaza ok, I improved all your suggestions. About the nesting in `mov` functions, I can't use jQuery functions because I use a 'local' copy of the gaming area before print it on the page... It is a simple `array[$SIZE][$SIZE]`. How can I mantain the page state avoiding the doubling in height of the table?

share/goodie/game2048/game2048.js
```diff
@@ -0,0 +1,274 @@
...
+            }
+        }
+        return false;
+    }
+
+    // If no moves or sums are possible, returns true
+
+    function checkLose(area) {
+        var count = 0;
+        for (var row = 0; row < $SIZE; row++) {
```
mintsoft: This loop and the one below `canDoSomething` can be merged into one iteration; twice as fast!

share/goodie/game2048/game2048.js
```diff
@@ -0,0 +1,274 @@
...
+            $('html').keydown(function(event){
+
+                event.stopPropagation();
+                event.preventDefault();
+
+                var move = false;
+
+                if (goOn) {
+
+                    if (event.keyCode === 87 || event.keyCode === 38) { // w or up arrow
```
mintsoft: It'd be cool to add the numpad too, either 8456 or 8426

share/goodie/game2048/game2048.js
```diff
@@ -0,0 +1,274 @@
...
+        var i,
+            points = 0,
+            moves = 0,
+            flag = false,
+            exit;
+
+        if (dir === 'a' || dir === 'd') {
+            area = transpose(area);
+        }
+        if (dir === 's' || dir === 'd') {
```
mintsoft: shouldn't that be `dir === 's' || dir === 'w'` ?

share/goodie/game2048/game2048.js
```diff
@@ -0,0 +1,274 @@
...
+                }
+                i = 0;
+            }
+            moves = 0;
+        }
+
+        if (dir === 's' || dir === 'd') {
+            area = swapRows(area);
+        }
+        if (dir === 'a' || dir === 'd') {
```
mintsoft: Same again, unless I'm totally misinterpreting this. if I am, I think it might be worth putting 's' 'a' and 'd' into const's that are more explicitly named as to their purpose?

puskin94:
@mintsoft nope. `swapRows` and `transpose` are used because I made only one movement for every direction ( ported to the 'up' movement = `w` ). If you draw a square in a paper, you can note that, if you want to make a move to the right ( `->` or `d` ), applying first `swapRows` and `transpose`, all the movements can be done using the same 'math movements' used for the `w` ( up ) action. 
I hope you have understood this explanation ;)

mintsoft:
@puskin94 I see; I think we could probably do with something to make that more obvious around the code there. It's pretty likely that other people will misunderstand as I did :)

puskin94:
@mintsoft should I put in comments how these two functions works?

puskin94:
@moollaza Very good news! As you can see in comments I fixed the table duplication using a boolean variable
 ```
/* 'started' is a boolean variable used in order to avoid the
            duplication of the gaming area. Moving around the DDG tabs the
            'onShow' function is executed over and over. This simple solution
            prevents the problem */
```
I think it is ready for another staff review ;)

**Comment by puskin94** [comment]
@moollaza Hei I reverted all the brackets and I really can't see about brackets ( except their position ) in the styleguide ( https://duck.co/duckduckhack/code_styleguide ). 
What can you tell me about the code?

File: lib/DDG/Goodie/Game2048.pm [hunk 16]
```diff
@@ -0,0 +1,65 @@
+package DDG::Goodie::Game2048;
```
moollaza: Looks like this file is indented with tabs? Please use 4 spaces.

lib/DDG/Goodie/Game2048.pm
```diff
@@ -0,0 +1,65 @@
...
+
+handle remainder => sub {
+
+	(my $inputNum, my $dimension) = split/ /;
+
+	return unless $inputNum;
+
+	my $base = 128;
+
+	return unless (($inputNum == $base) || # play 128
```
moollaza: I believe you can cut this down to 1 line:

```perl
return unless $Input % 128 == 0 && $input < 8192;
```

lib/DDG/Goodie/Game2048.pm
```diff
@@ -0,0 +1,65 @@
...
+					($inputNum == $base*16) || # play 2048
+					($inputNum == $base*32) || # play 4096
+					($inputNum == $base*64)); # play 8192
+
+
+	if (!$dimension || $dimension > 10 || $dimension < 3) {
+		$dimension = 4;
+	}
+
+
```
moollaza: Any need for the double whitespace? Let's reduce to single empy lines if the separation is needed.

lib/DDG/Goodie/Game2048.pm
```diff
@@ -0,0 +1,65 @@
...
+	return $text,
+	structured_answer => {
+		id => 'game2048',
+		name => '2048',
+		data => {
+			inputNum => $inputNum,
+			dimension => $dimension
+		},
+		templates => {
+			group => 'base',
```
moollaza: I think you can actually use the `text` template? You can build the `title` in the Perl and add it to `data` -- that way your template doesn't need to create the title. You also don't need to parse the title `<span>` to get the game size -- it's already being passed along to the JS via the `data` object :)

share/goodie/game2048/content.handlebars
```diff
@@ -0,0 +1,7 @@
+<i><b>Play <span id="game">{{inputNum}}</span>
+<div id="dimContainer"><span id="dimension">{{dimension}}</span> x {{dimension}}</b></i></div>
```
moollaza: Please nest HTML elements for legibility:

```html
<div id="dimContainer">
	{{dimension}} x {{dimension}}
</div>

<span class="counter">
	<span class="points">0</span> Points
</span>

<div id="2048-area"></div>
```

share/goodie/game2048/game2048.css
```diff
@@ -0,0 +1,42 @@
...
+    height: 50px;
+    width: 50px;
+    border: 3px solid #776E65;
+    text-align: center;
+    font-weight: bold;
+}
+
+.zci--game2048 .points {
+    font-weight: bold;
+    font-style: italic;
```
moollaza: Let's get rid of the italics -- it's not necessary. Our design team will definitely have some feedback on how this IA should look :)

share/goodie/game2048/game2048.js
```diff
@@ -0,0 +1,346 @@
...
+            return true;
+        }
+
+        return false;
+    }
+
+
+    // This function creates and prints on page the gaming table
+
+    function createTable() {
```
moollaza: This doesn't need to happen after rendering -- you can use logic in the template to add the correct number of rows and columns. The `<table>` could be put in template though and you can use a helper, like `{{loop}}` (https://duck.co/duckduckhack/spice_handlebars_helpers#loop) to build the `<tr>` and `<td>`.

As much of the initial UI building logic that you can put into the template, the better :)

lib/DDG/Goodie/Game2048.pm
```diff
@@ -0,0 +1,65 @@
...
+name "2048";
+description "Javascript IA for online 2048";
+primary_example_queries "2048 game", "play 512";
+secondary_example_queries "play 4096";
+category "entertainment";
+topics "gaming";
+code_url "https://github.com/duckduckgo/zeroclickinfo-goodies/blob/master/lib/DDG/Goodie/Game2048.pm";
+attribution github => ["https://github.com/puskin94", "puskin"];
+
+triggers any => "game", "play";
```
moollaza: We might as well just list out the triggers here:

```perl
my @games_modes = qw (128, 256, 512, 1024, 2048, 4096, 8192);
my @triggers = map { "play $_" } @game_modes;
triggers start => @triggers;

handle remainder => {
    return unless m/^\d{1,2}$/;
    ...
}
```

puskin94:
@moollaza Thanks for your advices! I improved all your suggestions except the last one, I think that
```
    # Play 128, 256, 512, 1024, 2048, 4096, 8192
    return unless $inputNum % 128 == 0 && $inputNum <= 8192;

    if (!$dimension || $dimension > 10 || $dimension < 3) {
        $dimension = 4;
    }
```
is better, cleaner and most intuitive than 
```
my @games_modes = qw (128 256 512 1024 2048 4096 8192);
my @triggers = map { "play $_" } @game_modes;
triggers start => @triggers;

handle remainder_lc => {
    return unless $_ eq '' || m/^(?<dimension>\d{1,2})$/; 
    my $dimension = $+{dimension} // 4;
    ...
}
```

moollaza:
@puskin94 my code is definitely more concise and "Perl-y" -- other Perl devs would be able to understand what I've written there with the `qw` and `map`. The biggest benefit though is that it considerably reduces the trigger space of this IA which we want.

With the triggers set to `play` -- we'll run your handler against every query that contains `play` -- which is potentially huge and leads to a lot of wasted computer cycles. Seeing as we know we **only** want to show this for a very small subset of those queries, being "Play 128", "Play 256" ... "Play 8192" it's best to set those exact queries as the triggers. 

However you want to build those triggers is fine, but my method is quite succinct and doesn't rely on a comment to indicate exactly what triggers we're expecting to other developers or the computer ;)

mintsoft:
@moollaza FWIW I definitely prefer:
```perl
my @games_modes = qw (128 256 512 1024 2048 4096 8192);
my @triggers = map { "play $_" } @game_modes;
triggers start => @triggers;
```
Nice and declarative!

puskin94:
@moollaza @mintsoft Hei guys. I improved triggers as you suggested me and I added a regex that should handle good the IA calling

moollaza:
@puskin94 very sorry for the delay, I've been busy but will try and get around to testing this again shortly!

puskin94:
No problem :+1:

javathunderman:
ping @moollaza

javathunderman:
@moollaza ping!

share/goodie/game2048/game2048.js
```diff
@@ -0,0 +1,341 @@
...
+
+                $container = $('#game2048__container');
+                $spanPoints = $('.game2048__points');
+                $tempArea = $('#game2048__area');
+                WINNUM = ops.data[0].inputNum;
+                SIZE = ops.data[0].dimension;
+
+                start();
+
+                $('html').keydown(function(event){
```
moollaza: Looks like we still need a way to reliquinsh control whn the game no longer has "focus". If the user clicks anywhere outside the game, (like in the searchbox, another tab, or the settings button etc) the user shouldn't be bound to these controls because it breaks navigation on the page.

moollaza:
@puskin94 I'm very sorry that this PR has been sitting idle for so long. Let's get it finished and out the door!

@abeyang any thoughts on design here? I feel like the game board should be a little bigger to fill the space?

lib/DDG/Goodie/Game2048.pm
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    my $text = 'Play '.$inputNum;
+
+    return $text,
+    structured_answer => {
+        id => 'game2048',
+        name => '2048',
+        data => {
+            inputNum => $inputNum,
```
moollaza: Any reason we aren't passing the `title` and `subtitle` along to the frontend from here? That way we don't need to bother formatting them (right now they look very plain)

share/goodie/game2048/game2048.js
```diff
@@ -0,0 +1,341 @@
...
+        var current = parseInt($spanPoints.text());
+        $spanPoints.text(current + points);
+    }
+
+    /* After every little table's change, the area is updated.
+    This function changes the cell class too */
+
+    function printArea() {
+        var val;
+        $("td").each(function(idx) {
```
moollaza: our array of "td"'s should definitely be cached in a variable. This function get's called very often, and every time it's called, `$("td")` is navigating the dom looking for any and all `td` elements.

We already have our `$container` cached which helps. We should follow suite and cache the `td`s in a similar fashion and let all these functions use that instead.

If you're curious to know more about the performance of the code, I suggest you benchmark it in your browser. I just played a game and benchmarked it in Firefox, which showed that this fn is quite costly!

share/goodie/game2048/game2048.js
```diff
@@ -0,0 +1,341 @@
...
+            $(this).html(val).attr("class","boxtile val-"+val);
+        });
+    }
+
+    // 'area' initialization
+
+    function getArea() {
+        var sub = [SIZE];
+        for (var row = 0; row < SIZE; row++) {
+            $("td").each(function(idx) {
```
moollaza: Same thing here, let's cache these in an array variable so we don't need to find them every time. Also, let's make the initial selector more specifc so we don't look for TD's in unnecessary places: `$container.fin("td")` is a better approach

share/goodie/game2048/game2048.js
```diff
@@ -0,0 +1,341 @@
...
+    position around the area. 4 has a 10% chance of being chosen. */
+
+    function getRand() {
+
+        var rand = Math.floor(Math.random() * 11);
+        var posX, posY;
+
+        rand = (rand < 10) ? 2 : 4;
+
+        do {
```
moollaza: Any way we can keep track of the non-empty tiles in an array? Then we could randomly generate a number based on the array index and we wouldn't need to waste cycles on generating numbers for tiles we can't use.

share/goodie/game2048/game2048.js
```diff
@@ -0,0 +1,341 @@
...
+        }
+        return false;
+    }
+
+    // If no moves or sums are possible, returns true
+
+    function checkLose() {
+        var count = 0;
+        var canDoSomething = false;
+        for (var row = 0; row < SIZE; row++) {
```
moollaza: Seeing as we know the board size always, how about we give each tile a number from 0 to 16 and loop over that. We can store the row and column for each in an object and then that way we won't need to use a nest for loop inside another loop.

I think there's lots of room for improvement in this code and I think it would be great for you to try and find more places we can speed this up :smile:

share/goodie/game2048/game2048.js
moollaza: I just realized that right now, this is capturing all key presses and so we're preventing the user from even being able to type into the search box!

I think you'll want to move the `stopPropagation` and `preventDefaul` below after we've verified it's one of the keys we care about and we know a game is ongoing.

abeyang:
@puskin94 sweet! Maybe someone mentioned this already (sorry, pretty long thread), but my keyboard is hijacked while the game is running.. need to fix that.

In the meantime, I'll try to work on the visuals for this!

abeyang:
..also, I noticed that you put down "4x4" in the IA, which led me to believe that you can play other sized boards (like 6x6).. but I wasn't able to? Is that in the works, or is this strictly for 4x4 only?

moollaza:
Hey @puskin94 

It looks like you've been away from this PR for a while. While we'd really like to have this game as an IA, it doesn't look like you have the time to finish. Given that it still has problems, we're going to mark this PR as "abandoned" and open it up for the rest of the community to complete. If you'd like to fix your bugs and resubmit, that's great too. Otherwise we'd like to let other interested developers take up the charge.

Thanks again for all your work so far, we really appreciate it and can't wait to see this live.

PR #1041, opened 2015-03-05
ColorCodes: added link to DDG's color picker

mattr555:
Now that we have duckduckgo/zeroclickinfo-spice#1506, it would be cool if we could link to that with the existing color code goodie:

screenshot - 03042015 - 11 16 11 pm

Perhaps we want to make it more prominent? (Though I think it's good as it is.)


moollaza:
@mattr555 looks like the tests need to be updated!

chrismorast:
Is this live somewhere?

moollaza:
@chrismorast here you go! https://moollaza.duckduckgo.com/?q=rgb+orange&ia=answer

mintsoft:
@mattr555 cool idea1

jagtalon:
Yeah this is a great idea! :)

mintsoft:
Thanks for addressing #1043 at the same time @mattr555 !

mattr555:
No problem! I also noticed that the mobile layout was pretty weird:

before

So I played with the spacing and made the swatch smaller:

after

mintsoft:
Cool that looks much better!

jagtalon:
@mattr555 one small thing is that the text is hard to see with the dark theme:

screen shot 2015-03-06 at 2 55 51 pm

They should have a class of `tx-clr--lt`

abeyang:
@jagtalon @mattr555 The colors are also hard to distinguish apart in the dark theme. :smile:

mintsoft:
@mattr555 @jagtalon @abeyang That's a known issue; @moollaza was looking into the best way to deal with it

mattr555:
@mintsoft Ok, well at any rate, here's a temporary fix

screenshot - 03062015 - 04 28 34 pm

chrismorast:
Would it make sense to link the main swatch to the color picker IA as well?

chrismorast:
Hey @mattr555 , one alignment pass on mobile and I think this is good to merge.

Here's the current screenshot with guidelines:
screen shot 2015-03-10 at 4 09 41 pm

And here's what I'd like to shoot for:
screen shot 2015-03-10 at 4 09 25 pm

mintsoft:
> Would it make sense to link the main swatch to the color picker IA as well? Or even all of the swatches for that matter. Whichever swatch you clicked would open the color picker with that color active.

That'd be cool!

moollaza:
@mattr555 this is looking great but it doesn't seem like it's perfect just yet:

This is a Nexus 6 on Browserstack.

2015-03-17_01h19_56

mattr555:
@moollaza Is this an issue? I can't really make the swatch bigger, so I decided to center it with the block of text rather than align it to the top.

moollaza:
@mattr555 sorry I meant the horizontal alignment of the large swatch. It should be in line with the swatches below, so it looks like it needs a bit more of a left-margin on mobile. Similarly the "Complementary" and "Analogous" sections need to be pushed a little more to the right so the left edge of the text is aligned with the above text.

Make sense?

chrismorast:
@moollaza , I've already posted feedback on that above.

mattr555:
@moollaza @chrismorast Yeah, whoops, totally missed that!

Updated:
img_2915

chrismorast:
:+1: Thanks @mattr555 !

PR #1042, opened 2015-03-05
Guitar Chord Goodie: Regex Fixes

charles-l:
I believe I've fixed it with this regex. It's not very lenient, so let me know if I should add more options for what is a valid guitar chord.

/cc @jagtalon @mintsoft @moollaza

### DIFF
--- a/lib/DDG/Goodie/GuitarChords.pm
+++ b/lib/DDG/Goodie/GuitarChords.pm
@@ -46,7 +46,7 @@ handle remainder => sub
 # that it will match a filename in the share directory.
 sub check_chord
 {
-    if ($_[0] =~ /(?<a>[a-gA-G])(?<b>#|b)?(?<c>dim|min|maj|add|aug|m|M)?(?<d>M|maj|m|min)?(?<e>[0-9]?(sus[0-9])?)?(\/(?<f>(#|b)?[0-9]+))?/) {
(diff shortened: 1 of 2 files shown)


lib/DDG/Goodie/GuitarChords.pm
```diff
@@ -42,9 +42,11 @@ handle remainder => sub
     return;
 };
 
+# Checks a chord to make sure it's valid and formats it so
+# that it will match a filename in the share directory.
 sub check_chord
 {
-    if ($_[0] =~ /(?<a>[a-gA-G])(?<b>#|b)?(?<c>dim|min|maj|add|aug|m|M)?(?<d>M|maj|m|min)?(?<e>[0-9]?(sus[0-9])?)?(\/(?<f>(#|b)?[0-9]+))?/) {
+    if ($_[0] =~ /^(?<a>[a-gA-G])(?<b>#|b)?\s*(?<c>dim|min|maj|add|aug|m|M)?\s*(?<d>M|maj|m|min)?(?<e>[0-9]?(sus[0-9])?)?(\/(?<f>(#|b)?[0-9]+))?$/) {
```
mintsoft: I just noticed that this will match `G m m`

moollaza: @charles-l @mintsoft it's times like these that I turn to debuggex :smile: 

https://www.debuggex.com/r/zd1y4H5J_lZFEqjA

charles-l: @mintsoft @moollaza true, but it double checks that it's a valid chord first (in the chord hash).

@moollaza that's a neat tool, I'll definitely use that in the future :)

mintsoft: @charles-l Fair enough then! Some things are definitely easier to validate from the matches post-regex

Issue #1038, opened 2015-03-04
Pi: Make it work with "pi to 10 digits"-type queries.


jagtalon:
https://duck.co/ia/view/pi

jagtalon:
Fixed in #1045

Issue #1037, opened 2015-03-04
Pi: Make it work with π


jagtalon:
Such as "60 digits of π"

https://duck.co/ia/view/pi

marianosimone:
Strangely enough, there's a test to support this (https://github.com/duckduckgo/zeroclickinfo-goodies/blob/master/t/Pi.t#L20), but it seems like when it's running on the server, it's not matching.

Might be related to encodings?

mintsoft:
@marianosimone You're right, although there isn't a specific test for queries of this pattern; personally I assumed it wasn't implemented because it isn't tested!

Fixing the problem might be as simple as adding `use utf8;` into the `.pm` and the `.t` !

PR #1045, opened 2015-03-06
Pi IA: triggers improved

puskin94:
@jagtalon Hei, I improved the triggers as suggested here https://github.com/duckduckgo/zeroclickinfo-goodies/issues/1038 and https://github.com/duckduckgo/zeroclickinfo-goodies/issues/1037

### DIFF
--- a/lib/DDG/Goodie/Pi.pm
+++ b/lib/DDG/Goodie/Pi.pm
@@ -15,7 +15,8 @@ attribution github => ["https://github.com/puskin94", "puskin94"],
             github => ["https://github.com/jmvbxx", "jmvbxx"];
 
 
-triggers startend => "pi", "digits of pi","π","digits of π", "pi to", "π to";
+triggers start => "pi","π", "pi to", "π to";
+triggers end => "digits of pi", "digits of π";
 
 # $PI now have 1 000 000 decimals
 
--- a/lib/DDG/Goodie/Pi.pm
+++ b/lib/DDG/Goodie/Pi.pm
@@ -15,7 +15,7 @@ attribution github => ["https://github.com/puskin94", "puskin94"],
             github => ["https://github.com/jmvbxx", "jmvbxx"];
 
 
-triggers startend => "pi", "digits of pi";
+triggers startend => "pi", "digits of pi","π","digits of π", "pi to", "π to";
 
 # $PI now have 1 000 000 decimals
 
--- a/lib/DDG/Goodie/Pi.pm
+++ b/lib/DDG/Goodie/Pi.pm
@@ -15,8 +15,7 @@ attribution github => ["https://github.com/puskin94", "puskin94"],
             github => ["https://github.com/jmvbxx", "jmvbxx"];
 
 
-triggers start => "pi","π", "pi to", "π to";
-triggers end => "digits of pi", "digits of π";
+triggers startend => "pi","π";
 
 # $PI now have 1 000 000 decimals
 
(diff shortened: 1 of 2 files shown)


moollaza:
@puskin94 thanks! BTW, for future PR's please append the name of the Instant Answer to the issue title so it's easier for everyone to know what it is you've improved the triggers for :)

puskin94:
@moollaza Passing the time on DDG, I learn complex things and I forget the simple one!

lib/DDG/Goodie/Pi.pm
```diff
@@ -15,7 +15,7 @@ attribution github => ["https://github.com/puskin94", "puskin94"],
             github => ["https://github.com/jmvbxx", "jmvbxx"];
 
 
-triggers startend => "pi", "digits of pi";
+triggers startend => "pi", "digits of pi","π","digits of π", "pi to", "π to";
```
mintsoft: Hmm this will work with queries like `4 pi to` and `digits of pi 10`; it might be worth switch this out into separate `triggers start` and `triggers end`

puskin94:
@mintsoft mmh, you are right. I made the changes and updated the test file too. Thanks for the review!

moollaza:
@puskin94 looking better but I noticed there's no tests for some of our trigger forms, i.e. "pi to X", which doesn't actually work right now :disappointed: 

And also, we should probably look for, and strip out "first" so that queries like "first 30 digits of pi" trigger this.

puskin94:
@moollaza Hei, I got `pi to X` working without changes:
pi-test1.
Is there any way to use as trigger a `start` trigger concatenated at a `end` trigger ( without using regex )? For i.e. if I use
`triggers start => "pi"` and `triggers end => "digits"` the query
`pi 30 digits` doesn't works... should I match the triggers using a regex?

moollaza:
>Is there any way to use as trigger a start trigger concatenated at a end trigger ( without using regex )?
Unfortunately this isn't possible.

>should I match the triggers using a regex?
You can certainly use a regex in your `handle` to make sure the query fits the pattern, but don't switch to a regex trigger as it's not necessary and degrades the performance.

puskin94:
@moollaza Thanks! I followed your advice and i updated the code... let me know if it can be good!

mintsoft:
@puskin94 The tests are broken now: https://travis-ci.org/duckduckgo/zeroclickinfo-goodies/jobs/54688565#L3904

puskin94:
@mintsoft It fails because $+{decimals} matches 3 if i use the `pi 3f5` query. How can I match ONLY numbers if the `(?<decimal>\d+)` is between the regex ? ( so I can't use `^` and `$` )

moollaza:
@puskin94 is this ready for review?

puskin94:
@moollaza the problem listed upper is still there. Idk if it can be considered a feature or a bug 
`It's Not a Bug, It's a Feature!`

lib/DDG/Goodie/Pi.pm
```diff
@@ -28,18 +28,16 @@ my $PI = '3.14159265358979323846264338327950288419716939937510582097494459230781
...
 
-handle remainder => sub {
+handle query_nowhitespace => sub {
 
-    my $decimal = $_;
-
-    return unless $decimal && $decimal =~ /^(\d+)$/ && $decimal > 0 && $decimal < $PI_max_digits;
-
-    my $answer = substr $PI, 0, ( $decimal + 2 );
+    return unless $_ =~ /(?:pi|π|to)?(?<decimal>\d+)(?:decimal|digits|decimals|digits of pi)?/i &&
```
moollaza: Looks like you have `digits` twice.

moollaza:
@puskin94 you can definitely use an anchored regex to match `pi 3 digits` and croak on `pi 3f digits` if you make the regex a non-greedy check for `\d+`

There's two other approaches:

1. Strip away parts of the query (regex string replace) until all you have left is the candidate string which we expect to be only `\d+` -- if that check fails (i.e. `3f`) return nothing
2. Similarly, generalize the regex to match `\b\w+\w` in the middle (our candidate number) and check if it's all digits. There's also Perl modules that can check if something looks like a number.

>Note: I highly suggest you checkout http://debuggex.com -- it's great

---

Spent a few minutes playing on debuggex -- this should work https://www.debuggex.com/r/iigOpLmZKXJr7ReF

`(?:(?:pi|π)\sto|first)?\s?(?<decimal>\d+)\s*(?:(?:decimal|digit)s?)\s+(?:of\s+(?:pi|π))?`

puskin94:
@moollaza really thanks for your time! I think that it is ready for the review ;)

jagtalon:
:+1:

jagtalon:
@puskin94 this fixes #1038 but looks like not https://github.com/duckduckgo/zeroclickinfo-goodies/issues/1037.

PR #1047, opened 2015-03-07
Adding use strict to all isAwesomes

mintsoft:
Just realised that I missed most of these in the last PR; we should be detected as 100% Perl 5 afterwards!


moollaza:
@mintsoft LGTM :+1:

mintsoft:
@moollaza It works!

image

Issue #1031, opened 2015-02-28
GreatestCommonFactor: Should support more than two inputs!


moollaza:
We can either improve the code manually, or switch to using a CPAN lib that already supports array input.

E.g. https://metacpan.org/pod/Acme::Tools#gcd

Likewise, the `$input` in the structured answer should be updated to show a comma separated list of inputs

PR #1055, opened 2015-03-11
Gcd

pfirsichbluete:
Adding additional trigger for GCD, and fixing
#1031 (GreatestCommonFactor: Should support more than two inputs!)

### DIFF
--- a/lib/DDG/Goodie/GreatestCommonFactor.pm
+++ b/lib/DDG/Goodie/GreatestCommonFactor.pm
@@ -19,12 +19,12 @@ attribution github => [ 'https://github.com/austinheimark', 'Austin Heimark' ];
 
 handle remainder => sub {
 
-    return unless /^\s*\d+(\s+\d+)*\s*$/;
+    return unless /^\s*\d+(?:(?:\s|,)+\d+)*\s*$/;
 
     # Here, $_ is a string of digits separated by whitespaces. And $_
     # holds at least one number.
 
-    my @numbers = grep(/^\d/, split /\s+/);
+    my @numbers = grep(/^\d/, split /(?:\s|,)+/);
     @numbers = sort { $a <=> $b } @numbers;
 
     my $formatted_numbers = join(', ', @numbers);
--- a/lib/DDG/Goodie/GreatestCommonFactor.pm
+++ b/lib/DDG/Goodie/GreatestCommonFactor.pm
@@ -7,7 +7,7 @@ use DDG::Goodie;
 zci answer_type => "greatest_common_factor";
 zci is_cached   => 1;
 
-triggers startend => 'greatest common factor', 'GCF', 'gcf';
+triggers startend => 'greatest common factor', 'gcf';
 
 primary_example_queries 'GCF 121 11';
 secondary_example_queries '99 9 greatest common factor';
--- a/lib/DDG/Goodie/GreatestCommonFactor.pm
+++ b/lib/DDG/Goodie/GreatestCommonFactor.pm
@@ -7,7 +7,7 @@ use DDG::Goodie;
 zci answer_type => "greatest_common_factor";
 zci is_cached   => 1;
 
-triggers startend => 'greatest common factor', 'gcf';
+triggers startend => 'greatest common factor', 'gcf', 'greatest common divisor', 'gcd';
 
 primary_example_queries 'GCF 121 11';
 secondary_example_queries '99 9 greatest common factor';
--- a/lib/DDG/Goodie/GreatestCommonFactor.pm
+++ b/lib/DDG/Goodie/GreatestCommonFactor.pm
@@ -19,14 +19,25 @@ attribution github => [ 'https://github.com/austinheimark', 'Austin Heimark' ];
 
 handle remainder => sub {
 
-    return unless /^(?<fn>\d+)\s(?<sn>\d+)$/;
+    return unless /^\s*\d+(\s+\d+)*\s*$/;
 
-    my ($first, $second) = sort { $a <=> $b } ($+{'fn'}, $+{'sn'});
-    my $result = gcf($first, $second);
+    # Here, $_ is a string of digits separated by whitespaces. And $_
+    # holds at least one number.
 
-    return 'Greatest common factor of ' . $first . ' and ' . $second . ' is ' . $result . '.',
+    my @numbers = grep(/^\d/, split /\s+/);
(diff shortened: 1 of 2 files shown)


lib/DDG/Goodie/GreatestCommonFactor.pm
```diff
@@ -19,14 +19,25 @@ attribution github => [ 'https://github.com/austinheimark', 'Austin Heimark' ];
...
-    return unless /^(?<fn>\d+)\s(?<sn>\d+)$/;
+    return unless /^\s*\d+(\s+\d+)*\s*$/;
 
-    my ($first, $second) = sort { $a <=> $b } ($+{'fn'}, $+{'sn'});
-    my $result = gcf($first, $second);
+    # Here, $_ is a string of digits separated by whitespaces. And $_
+    # holds at least one number.
 
-    return 'Greatest common factor of ' . $first . ' and ' . $second . ' is ' . $result . '.',
+    my @input = grep(/^\d/, split /\s+/);
```
moollaza: Might be good to split on `,` as well? e.g. `gcd 50, 100, 5`

moollaza:
@pfirsichbluete thanks a lot! We'll review this shortly

t/00-roles.t
```diff
@@ -475,7 +475,7 @@ subtest 'ImageLoader' => sub {
             use MIME::Base64;
             with 'DDG::GoodieRole::ImageLoader';
             our $tmp_dir = Path::Class::tempdir(CLEANUP => 1);
-            our $tmp_file = file(($tmp_dir->tempfile(TEMPLATE => 'img_XXXXXX', suffix   => '.gif'))[1]);
+            our $tmp_file = file(($tmp_dir->tempfile(TEMPLATE => 'img_XXXXXX', SUFFIX => '.gif'))[1]);
```
moollaza: Looks like this shouldn't be here? The related PR that added this has now been merged.

lib/DDG/Goodie/GreatestCommonFactor.pm
pfirsichbluete: Done in d32098213250d428f0a2e4aa04953b7ecbdac01b

Pull Request updated.

t/00-roles.t
pfirsichbluete: That should be fine.

This PR is based on the other (for 846d971) to allow tests to pass.
Hence, the 846d971 commit is in both PRs.
Since the other PR is now merged, 846d971 already has a "green check" icon in the top list
of commits.

If you do not trust what github is showing, let me know, and I'll rebase this PR on master.

pfirsichbluete: This PR is base on the other PR to make tests pass.

Since it's not only the same code, but really the same commit, this should not be an issue.

But I rebased the PR to make that commit vanish from this PR.

moollaza:
@pfirsichbluete this looks great. I've run into a small problem while testing this (I don't believe it's your fault!) so I'll hold off on merging until I know for sure.

mintsoft:
@pfirsichbluete @moollaza One quick question; what does this output in the event that there *are* no common factors?

moollaza:
@mintsoft it seems like it only handles positive integers, so I think there will always be a valid GCF?

lib/DDG/Goodie/GreatestCommonFactor.pm
```diff
@@ -19,14 +19,25 @@ attribution github => [ 'https://github.com/austinheimark', 'Austin Heimark' ];
...
+    foreach (@numbers_to_consider) {
+        $result = gcf($result, $_)
+    }
+
+    my $description = 'Greatest common factor of ' . join(', ', @input) . ' is ' . $result . '.';
+    $description =~ s/, ([^,]*)$/ and $1/;
+    return $description,
       structured_answer => {
-        input     => [$first, $second],
+        input     => [@input],
```
zachthompson: This should be the original input html encoded (or the join(', ', @input) string)

pfirsichbluete: @zachthompson The list of the number looks fine to me.

The original code used the list of numbers without issues.

And [docs](https://duck.co/duckduckhack/goodie_advanced_handle_functions#using-structured-responses) only call for
```
  "What the input was, usually highlighting how it has been parsed"
```

The list of the numbers seems to highlight nicely how it has been parsed. In fact it exposes it better than anything else.

It seems basically all (only exception being the `-ANY-` inputs) goodies use lists for `input` too.
And other goodies are also using lists with >1 elements too without issues. Like for example [Password](https://github.com/duckduckgo/zeroclickinfo-goodies/blob/master/lib/DDG/Goodie/Password.pm#L88), or [RandomNumber](https://github.com/duckduckgo/zeroclickinfo-goodies/blob/master/lib/DDG/Goodie/RandomNumber.pm#L49).

pfirsichbluete:
@mintsoft I agree with @moollaza that the gcd should always exist for the non-negative numbers
matched by this goodie.

I added tests for the degenerate cases in commit 3bb5780 to make this more explicit.
Does this address your question, or did you have some other specific setting in mind?

lib/DDG/Goodie/GreatestCommonFactor.pm
zachthompson: @pfirsichbluete The reason all of those work is because they have less than 3 inputs.  We handle auto-templating (which is what structured_answer is telling the system to do) for 0-2.  If you wanted to do something with the individual inputs in a custom template on the frontend then, yes, it might make sense to pass them that way.

mintsoft:
@pfirsichbluete No that answers my question, I didn't spot that it only accepted positives!

lib/DDG/Goodie/GreatestCommonFactor.pm
moollaza: @pfirsichbluete once the input is fixed (simply joining the numbers with a `,` should be perfect) this is ready for merge!

pfirsichbluete: @zachthompson Argh. True. Thanks. Done.

pfirsichbluete: I tried to make the “<3 items” requirement for auto-templating more visible by
* https://github.com/duckduckgo/duckduckgo-documentation/pull/227
* https://github.com/duckduckgo/p5-app-duckpan/pull/240

moollaza:
LGTM :+1:

PR #1056, opened 2015-03-11
Fix casing of 'suffix' argument for temp files in tests

pfirsichbluete:
The role tests failed with Perl <5.17.10 installs, due to the used
File::Temp (<0.23) still caring about the casing of the
arguments. Newer versions of File::Temp no longer care about the
casing.

By uppercasing 'suffix' in the tests, we make the tests pass for both
Perl 5.16 and Perl 5.18.


moollaza:
LGTM :+1:

mintsoft:
Nice catch @pfirsichbluete

PR #1066, opened 2015-03-17
Added SHA3 Goodie

rafacas:
Computes the SHA-3 cryptographic hash function.
Users interested in the SHA3 hash of a string will be able to get it as an Instant Answer instead of going to a specific page that calculates it or writing a small script.
None
N/A
N/A
sha3 this string
sha3sum this string
sha3-224 hash this string
sha3-256 hash of this string
sha3-384 this string
sha3-512 this string
shake128 this string
shake256 this string
People interested in cryptography or users who already use the Instant Answer MD5 or SHA.
No
None

sha3

## Checklist

```
- [x] Tested cross-browser compatibility
    - Mac OSX
        - [x] Google Chrome
        - [x] Firefox
```


lib/DDG/Goodie/SHA3.pm
```diff
@@ -0,0 +1,71 @@
...
+                          "sha3-384 hash of this string",
+                          "shake128 this string",
+                          "shake256 this string";
+category "calculations";
+topics "cryptography";
+code_url "https://github.com/duckduckgo/zeroclickinfo-goodies/blob/master/lib/DDG/Goodie/SHA3.pm";
+attribution github => ["rafacas", "Rafa Casado"],
+            twitter => "rafacas";
+
+my @triggers = qw(sha3 sha3sum sha3-224 sha3-256 sha3-384 sha3-512 shake128 shake256);
```
mintsoft: It doesn't quite look like these triggers match the regex below. The regex will match "shake-256" whilst the triggers won't

mintsoft:
@rafacas Thanks for this! You're clearly on a hashing-spree!

lib/DDG/Goodie/SHA3.pm
```diff
@@ -0,0 +1,71 @@
...
+
+    my $alg_name  = $alg eq "sha3" ? $alg . '_' : $alg; # The functions prefix for sha3 is "sha3_"
+    my $func_name = 'Digest::SHA3::' . $alg_name . $ver . '_' . $enc;
+    my $func      = \&$func_name;
+
+    my $out = $func->($str);
+
+    # By convention, CPAN Digest modules do not pad their Base64 output. So any
+    # necessary padding will be implemented here
+    if ($enc eq 'base64'){
```
mintsoft: Btw, didn't mention this previously but you could do you know about the `x` operator?
http://perldoc.perl.org/perlop.html#Multiplicative-Operators

rafacas:
Hi @mintsoft. Yes, kind of a hash-spree, but I promise this is the last Goodie hash-related ;)

lib/DDG/Goodie/SHA3.pm
rafacas: Thanks! I didn't know about the 'repetition' operator (```x```) in perl. I've changed the while loop for it. Do you want me to change it in the other PRs too? I guess it would be good for consistency among Goodies.

lib/DDG/Goodie/SHA3.pm
rafacas: Thanks! I didn't know about the 'repetition' operator (```x```) in perl. I've changed the while loop for it. Do you want me to change it in the other PRs too? I guess it would be good for consistency among Goodies.

lib/DDG/Goodie/SHA3.pm
rafacas: I've written the regex to check valid queries only after they have been triggered. ```shake-256``` will never be triggered so it doesn't matter if the regex matches it or not (it will never happen).
In any case, If you think it is better to change the regex to match exactly the triggers I don't mind changing it. I just didn't do it because I thought it wasn't necessary :)

mintsoft: Is there a reason why `shake-256` isn't supported? I often use `AES 128` `AES128` and `AES-128` interchangeably for example

rafacas: The only reason was that I hadn't seen it use it that way. I've just looked it up and I've seen people using shake-128 and shake-256 (instead of shake128 and shake256) so I'll add them.

javathunderman:
@mintsoft Are we still waiting on that dependency before this gets merged?

MrChrisW:
@rafacas Looks like there's a merge conflict with the `dist.ini` file! Please merge with master to fix the conflict :+1:

rafacas:
Hi @MrChrisW! I've rebased it from master so the ```dist.ini``` conflict is fixed :)

mintsoft:
@javathunderman Yeah, waiting on the go-ahead from @moollaza or @jagtalon that the dependency has been added to the environments

moollaza:
@rafacas sorry about the delay, the dep still needs to be added, should be done this week. There's a merge conflict now though -- we recently updated the dist.ini to use AutoPrereqs (so you don't need to manually update it. Can you merge in master (and remove the dist.ini update) please?

rafacas:
Hi @moollaza. I've just merged in master and fixed the dist.ini conflict.

moollaza:
@rafacas thanks, will merge as soon as the team adds in the deps :+1:

moollaza:
Deps added :+1:

moollaza:
@rafacas congrats! It's live: https://duckduckgo.com/?q=sha3+this+string&ia=answer

PR #1040, opened 2015-03-05
New IA idea for onion addresses

ilv:
**What does your Instant Answer do?**
Detect when someone searchs an onion address that refers to an HTTP onion/hidden service and provide quick links to access it.

**What problem does your Instant Answer solve (Why is it better than organic links)?**
It provide quick links to access HTTP onion/hidden services. Normally, it would take several clicks to find how to access .onion addresses if you are not familiar with it.

**What are some example queries that trigger this Instant Answer?**
Example 1: https://3g2upl4pq6kufc4m.onion
**Example 2: How do I access https://3g2upl4pq6kufc4m.onion/?**
**Example 3: What is https://3g2upl4pq6kufc4m.onion:5000/?q=dont+track+us?**

**Which communities will this Instant Answer be especially useful for? (gamers, book lovers, etc)**
People interested in privacy, freedom of speech, anti-censorship, whistleblowing.

**Is this Instant Answer connected to a DuckDuckHack Instant Answer idea?**
Nope.

**What does the Instant Answer look like? (Provide a screenshot for new or updated Instant Answers)**
ddg_onion_address_ia

Checklist

Please place an 'X' where appropriate.

- [X] Added metadata and attribution information (https://duck.co/duckduckhack/metadata)
- [X] Wrote test file and added to t/ directory (https://duck.co/duckduckhack/goodie_tests)
- [X] Verified that instant answer adheres to design guidelines (https://duck.co/duckduckhack/design_styleguide)
- [X] Verified that instant answer adheres to code styleguide (https://duck.co/duckduckhack/code_styleguide)
- [] Tested cross-browser compatibility

    Please let us know which browsers/devices you've tested on:
    - So far I've only tested it on the Virtual Box virtual machine provided for development.

**Note**: The proposed IA provides a link to visit an HTTP onion service via Tor2web. It seems that for the moment this service is not working properly, so it's possible you won't be able to access the examples via Tor2web.

-----
IA Page: https://duck.co/ia/view/onion_address

PR #1067, opened 2015-03-17
New IA idea for GetTor (tips to download the Tor Browser and get bridges)

ilv:
It gives useful tips to get the Tor Browser and bridges, especially for users that don't have access to Tor Project's official website (TPO).

If you can't access TPO (i.e. it's blocked), then is probably that you won't know these tips. DDG will provide an alternative source of information for people looking to circumvent censorship.

Example 1: get tor
Example 2: get tor browser
Example 3: get bridges
Example 4: get tor bridges

People that live and/or work in highly censored places.

It's not.

None that I'm aware of.

I'm not sure if the amount of text displayed is reasonable. What do you think?

ddg-gettor

ddg-getbridges

## Checklist

```
- [X] Verified that instant answer adheres to design guidelines (https://duck.co/duckduckhack/design_styleguide)
- [X] Verified that instant answer adheres to code styleguide (https://duck.co/duckduckhack/code_styleguide)

    - So far I've only tested it on the Virtual Box virtual machine provided for development.

### DIFF
--- a/lib/DDG/Goodie/GetTor.pm
+++ b/lib/DDG/Goodie/GetTor.pm
@@ -1,0 +1,46 @@
+package DDG::Goodie::GetTor;
+# ABSTRACT: Tips to get the Tor Browser and bridges for places where access to Tor Project's website is blocked.
+
+use strict;
+use DDG::Goodie;
+
+zci is_cached => 1;
+zci answer_type => "get_tor";
+
+primary_example_queries 'get tor browser';
+secondary_example_queries 'get bridges';
+description "Tips to get the Tor Browser and bridges for places where access to Tor Project's website is blocked.";
+name 'GetTor';
+code_url 'https://github.com/duckduckgo/zeroclickinfo-goodies/blob/master/lib/DDG/Goodie/GetTor.pm';
+category 'special';
+topics 'social', 'special_interest', 'cryptography';
+attribution email => 'ilv@torproject.org';
+
+# this triggers should cover the space search for someone that wants to
+# download the Tor Browser or get some bridges
+triggers any => 'gettor',
+                'get tor',
+                'gettor browser',
+                'gettor bridges',
+                'get bridges',
+                'tor bridges',
+                'tor browser',
+                'torbrowser';
+
+handle query_lc => sub {
+    my $html_output = "<div class='zci__caption'>";
+
+    if ($_ =~ m/bridges?/) {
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by ilv** [comment]
Some links about the current approaches to deliver the Tor Browser and bridges:

1) [GetTor](https://bridges.torproject.org/)
2) [BridgeDB](https://bridges.torproject.org/)

**Comment by moollaza** [comment]
@ilv thanks a lot for contributing, we really appreciate it! We'll be sure to review this shortly and give you some feedback!

**Comment by ilv** [comment]
@moollaza Thanks to you people! I look forward to your feedback!

**Comment by isislovecruft** [comment]
@ilv Perhaps it would be better to direct users to the recommended pages at https://bridges.torproject.org and https://www.torproject.org/download/download-easy.html first, rather than telling them about the failsafe distribution mechanisms as the "default" answer? Or perhaps we should tell them about the email options if the user's query was more like "tor browser website not working"?

**Comment by ilv** [comment]
@isislovecruft Thanks for the feedback. I agree on adding a reference to the recommended pages (will work on it during the week). I do think we should keep the same triggers, **and** add more triggers of the form "tor browser website not working", "bridge website is blocket", etc.

**Comment by moollaza** [comment]
@ilv terribly sorry for the delay. I'll try and get around to this soon!

File: lib/DDG/Goodie/GetTor.pm [hunk 1]
```diff
@@ -0,0 +1,46 @@
...
+triggers any => 'gettor',
+                'get tor',
+                'gettor browser',
+                'gettor bridges',
+                'get bridges',
+                'tor bridges',
+                'tor browser',
+                'torbrowser';
+
+handle query_lc => sub {
```
moollaza: Goodie templates have now been documented. Let's switch this over to use the `text` template!

https://duck.co/duckduckhack/goodie_displaying

moollaza: Let's verify the query fits our expected structure using a regex, or at the very least, lets make a check that the remiander of the query doesn't contains things we don't want.

The regex approach is the stronges:

```perl
return unless m/(?:how can I|how to)?(?:get ?|download ) tor(?: bridges|browser)?/;
```

Something like that^ should work. I haven't tested it, just writing that on-the-fly!

ilv:
@moollaza no worries. I'll try to take a look at your notes soon, probably next week.

mintsoft:
Already merged in #1040

PR #1068, opened 2015-03-18
Remove dynamic generation of .travis.yml from dzil

zachthompson:
* Remove the generation of .travis.yml from dzil.
* Add Travis container routing to .travis.yml so that we can hopefully speed things up and take advantage of caching.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,11 +1,12 @@
 ---
+sudo: false
+addons:
+  apt_packages:
+    - libmpfr-dev
 before_install:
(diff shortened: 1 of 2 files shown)


moollaza:
@zachthompson LGTM -- looks like the container based builds take almost a minute less in some cases

mintsoft:
@moollaza @zachthompson I presume this is related the conversation previously about moving to docker for Travis builds?

moollaza:
@mintsoft correct -- we're trying to cleanup and get rid of as much unnecessary Dzil stuff as possible to speed up builds and simplifying things.

PR #1070, opened 2015-03-18
Wavelength: Fix string interpolation in test

pfirsichbluete:
'\S' is not a valid escape and makes prove complain to stderr (though
the tests pass). Since it seems the '\S' got only introduced to
separate the variable from the word 'Speed', we proper separate them
by switching from the unrecognized escape to proper curly braces.

### DIFF
--- a/t/Wavelength.t
+++ b/t/Wavelength.t
@@ -37,7 +37,7 @@ sub mk_test {
         $expect,
         structured_answer => {
             input => [],
-            operation => "Wavelength of $freq_value $freq_units ($vf_text\Speed of light in a vacuum)",
+            operation => "Wavelength of $freq_value $freq_units (${vf_text}Speed of light in a vacuum)",
             result => $expect,
         }
     );


t/Wavelength.t
```diff
@@ -37,7 +37,7 @@ sub mk_test {
         $expect,
         structured_answer => {
             input => [],
-            operation => "Wavelength of $freq_value $freq_units ($vf_text\Speed of light in a vacuum)",
+            operation => "Wavelength of $freq_value $freq_units (${vf_text}Speed of light in a vacuum)",
```
mintsoft: I think this is actually supposed to be a `\` literal, so I would think:
```perl
operation => "Wavelength of $freq_value $freq_units (${vf_text}\\Speed of light in a vacuum)"
```
Is what was supposed to be there

pfirsichbluete: > I think this is actually supposed to be a \

I doubt it, because the main Goodie code does not have any `\` in there, so the test probably
is not expected to have either. The relevant line from the Goodie is:
```
my $velocity_text = ($vf == 1 ? '' : "$vf × ").'Speed of light in a vacuum';
```
-------------------------

Also `$vf_text` is either empty or something like '5 x ' (like  'five times'). Hence a `\` between
that and `Speed of light in a vacuum` would not make a meaningful statement for me.

mintsoft: OH right, yeah! Sorry I totally misinterpreted what was going on there!

mintsoft:
Thanks for the fix @pfirsichbluete

Issue #402, opened 2014-05-07
AltCal: no results for example query


jdorweiler:

moollaza:
This one should be converted to a structured answer, or it could use the text template!

marianosimone:
Seems like the code for this was already converted to a structured answer, but it's somehow disabled.

Anything we can do to help get it released (or discarded)?

zekiel:
@rafacas  do you need any particular help?

jdorweiler:
This is it? https://duckduckgo.com/?q=meiji+1&ia=answer

javathunderman:
@marianosimone any update on this?

jdorweiler:
Looks like it works now.

PR #1078, opened 2015-03-25
AltCalendars converted to a structured answer (#402)

rafacas:
Converted the AltCalendars Goodie and its tests to structured answer.

### DIFF
--- a/lib/DDG/Goodie/AltCalendars.pm
+++ b/lib/DDG/Goodie/AltCalendars.pm
@@ -42,25 +42,32 @@ handle query_parts => sub {
         my $year = $eras{$era_name}[0] + $era_year;
         my $result = $1.$year.$4;
         my $wiki = 'https://en.wikipedia.org/wiki/';
-        my $answer;
 
-        if ($result =~ /^[0-9]{4}$/) {
-            $answer = "$era_name $era_year is equivalent to $year in the Gregorian Calendar";
-        } else {
-            $answer = "$result ($era_name $era_year is equivalent to $year in the Gregorian Calendar)";
-        }
-
-        my $answer_html = $answer.'<br><a href="'.$wiki.$eras{$era_name}[1].'">More at Wikipedia</a>';
+        my $answer = "$era_name $era_year is equivalent to $year in the Gregorian Calendar";
 
         return $answer,
             structured_answer => {
-            input     => [html_enc($era_name . ' ' . $era_year)],
-            operation => html_enc('Gregorian conversion of ' . $era_name . ' ' . $era_year),
-            result    => html_enc($answer_html)
+                id => 'altcalendars',
+                name => 'Calendar',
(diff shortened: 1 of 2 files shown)


moollaza:
@rafacas thanks a lot for this!

If you're really interested in making this one awesome, you could take a stab at converting it to use the new Goodie templates? We don't have any docs yet (still in the works but you will want to mimic the [Minecraft Goodie](https://github.com/duckduckgo/zeroclickinfo-goodies/blob/master/lib/DDG/Goodie/Minecraft.pm#L54)

The difference is that you can now pass data along to one of our built-in templates (the ones we currently use for Spice IAs). In this case, I'd suggest the `text` template. And you can automatically generate the More at Wikipedia link now by setting the `sourceUrl` and `sourceName` in the `Meta` section of the `structured_answer` object.

moollaza:
@rafacas haha I just remembered why this one wasn't a structured answer -- it returns HTML which right now get's escaped. Even though you've encoded it, we don't treat the text as raw HTML so if you run it in DuckPAN you'll see the result is broken:

2015-03-25_18h34_06

Looks like this one *needs* to be moved to a new Goodie template as I've described above!

moollaza:
I suppose the alternative is to simply get rid of the wiki URL -- I'm fine with that as well.

rafacas:
@moollaza haha, I usually test it with ```duckpan query``` and ```prove```. I'll start using ```duckpan server``` too.
I think I'm going to try the new ```Text``` template :)

moollaza:
@rafacas awesome! Thanks a lot :)

rafacas:
Hi @moollaza I just updated the branch with the ```Text``` template.

Here is a screenshot:

screen shot 2015-03-26 at 16 50 03

BTW, the templates are pretty cool ;)

moollaza:
@rafacas this is great -- let's switch our title and subtitle around though. The actual answer should be the most prominently displayed item as it's the answer. The context (Meiji Calendar Conversion) is less important so it should be the subtitle.

rafacas:
Sorry @moollaza, I've just seen your comment. I'll change it and wait for @abeyang to comment on it.

rafacas:
I've just changed the name from ```Calendar``` to ```Calendar Conversion```.

screen shot 2015-03-27 at 12 19 18

abeyang:
Hi @rafacas , it's almost there. Usually, for the subtitle, it would be ideal to see what the converted date is (in this case, "Meiji 42"). Thus, can we do something like this (notice I removed "calendar"):

> **1909**
> Meiji 42 - Equivalent Gregorian Year

What do you think?

rafacas:
@abeyang That makes perfect sense :)

I've just updated it. Here is a screenshot:
screen shot 2015-03-27 at 16 46 34

abeyang:
Looks good! :+1:

PR #1082, opened 2015-04-02
Chess960: Use HTML output provided by FenViewer

rouzbeh:
This changes the output of the Chess960 IA to use the same format as FenViewer, purely done in HTML instead of querying an external website for an image to display.

### DIFF
--- a/lib/DDG/Goodie/Chess960.pm
+++ b/lib/DDG/Goodie/Chess960.pm
@@ -3,7 +3,7 @@ package DDG::Goodie::Chess960;
 
 use strict;
 use DDG::Goodie;
-use DDG::Goodie::FenViewer qw(parse_position draw_chessboard_html draw_chessboard_ascii);
+with 'DDG::GoodieRole::Chess';
 
 triggers any => 'random', 'chess960';
 zci is_cached => 0;
--- a/lib/DDG/Goodie/Chess960.pm
+++ b/lib/DDG/Goodie/Chess960.pm
@@ -3,6 +3,7 @@ package DDG::Goodie::Chess960;
 
 use strict;
 use DDG::Goodie;
+use DDG::Goodie::FenViewer qw(parse_position draw_chessboard_html draw_chessboard_ascii);
 
 triggers any => 'random', 'chess960';
 zci is_cached => 0;
@@ -115,6 +116,13 @@ RBKRNQBN RKRBNQBN RKRNQBBN RKRNQNBB BBRKRNNQ BRKBRNNQ BRKRNBNQ BRKRNNQB RBBKRNNQ
 RKBRNBNQ RKBRNNQB RBKRBNNQ RKRBBNNQ RKRNBBNQ RKRNBNQB RBKRNNBQ RKRBNNBQ RKRNNBBQ RKRNNQBB
 );
 
+sub make_fen {
+    my ($position) = @_;
+    my ($position_lc) = lc$position;
+    my ($fen) = "${position_lc}/pppppppp/8/8/8/8/PPPPPPPP/${position}";
+    return parse_position($fen);
+}
+
 handle query => sub {
     # Ensure rand is seeded for each process
     srand();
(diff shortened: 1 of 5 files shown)


lib/DDG/Goodie/Chess960.pm
```diff
@@ -3,6 +3,7 @@ package DDG::Goodie::Chess960;
 
 use strict;
 use DDG::Goodie;
+use DDG::Goodie::FenViewer qw(parse_position draw_chessboard_html draw_chessboard_ascii);
```
mintsoft: Hmm, we don't usually cross-goodie access like this; for situations where we share components we create roles. However that may be a little industrial for this, how many Chess related goodies can there be!?

lib/DDG/Goodie/Chess960.pm
```diff
@@ -115,6 +116,13 @@ RBKRNQBN RKRBNQBN RKRNQBBN RKRNQNBB BBRKRNNQ BRKBRNNQ BRKRNBNQ BRKRNNQB RBBKRNNQ
 RKBRNBNQ RKBRNNQB RBKRBNNQ RKRBBNNQ RKRNBBNQ RKRNBNQB RBKRNNBQ RKRBNNBQ RKRNNBBQ RKRNNQBB
 );
 
+sub make_fen {
```
mintsoft: Is this parsing different to the FenViewer IA's parsing due to a differing input format only?

lib/DDG/Goodie/FenViewer.pm
```diff
@@ -8,6 +8,8 @@ use DDG::Goodie;
 use strict;
 use Scalar::Util qw(looks_like_number);
 use Try::Tiny;
+use Exporter qw(import);
```
mintsoft: Not sure if `Exporter` is in `dist.ini`; can you add it if it is?

t/Chess960.t
```diff
@@ -13,19 +13,15 @@ ddg_goodie_test(
...
-       ([BKNQRP] ?){0,8}
-
-\(where B is for bishop,
-       K is for king,
-       N is for knight,
-       Q is for queen,
-       R is for rook, and
-       P is for pawn\)$/,
-            html    => qr|^<img src='/iu/\?u=http://www\.apronus\.com/chess/stilldiagram.php\?d=[KNQRPB]{17}________________________________[knqrpb]{16}0\.jpg&w=8&h=8'/>$|,
+        $_ => test_zci(qr/^[qrkbn]{1,8}
```
mintsoft: Is there a reason that the link to `apronus.com` has been removed?

lib/DDG/Goodie/Chess960.pm
mintsoft: @moollaza @jagtalon Any thoughts on this? 
* Does it present a problem with deployments etc?
* Is there an issue having dependencies between Goodies?
* Would a role be preferable?

mintsoft:
Thanks for this @rouzbeh. Love anything that harmonizes things a bit! I'd like to get an alternative viewpoint around the dependencies but on-the-whole, it looks good!

lib/DDG/Goodie/Chess960.pm
moollaza: @mintsoft I don't actually know if it'll cause any problems. I'll have to run this on a dev machine when I get a chance to see how it all works. 

I think using a role would be fine though, at least we know it won't cause problems :)

rouzbeh:
I'm going to try to answer to everything at once. Apologies if I forget a point.

@mintsoft I was not sure what would be the best way to do this. Just exporting the functions and importing them here seems to work. I could also just copy/paste them, but that's probably not the best thing to do. @moollaza How would I go about making this into a role? 

About chess related goosdies, I guess we can think of a few. Another example that comes to mind is a goodie that shows particular opening moves on the board, when asked by them name. Something like "Chess opening French defense".

About the parsing, in Chess960 we basically only record the position on the last (or first) row, and the rest is always the same. For FEN, we need to explicitly provide the other rows (of the board) as well. In the current goodie, the parsing is basically handled by apronous, which sends a jpeg back. Here we would do parsing directly, and not bother with an external service. Hence its removal.

And finally, what do I need to do to dist.ini?

mintsoft:
@rouzbeh For the dist.ini you just need to add `Exporter = 0` under the `[Prereqs]` section.

I think if there's scope for more chess goodies then a Role makes the most sense. Essentially we can have a module designated for parsing and displaying chessboards based on your FenViewer work!

We don't really have any docs for this as I think only myself and an ex-staff member have actually made goodie roles! However you essentially move the core code into `/lib/DDG/GoodieRole/Chess.pm`, adding `use Moo::Role;`. Then any subs that are defined will be accessible from any package using the role. 

In the Chess960 or FenViewer Goodies you simply add `with 'DDG::GoodieRole::Chess';` to compose the role into the package.

That's a little vague, however an example is:
* https://github.com/duckduckgo/zeroclickinfo-goodies/blob/master/lib/DDG/Goodie/CalendarConversion.pm
* https://github.com/duckduckgo/zeroclickinfo-goodies/blob/master/lib/DDG/GoodieRole/Dates.pm

Both of these are pretty big but I hope you can see how it all fits together. Any problems just ping me and I'll take a look!

moollaza:
>@rouzbeh For the dist.ini you just need to add Exporter = 0 under the [Prereqs] section.

@rouzbeh @mintsoft we actually *just* switched to use `[AutoPrereqs]` so developers no longer have to update the dist.ini :)

mintsoft:
@moollaza that's awesome

rouzbeh:
@rouzbeh @mintsoft I thought I don't need to bother with dist.ini since I went with the role anyways. Is that wrong? Do I need to do anything else?

mintsoft:
@rouzbeh That is correct!

moollaza:
Hey guys, just checking in here -- is this ready for review/merge?

rouzbeh:
Hey @moollaza. As far as I am concerned, yes!

mintsoft:
@moollaza I want to merge this in as I think it's worth having this html approach in master as it'll make it easier to convert to templates.

moollaza:
Sorry for they delay guys! This LGTM :+1:

PR #1085, opened 2015-04-04
Zaahir/cheatsheets working

moollaza:
Just about finished. The mobile layout needs help though.

I'm using css to display only the first 3 items, for the first 2 sections. It works for now, probably doesn't work for IE8 though. We can revise that and use JS + have the dev signify which elements should be show. The CSS route seemed simplest for now.


moollaza:
Merging for the Hackathon, can revise later.

lib/DDG/Goodie/CheatSheets.pm
```diff
@@ -0,0 +1,56 @@
...
+description 'Cheat sheets';
+primary_example_queries 'universal help', 'universal cheat sheet', 'universal example';
+code_url 'https://github.com/duckduckgo/zeroclickinfo-goodies/blob/master/lib/DDG/Goodie/CheatSheets.pm';
+category 'cheat_sheets';
+topics qw'computing geek programming sysadmin';
+attribution github  => [zachthompson => 'Zach Thompson'];
+
+
+triggers startend => (
+	'char', 'chars',
```
mintsoft: @jagtalon wouldn't be happy if he saw these tabs!

mintsoft:
As just discussed with @moollaza the regex cheatsheet goodie also provides some more targeted help: https://duckduckgo.com/?q=regex+%5Bb-j%5D&ia=answer or https://duckduckgo.com/?q=regex+%7B4%2C9%7D&ia=answer

Would the cheatsheets be extended to handle this or does it make more sense to keep regex separate? I suppose a third option would be to move the cheatsheet component and keep the interactive IA separate.

jagtalon:
@moollaza should JSON::XS be in dist.ini?

zachthompson:
@jagtalon @moollaza https://github.com/duckduckgo/zeroclickinfo-goodies/pull/1100

@mintsoft Not quite sure yet if we want to just keep CheatSheets simple and consistent or allow for cheat sheet-specific subs to somehow modify the json.

mintsoft:
@zachthompson fair enough, feel free to ping me for an outside opinion if wanted!

zachthompson:
@mintsoft outside opinions are our MO.  Let us know if you have ideas about how to move CheatSheets forward.

PR #1089, opened 2015-04-04
New cheat sheet for GNU Screen

dnmfarrell:
https://duck.co/ia/view/gnu_screen_cheat_sheet


mintsoft:
@dnmfarrell as @jagtalon said, you could have kept the previous PR open. Any commits that are pushed to your `master` will get automatically added to this PR. It's handy for us to keep the discussions around individual together etc :+1:

dnmfarrell:
Ok, finally the CSS is loading so the cheat sheet text shows in two columns. See what you think!

jagtalon:
@dnmfarrell I like it. Adding @abeyang in for a design review.

zekiel:
ping @abeyang

moollaza:
@dnmfarrell hey there! I'm very sorry it's taken me so long to get around to this. We've been hard at work lately on CheatSheets in particular, and I've finally finished up the next version of CheatSheets. There will be documentation and a proper announcement coming soon, but we've greatly simplified the way CheatSheets work and you no longer need to touch any Perl or HTML. All that's needed now is a JSON file in `/share/Goodie/CheatSheets/`. In this case you would need to make `gnu_screen.json`.

Would you be interested in converting this to a JSON file? We've already converted a few Goodies to give you guys some examples: https://github.com/duckduckgo/zeroclickinfo-goodies/blob/master/share/goodie/cheat_sheets/tmux.json

moollaza:
ping! @dnmfarrell the CheatSheets are now live and so is the updated CheatSheet Documention. Would you be interested in updating this to use the new JSON file? It will greatly simplify this IA -- only requires creating 1 new file, rather than 5 :)

https://duck.co/duckduckhack/goodie_cheat_sheets

dnmfarrell:
OK I backed out the old code and added a `JSON` version instead. Have a look, see what you think!

share/goodie/cheat_sheets/gnu_screen.json
```diff
@@ -0,0 +1,116 @@
...
+            "key": "List sessions"
+        },
+        {
+            "val": "screen -r [name]",
+            "key": "Re-attach session"
+        }
+        ],
+        "Copying and Scrolling": [
+        {
+            "val": "Ctrl-a, [",
```
moollaza: I think you'll want to to wrap these in `[]` and/or `{}`, so that the individual keys are wrapped in `<code></code>`

```json
"val": "[Ctrl-a] {\\[}",
"val": "[Ctrl-a], [Space]"
```

*Note*: You'll need to escape any `[]{}` that are inside a pair of `[]` or `{}` using `\\`

moollaza: Make sure you test it in DuckPAN to see the difference, or take a peak at https://duckduckgo.com/?q=tmux+cheatsheet&t=ffab&ia=cheatsheet

each separate block has been wrapped: https://github.com/duckduckgo/zeroclickinfo-goodies/blob/master/share/goodie/cheat_sheets/tmux.json

moollaza:
@dnmfarrell aside from wrapping these in `[]` everything is looking good!

dnmfarrell:
I've corrected the markup and tested it using Duckpan server. Looks good to me. One issue though is I can only get the cheat sheet to appear if I typed `gnu_screen keyword` (with the underscore). Is there a `.json` attribute I can add for matching names? (e.g. "gnu screen", "screen" etc)

mintsoft:
@dnmfarrell if you symlink the aliases to the sheet it'll work.It's how the regex sheet is done: https://github.com/duckduckgo/zeroclickinfo-goodies/tree/master/share/goodie/cheat_sheets a

moollaza:
@dnmfarrell we actually check for `-` in the filename and split on that. I'm going to update that to check for `_` because most of our filenames use underscores instead of dashes. With that fixed it should trigger on `gnu screen cheat sheet`

moollaza:
actually I misspoke, there's no need for the `_` in your filename, just use spaces :+1:, i.e. `gnu screen.json`

dnmfarrell:
Spaces didn't work for me. But when I replaced the space with a dash, it triggers fine now. I think it's safer to use "gnu screen" rather than just "screen" as I bet a bunch of people are searching about their computer monitor rather than the software. Do you agree? It does mean that the cheatsheet it less likely to trigger ...

moollaza:
LGTM :+1: 

@dnmfarrell I think we can keep it this way for now and see what kind of reaction it gets. If users want it to simply trigger on "screen" I see no harm in allowing it, but I do see potentially for false-positives in the case of "screen help" or similar.

moollaza:
@dnmfarrell Thanks again for taking the time to contribute! We really appreciate it. Feel free to stick around and comment/help on other PR's & Issues or even submit more Instant Answers!

dnmfarrell:
No problem, thanks for working with me to get it fixed :)

moollaza:
@dnmfarrell congrats! It's live: https://duckduckgo.com/?q=gnu+screen+cheat+sheet&ia=cheatsheet

dnmfarrell:
Thanks for your help!

PR #1112, opened 2015-04-21
Created Weight Goodie

wongalvis:
# Goodie Pull Request Template

Using this template will help us better understand your Instant Answer and assist you when necessary.  Simply copy and paste the markdown below into the description of your GitHub pull request and complete as appropriate.

Calculate the weight of provided mass (default in kg).

It calculates the weight with the value of the mass provided by the user.

http://physics.nist.gov/cgi-bin/cuu/Value?gn (For taking value of acceleration due to gravity on Earth as 9.80665m/s^2)

It is from a trustworthy authority (National Institute of Standards and Technology).

Yes there might be more accurate values of g (values with more decimal places).

"Weight 5", "Weight 5.12g", "5.1 lb weight on Earth", "Weight of 5.12"

Students or people who study physics.

I'm not 100% sure whether it overlaps with one but it should be not.

I had some problems dealing with perl and regex as a first time perl coder but eventually the code with the difficult parts were removed.

screencapture-tempo-rudolf-codio-io-5000-1429699724232

## Checklist

```

    - Windows 8
        - [X] Google Chrome
        - [] Firefox
        - [] Opera
        - [] IE 10

    - Windows 7
        - [] Google Chrome
        - [] Firefox
        - [] Opera
        - [] IE 8
        - [] IE 9
        - [] IE 10

    - Windows XP
        - [] IE 7
        - [] IE 8

    - Mac OSX
        - [] Google Chrome
        - [] Firefox
        - [] Opera
        - [] Safari

    - iOS (iPhone)
        - [] Safari Mobile
        - [] Google Chrome
        - [] Opera

    - iOS (iPad)
        - [] Safari Mobile
        - [] Google Chrome
        - [] Opera

    - Android
        - [] Firefox
        - [] Native Browser
        - [] Google Chrome
        - [] Opera

```

-----
IA Page: https://duck.co/ia/view/weight

### DIFF
--- a/lib/DDG/Goodie/Weight.pm
+++ b/lib/DDG/Goodie/Weight.pm
@@ -30,7 +30,14 @@ handle remainder => sub {
 
     # Return only if $_ is a number
     if (looks_like_number($_)){
-        return "Weight of a ".$_."kg mass on Earth is ".$_*9.80665."N. Note: Taking value of acceleration due to gravity on Earth as 9.80665m/s^2.";
+    
+        return "Weight of a ".$_."kg mass on Earth is ".$_*9.80665."N.",
+            structured_answer => {
+                input     => [],
+                operation => "Taking value of acceleration due to gravity on Earth as 9.80665m/s^2.",
+                result    => "Weight of a ".$_."kg mass on Earth is ".$_*9.80665."N.",
+            };
+            
     }
 
 
--- a/lib/DDG/Goodie/Weight.pm
+++ b/lib/DDG/Goodie/Weight.pm
@@ -1,8 +1,7 @@
 package DDG::Goodie::Weight;
-# ABSTRACT: Write an abstract here
-# Start at https://duck.co/duckduckhack/goodie_overview if you are new
-# to instant answer development
+# ABSTRACT: Calculate the weight of provided mass (in kg).
 
+use strict;
 use DDG::Goodie;
 use Scalar::Util qw(looks_like_number);
 
@@ -31,11 +30,19 @@ handle remainder => sub {
     # Return only if $_ is a number
     if (looks_like_number($_)){
     
-        return "Weight of a ".$_."kg mass on Earth is ".$_*9.80665."N.",
+        my $mass = $_;
+    
+        # Value of acceleration due to gravity on Earth in m/s^2.
+        my $g = 9.80665;
+        
+        # Weight = Mass (in kg) * Acceleration due to gravity (in m/s^2)
+        my $weight = $mass*$g;
+    
+        return "Weight of a ".$mass."kg mass on Earth is ".$weight."N.",
             structured_answer => {
                 input     => [],
-                operation => "Taking value of acceleration due to gravity on Earth as 9.80665m/s^2.",
-                result    => "Weight of a ".$_."kg mass on Earth is ".$_*9.80665."N.",
+                operation => "Taking value of acceleration due to gravity on Earth as ".$g."m/s^2.",
+                result    => "Weight of a ".$mass."kg mass on Earth is ".$weight."N.",
             };
             
     }
--- a/lib/DDG/Goodie/Weight.pm
+++ b/lib/DDG/Goodie/Weight.pm
@@ -12,14 +12,14 @@ zci is_cached   => 1;
 # Metadata.  See https://duck.co/duckduckhack/metadata for help in filling out this section.
 name "Weight";
 description "Calculate the weight of provided mass (in kg).";
-primary_example_queries "Weight 5", "Weight 5.12";
+primary_example_queries "Weight 5", "Weight 5.12", "5.1 weight";
 category "physical_properties";
 topics "math", "science";
 code_url "https://github.com/duckduckgo/zeroclickinfo-goodies/blob/master/lib/DDG/Goodie/Weight.pm";
 attribution github => ["https://github.com/wongalvis", "wongalvis"];
 
 # Triggers
-triggers start => "weight";
+triggers startend => "weight";
 
 # Handle statement
 handle remainder => sub {
--- a/lib/DDG/Goodie/Weight.pm
+++ b/lib/DDG/Goodie/Weight.pm
@@ -1,0 +1,39 @@
+package DDG::Goodie::Weight;
+# ABSTRACT: Write an abstract here
+# Start at https://duck.co/duckduckhack/goodie_overview if you are new
+# to instant answer development
+
+use DDG::Goodie;
+use Scalar::Util qw(looks_like_number);
+
+zci answer_type => "weight";
+zci is_cached   => 1;
+
+# Metadata.  See https://duck.co/duckduckhack/metadata for help in filling out this section.
+name "Weight";
+description "Calculate the weight of provided mass (in kg).";
+primary_example_queries "Weight 5", "Weight 5.12";
+category "physical_properties";
+topics "math", "science";
+code_url "https://github.com/duckduckgo/zeroclickinfo-goodies/blob/master/lib/DDG/Goodie/Weight.pm";
+attribution github => ["https://github.com/wongalvis", "wongalvis"];
+
+# Triggers
+triggers start => "weight";
+
+# Handle statement
+handle remainder => sub {
+    
+    return unless $_;
+    
+    return unless looks_like_number ($_);
+
+    # Return only if $_ is a number
+    if (looks_like_number($_)){
+        return "Weight of a ".$_."kg mass on Earth is ".$_*9.80665."N. Note: Taking value of acceleration due to gravity on Earth as 9.80665m/s^2.";
+    }
+
+
+};
+
+1;
--- a/lib/DDG/Goodie/Weight.pm
+++ b/lib/DDG/Goodie/Weight.pm
@@ -26,21 +26,28 @@ handle remainder => sub {
     return unless $_;
 
     return unless m/^(?<mass>\d+(?:\.\d+)?)\s*(?<unit>\w+)?$/;
-    my $mass = $+{mass};
+    my $mass_input = $+{mass};
     my $unit = $+{unit} // my $default_unit;
     
     my %units = (
         "kg" => 1,
+        "kgs" => 1,
         "g"  => 0.001,
+        "gs"  => 0.001,
         "mg" => 0.00001,
+        "mgs" => 0.00001,
         "t" => 1000,
+        "ts" => 1000,
         "lb" => 0.453,
+        "lbs" => 0.453,
         "oz" => 0.0283,
+        "ozs" => 0.0283,
     );
     
+    my $mass = $mass_input;
     if ($unit){
         $mass *= $units{$unit} if exists $units{$unit};
-        }
+    }
     
     # Value of acceleration due to gravity on Earth in m/s^2.
     my $g = 9.80665;
@@ -48,11 +55,20 @@ handle remainder => sub {
     # Weight = Mass (in kg) * Acceleration due to gravity (in m/s^2)
     my $weight = $mass*$g;
     
-    return "Weight of a ".$mass."kg mass on Earth is ".$weight."N.",
+    # Text to be shown to indicate conversion done
+    my $conversiontext = "(".$mass." kg) ";
+    if (!$unit) {
+        $unit = "kg";
+    }
+    if ( $unit eq "kg" || $unit eq "kgs" ) {
+        $conversiontext = "";
+    }
+    
+    return "Weight of a ".$mass_input.$unit." mass on Earth is ".$weight."N.",
             structured_answer => {
                 input     => [],
                 operation => "Taking value of acceleration due to gravity on Earth as ".$g."m/s^2.",
-                result    => "Weight of a ".$mass."kg mass on Earth is ".$weight."N.",
+                result    => "Weight of a ".$mass_input.$unit." ".$conversiontext."mass on Earth is ".$weight."N.",
             };
 
 
--- a/lib/DDG/Goodie/Weight.pm
+++ b/lib/DDG/Goodie/Weight.pm
@@ -25,7 +25,7 @@ handle remainder => sub {
 
     return unless $_;
 
-    return unless m/^(?<mass>\d+\.*\d*)\s*(?<unit>\w+)?$/;
+    return unless m/^(?<mass>\d+(?:\.\d+)?)\s*(?<unit>\w+)?$/;
     my $mass = $+{mass};
     my $unit = $+{unit} // my $default_unit;
     
--- a/lib/DDG/Goodie/Weight.pm
+++ b/lib/DDG/Goodie/Weight.pm
@@ -4,6 +4,7 @@ package DDG::Goodie::Weight;
 use strict;
 use DDG::Goodie;
 use Scalar::Util qw(looks_like_number);
+use Text::Trim;
 
 zci answer_type => "weight";
 zci is_cached   => 1;
@@ -18,7 +19,7 @@ code_url "https://github.com/duckduckgo/zeroclickinfo-goodies/blob/master/lib/DD
 attribution github => ["https://github.com/wongalvis", "wongalvis"];
 
 # Triggers
-triggers startend => "weight";
+triggers startend => "weight", "earth weight", "weight on earth";
 
 my %units = (
     "kg" => 1,
@@ -42,6 +43,8 @@ use constant g => 9.80665;
 handle remainder => sub {
 
     return unless $_;
+    s/\bof\b//;
+    trim $_;
 
     return unless m/^(?<mass>\d+(?:\.\d+)?)\s*(?<unit>\w+)?$/;
     my $mass_input = $+{mass};
--- a/lib/DDG/Goodie/Weight.pm
+++ b/lib/DDG/Goodie/Weight.pm
@@ -11,7 +11,7 @@ zci is_cached   => 1;
 # Metadata.  See https://duck.co/duckduckhack/metadata for help in filling out this section.
 name "Weight";
 description "Calculate the weight of provided mass (in kg).";
-primary_example_queries "Weight 5", "Weight 5.12", "5.1 weight";
+primary_example_queries "Weight 5", "Weight 5.12g", "5.1 kg weight";
 category "physical_properties";
 topics "math", "science";
 code_url "https://github.com/duckduckgo/zeroclickinfo-goodies/blob/master/lib/DDG/Goodie/Weight.pm";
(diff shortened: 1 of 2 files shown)


wongalvis:
There are improvements to be made such as supporting different units of mass (e.g. gram).

moollaza:
@wongalvis thanks a lot for contributing, we really appreciate it! We'll be sure to review this shortly and give you some feedback!

lib/DDG/Goodie/Weight.pm
```diff
@@ -0,0 +1,46 @@
...
+
+# Triggers
+triggers startend => "weight";
+
+# Handle statement
+handle remainder => sub {
+    
+    return unless $_;
+    
+    return unless looks_like_number ($_);
```
mintsoft: Could we extend this to handle queries like:
`weight 5kg or weight 2lbs` ? If so it might be easier with a regex rather than `looks_like_number` checking. Maybe using the numbers role? (https://duck.co/duckduckhack/goodie_helpers#numbers)

t/Weight.t
```diff
@@ -0,0 +1,25 @@
...
+use warnings;
+use Test::More;
+use DDG::Test::Goodie;
+
+zci answer_type => "weight";
+zci is_cached   => 1;
+
+ddg_goodie_test(
+    [qw( DDG::Goodie::Weight )],
+    # At a minimum, be sure to include tests for all:
```
mintsoft: Can you bomb out the instructional comments?

lib/DDG/Goodie/Weight.pm
```diff
@@ -0,0 +1,46 @@
+package DDG::Goodie::Weight;
+# ABSTRACT: Write an abstract here
+# Start at https://duck.co/duckduckhack/goodie_overview if you are new
+# to instant answer development
+
+use DDG::Goodie;
+use Scalar::Util qw(looks_like_number);
```
mintsoft: Can you add `use strict;` ?

lib/DDG/Goodie/Weight.pm
```diff
@@ -0,0 +1,46 @@
...
+handle remainder => sub {
+    
+    return unless $_;
+    
+    return unless looks_like_number ($_);
+
+    # Return only if $_ is a number
+    if (looks_like_number($_)){
+    
+        return "Weight of a ".$_."kg mass on Earth is ".$_*9.80665."N.",
```
mintsoft: I'd probably pull `$_*9.80665` into a variable (`$weight` maybe) and define `g` (~9.8) as a constant?

javathunderman:
@wongalvis Glad to see this PR!

wongalvis:
`The command "dzil authordeps | grep -ve '^\W' | xargs -n 5 -P 10 cpanm --quiet --notest --mirror http://www.cpan.org/ --mirror http://duckpan.org" failed and exited with 123 during .`
Why is this happening?

MrChrisW:
@wongalvis Don't worry it's nothing to do with your code. It's an internal issue related to a dependence which will be resolved soon :+1:

t/Weight.t
wongalvis: https://github.com/wongalvis/zeroclickinfo-goodies/commit/3f82069577c50c5bc102d7e2134668f3be6ff86c

lib/DDG/Goodie/Weight.pm
wongalvis: https://github.com/wongalvis/zeroclickinfo-goodies/commit/3f82069577c50c5bc102d7e2134668f3be6ff86c

lib/DDG/Goodie/Weight.pm
wongalvis: https://github.com/wongalvis/zeroclickinfo-goodies/commit/3f82069577c50c5bc102d7e2134668f3be6ff86c

moollaza:
@wongalvis apologies for the confusion -- this *should* be resolved now. Please let me know if you're still having any trouble! You may need to run `duckpan upgrade` to fix things...

moollaza:
Also, is this one ready for me to look at?

wongalvis:
@moollaza What was the problem you were referring to?
And yes this is ready.

lib/DDG/Goodie/Weight.pm
```diff
@@ -0,0 +1,61 @@
...
+# Handle statement
+handle remainder => sub {
+
+    return unless $_;
+
+    return unless m/^(?<mass>\d+(?:\.\d+)?)\s*(?<unit>\w+)?$/;
+    my $mass = $+{mass};
+    my $unit = $+{unit} // my $default_unit;
+    
+    my %units = (
```
moollaza: looks like you need to also support plural forms here. my first query was `weght 5kgs` and it was misinterpretted

lib/DDG/Goodie/Weight.pm
```diff
@@ -0,0 +1,61 @@
...
+        "g"  => 0.001,
+        "mg" => 0.00001,
+        "t" => 1000,
+        "lb" => 0.453,
+        "oz" => 0.0283,
+    );
+    
+    if ($unit){
+        $mass *= $units{$unit} if exists $units{$unit};
+        }
```
moollaza: nit: indent is off here

lib/DDG/Goodie/Weight.pm
```diff
@@ -0,0 +1,61 @@
...
+    my $g = 9.80665;
+    
+    # Weight = Mass (in kg) * Acceleration due to gravity (in m/s^2)
+    my $weight = $mass*$g;
+    
+    return "Weight of a ".$mass."kg mass on Earth is ".$weight."N.",
+            structured_answer => {
+                input     => [],
+                operation => "Taking value of acceleration due to gravity on Earth as ".$g."m/s^2.",
+                result    => "Weight of a ".$mass."kg mass on Earth is ".$weight."N.",
```
moollaza: I found it a little confusing that we convert the input weight to Kgs in the answer, I was expecting to see my input weight/units.

I think we should put the original amount and unit and specify the converted amount in Kgs in brackets:

>Weight of a 5lb (2.265kg) mass on Earth is 22.21206225N.

moollaza:
@wongalvis almost there! Needs some further testing and handling of units though

lib/DDG/Goodie/Weight.pm
wongalvis: Sorry but I don't get where went wrong.

moollaza: the closing brace, `}` looks like it's indented too far:

```perl
if ($unit){
    $mass *= $units{$unit} if exists $units{$unit};
}
```

wongalvis: I noticed it and it's fixed in the latest commit.

moollaza:
@wongalvis whoops! I didn't notice you pushed a new commit. Please let us know when you've updated the code as GitHub unfortunately doesn't create notifications for that.

lib/DDG/Goodie/Weight.pm
```diff
@@ -0,0 +1,77 @@
...
+# Handle statement
+handle remainder => sub {
+
+    return unless $_;
+
+    return unless m/^(?<mass>\d+(?:\.\d+)?)\s*(?<unit>\w+)?$/;
+    my $mass_input = $+{mass};
+    my $unit = $+{unit} // my $default_unit;
+    
+    my %units = (
```
moollaza: Let's move this outside the `handle` function. Static functions and variables can be declared before the `handle` so that they're only created once, rather than each time the IA runs.

lib/DDG/Goodie/Weight.pm
```diff
@@ -0,0 +1,77 @@
...
+        "ozs" => 0.0283,
+    );
+    
+    my $mass = $mass_input;
+    if ($unit){
+        $mass *= $units{$unit} if exists $units{$unit};
+    }
+    
+    # Value of acceleration due to gravity on Earth in m/s^2.
+    my $g = 9.80665;
```
moollaza: We can move this outside the `handle` function as well. 

If you really want you could define `g` or `G` as a `constant`: http://perldoc.perl.org/constant.html

t/Weight.t
```diff
@@ -0,0 +1,22 @@
...
+
+zci answer_type => "weight";
+zci is_cached   => 1;
+
+ddg_goodie_test(
+    [qw( DDG::Goodie::Weight )],
+    # - primary_example_queries
+    'weight 5' => test_zci("Weight of a 5kg mass on Earth is 49.03325N.", structured_answer => { input =>[], operation => "Taking value of acceleration due to gravity on Earth as 9.80665m/s^2.", result =>"Weight of a 5kg mass on Earth is 49.03325N."}),
+    'weight 5.12g' => test_zci("Weight of a 5.12g mass on Earth is 0.050210048N.", structured_answer => { input =>[], operation => "Taking value of acceleration due to gravity on Earth as 9.80665m/s^2.", result =>"Weight of a 5.12g (0.00512 kg) mass on Earth is 0.050210048N."}),
+    '5.1 kg weight' => test_zci("Weight of a 5.1kg mass on Earth is 50.013915N.", structured_answer => { input =>[], operation => "Taking value of acceleration due to gravity on Earth as 9.80665m/s^2.", result =>"Weight of a 5.1kg mass on Earth is 50.013915N."}),
```
moollaza: Let's add a test or two that use some of the other possible units (like `g` or `oz`)

moollaza:
@wongalvis this is *nearly* done! Just a few more changes that should really improve this IA :+1:

lib/DDG/Goodie/Weight.pm
```diff
@@ -0,0 +1,77 @@
...
+        "mgs" => 0.00001,
+        "t" => 1000,
+        "ts" => 1000,
+        "lb" => 0.453,
+        "lbs" => 0.453,
+        "oz" => 0.0283,
+        "ozs" => 0.0283,
+    );
+    
+    my $mass = $mass_input;
```
moollaza: Looks like this can be removed and line 29, can simple be `my $mass = $+{mass};`

wongalvis: $mass_input is to keep a copy of the user inputted mass value which is returned in the answer.

wongalvis:
Updated @moollaza

lib/DDG/Goodie/Weight.pm
```diff
@@ -0,0 +1,77 @@
...
+name "Weight";
+description "Calculate the weight of provided mass (in kg).";
+primary_example_queries "Weight 5", "Weight 5.12g", "5.1 kg weight";
+category "physical_properties";
+topics "math", "science";
+code_url "https://github.com/duckduckgo/zeroclickinfo-goodies/blob/master/lib/DDG/Goodie/Weight.pm";
+attribution github => ["https://github.com/wongalvis", "wongalvis"];
+
+# Triggers
+triggers startend => "weight";
```
moollaza: let's add some triggers:

```perl
triggers startend => "weight", "earth weight", "weight on earth";
```

Then, let's also strip `s/of//;` to allow queries like "weight of 5kg", etc

lib/DDG/Goodie/Weight.pm
```diff
@@ -0,0 +1,77 @@
...
+use constant g => 9.80665;
+
+# Handle statement
+handle remainder => sub {
+
+    return unless $_;
+
+    return unless m/^(?<mass>\d+(?:\.\d+)?)\s*(?<unit>\w+)?$/;
+    my $mass_input = $+{mass};
+    my $unit = $+{unit} // my $default_unit;
```
moollaza: I'm not sure what this line's trying to achieve. 

You're setting the value of `$unit` to `$+{unit}`, but when that is undefined, you're setting it to a reference to a new variable, `my $default_unit` -- I think you should be defining `my $default_unit` first? and then it should look like:

```perl
my $defualt_unti = "kg";
my $unit = $+{unit} // $defaul_unit;
```

Let's make sure we have a test that captures this behaviour to ensure things are working as expected!

lib/DDG/Goodie/Weight.pm
```diff
@@ -0,0 +1,77 @@
...
+    my $conversiontext = "(".$mass." kg) ";
+    if (!$unit) {
+        $unit = "kg";
+    }
+    if ( $unit eq "kg" || $unit eq "kgs" ) {
+        $conversiontext = "";
+    }
+    
+    return "Weight of a ".$mass_input.$unit." mass on Earth is ".$weight."N.",
+            structured_answer => {
```
moollaza: Looks like this is indented a little too far?

lib/DDG/Goodie/Weight.pm
```diff
@@ -0,0 +1,77 @@
...
+handle remainder => sub {
+
+    return unless $_;
+
+    return unless m/^(?<mass>\d+(?:\.\d+)?)\s*(?<unit>\w+)?$/;
+    my $mass_input = $+{mass};
+    my $unit = $+{unit} // my $default_unit;
+    
+    my $mass = $mass_input;
+    if ($unit){
```
moollaza: would move the `exists` clause below into your `if` here:

```perl
if (foo && exists bar{foo}) {
    ....
}
```

lib/DDG/Goodie/Weight.pm
```diff
@@ -0,0 +1,77 @@
...
+        $mass *= $units{$unit} if exists $units{$unit};
+    }
+    
+    # Weight = Mass (in kg) * Acceleration due to gravity (in m/s^2)
+    my $weight = $mass*g;
+    
+    # Text to be shown to indicate conversion done
+    my $conversiontext = "(".$mass." kg) ";
```
moollaza: You can use string interpolation with perl variables :)

```perl
my $string = "($mass kg) ";
```

lib/DDG/Goodie/Weight.pm
```diff
@@ -0,0 +1,77 @@
...
+    if (!$unit) {
+        $unit = "kg";
+    }
+    if ( $unit eq "kg" || $unit eq "kgs" ) {
+        $conversiontext = "";
+    }
+
```
moollaza: I would suggest you build a new string from `$mass_input.$unit` and then you can simplify your strings below:

```perl
my $massUnit = $mass.$unit;
return "Weight of a $massUnit mass on Earth is $weight"."N."

File: lib/DDG/Goodie/Weight.pm (hunk 12 above)
Comment (wongalvis) [inline lib/DDG/Goodie/Weight.pm]: Don't get ```Then, let's also strip s/of//; (in your handle function)```

**Comment by wongalvis** [comment]
Updated @moollaza

File: lib/DDG/Goodie/Weight.pm (hunk 12 above)
Comment (moollaza) [inline lib/DDG/Goodie/Weight.pm]: >Having triggers startend => "weight", "earth weight", "weight on earth"; will cause errors with the unit interpretation, at this stage it seems incapable of having these triggers.

What errors? Why does it cause them? As far as I can tell, it should have no effect...

Comment (moollaza) [inline lib/DDG/Goodie/Weight.pm]: I just tested and it works:

```perl
handle remainder => sub {
  return unless $_;
  s/\bof\b//;
  trim $_   # add 'use Text::Trim;` above
  ...
}
```

wongalvis: Updated in the latest commit https://github.com/wongalvis/zeroclickinfo-goodies/commit/e71af1ad8c2526b97ecf65331fbd4a3cc3901d5a. It somehow worked haha :) @moollaza

wongalvis:
ping @moollaza

mintsoft:
@moollaza This LGTM

moollaza:
@wongalvis @mintsoft this is looking good!

@abeyang this one is ready for a design check!

abeyang:
@wongalvis It's not immediately clear to me, based on the query, what this is doing. @moollaza I think we need to use a more specific trigger? For example, is this converting kg to newtons on earth? If so, then I think the trigger should be a bit clearer, like `weight to N 5.12` or something.

And if indeed this is the conversion, I think the result should be:

> **50.210048N**
> Weight of a 5.12kg mass on Earth

I would assume that if it's on earth, it would take the earth's gravity into account, thus we don't need to explicitly call that out...

mintsoft:
@abeyang *technically* acceleration due Earth's gravity varies depending on location; so it might be worth actually stating `assuming g= 9.8xxxxxx m/s^2`

wongalvis:
@abeyang The SI unit of mass is kg, while that for weight, which is a force, is N.
People conventionally call the "weight" of something/somebody is XXX kg, yet it is not correct in physics.

abeyang:
@mintsoft It does? Where (on earth) would it differ, and what would `g` be? And if there is a difference, wouldn't that be negligible?

@wongalvis are you responding to my comment re: the query? I understand what this IA is doing; it's the query itself that's unclear. For example, *this* query *is* clear: [5 lb to oz](https://duckduckgo.com/?q=5+lb+to+oz&ia=answer)

mintsoft:
@abeyang it does indeed. To 1 d.p it'll always round out to 9.8 m/s^2 but it's in the range of like 9.76 -> 9.84 IIRC; which isn't exactly nothing!

wongalvis:
@abeyang Any suggestions how to improve the query?

abeyang:
@mintsoft wow, [you're right](https://en.wikipedia.org/wiki/Gravity_of_Earth)! But that does make sense...

@wongalvis Actually, after reading more on this, I think it's ok to have your query. However, though we could default it to kg, would you be able to accept lbs as well? So "weight 10lbs", for example. And I think the answer, given @mintsoft 's suggestion, should be more like this:

> **50.210048N**
> 5.12kg - weight on Earth (g = 9.80665 m/s^2 )

wongalvis:
Indeed it does accept "lbs" and other units https://github.com/duckduckgo/zeroclickinfo-goodies/pull/1112/files#diff-cd5c2be32dd8c17ce6dd374c3542d11eR33
Just updated the Request Template

moollaza:
@wongalvis just noticed a subtle bug: While trying to test I made a typo and entered "weight of 5 k", it gives an answer with "5k" as the input number/unit. It looks like if you enter anything after the number though, we assume it’s a unit which is incorrect:

weight_of_5_foo_at_duckduckgo

If we detect letters after the number, we need to verify it's a legitimate "unit" and if not, bail.

lib/DDG/Goodie/Weight.pm
```diff
@@ -0,0 +1,82 @@
...
+use constant g => 9.80665;
+
+# Handle statement
+handle remainder => sub {
+
+    return unless $_;
+    s/\bof\b//;
+    trim $_;
+
+    return unless m/^(?<mass>\d+(?:\.\d+)?)\s*(?<unit>\w+)?$/;
```
moollaza: We should ensure the unit is just alpha characters: `(?<unit>[a-z]+)?`

lib/DDG/Goodie/Weight.pm
```diff
@@ -0,0 +1,82 @@
...
+    "lbs" => 0.453,
+    "oz" => 0.0283,
+    "ozs" => 0.0283,
+);
+
+# Value of acceleration due to gravity on Earth in m/s^2.
+use constant g => 9.80665;
+
+# Handle statement
+handle remainder => sub {
```
moollaza: We'll want to `handle remainder_lc` to normalize the unit input

lib/DDG/Goodie/Weight.pm
mintsoft: If support for decimals is required, the numbers role will provide a regex

wongalvis:
Updated @moollaza

moollaza:
@wongalvis this LGTM :+1: Excellent work, sorry that it's taken so long, but thanks a lot for sticking around!

Thanks again for contributing (we really appreciate it :smiley:). We invite you to stick around and comment/help on other PR's & Issues or even submit another Instant Answer!

moollaza:
@wongalvis congrats! It's live: https://duckduckgo.com/?q=weight+5.2oz&ia=answer

PR #1123, opened 2015-05-02
Improve DuckDuckGo's hiring links

bpollack:
I shot my mouth off on Twitter, so: here's a better link. That said, I think this is actually at least partially underlying lousy visual styling.


zekiel:
lol thanks @bpollack  --merging

PR #1131, opened 2015-05-12
Update CheatSheets for v1

moollaza:
Getting CheatSheets ready for V1 release:

- [x] change `meta` to `metadata` to prevent any collisions
- [x] update UI to meet design comp
- [x] replace `[]` with `<code></code>` in JSON Keys

/cc @russellholt @zachthompson


lib/DDG/Goodie/CheatSheets.pm
```diff
@@ -41,7 +41,7 @@ handle remainder => sub {
         id => 'cheat_sheets',
         name => 'Cheat Sheet',
         data => $data,
-        meta => \%{$data->{meta}},
+        meta => \%{$data->{metadata}},
```
zachthompson: If I recall correctly, the whole reason we added the json metadata to meta at this level was because we didn't know where to put it.  I don't think you need to add it here any longer.

moollaza: Right! Fixes in newest commit :+1:

moollaza:
Ready for review!

zachthompson:
Should we replace the "/" as a way to specify alternates with the new "[] or []"?  Look at tabclose for example which still has the / as part of the command:

vim_cheat_sheet

@abeyang Does the layout look like what was discussed last week?

[vim](https://zach.duckduckgo.com/?q=vim+cheat+sheet&ia=cheatsheet) [tmux](https://zach.duckduckgo.com/?q=tmux+cheat+sheet&ia=cheatsheet) [regex](https://zach.duckduckgo.com/?q=regex+cheat+sheet&ia=cheatsheet)

abeyang:
@zachthompson I think it's almost there! For visuals, let's modify the css a bit:

``` css
.zci--cheat_sheets .cheatsheet__data code {
  padding: 4px 5px 2px;
  font-size: 0.833em;
}
```

And yes, I agree that we should show alternates in new code blocks, like:

> [:tabclose] / [:tabc]

@moollaza are these other tasks still in progress:

screen shot 2015-05-15 at 10 23 16 am

russellholt:
But we're just treating it as text with the introduction of the [ ] syntax. so the idea of alternates don't even factor into this, as something we explicitly "support", except as a standard way of writing it... right?

abeyang:
@russellholt sorry, not sure I follow you... all I'm saying is that `:tabclose / :tabc` should **not** be in the same `<code>` block. Maybe I should've written it like:

`:tabclose` / `:tabc`

Does that make sense? Not sure if we were on different wavelengths before..

russellholt:
@abeyang yeah I understand

zachthompson:
btw, same with the ', ' variation, e.g. 'gt, :tabnext, :tab".   I introduced these variations as examples of how we could specify them in the json when Russell and I were first discussing it as a POC.  I don't see them in the originals.   We may want to verify, compare, and/or sync with the originals.

moollaza:
Alright, it's live here: https://moollaza.duckduckgo.com/?q=regex+cheatsheet&ia=cheatsheet

@abeyang reducing the `code` font-size to `0.8em` might be necessary. When the keys get longer we run into problems with the text folding, not sure about the most graceful way to handle that...

mintsoft:
@moollaza is it anticipated that the targeted help doesn't work?

https://duckduckgo.com/?q=regex+%5Ba-x%5D&ia=answer 
https://moollaza.duckduckgo.com/?q=regex+%5Ba-x%5D

russellholt:
@mintsoft we're thinking about that, will most likely be for v2 of this though.

russellholt:
LGTM. We might move the string escape logic to the backend (the cheatsheet module) for v2.

Status: awaiting final design review

abeyang:
Hi, gonna give some feedback. First, just on **tmux** specifically, can we get rid of the commas? I think it looks cleaner (and still understandable) without them, similar to what I have for Emacs here:

screen shot 2015-05-27 at 9 52 07 am

abeyang:
Now, with regards to the Cheatsheet Template itself:

First, can we have variable column widths? Gutter between the two should be 14px (which is **0.972em**, I believe). It looks like the column widths are all the same, leading to some of the keys running into the description (in pink below):

pasted_image_5_27_15__9_58_am

Second, as noted in the blue above, can we have the cheatsheet fill up the entire width of the gray area? 

Third, need wider gutters between columns. Here's the ideal (yours is about half that right now):

column_sizes

Fourth, there needs to be a bigger gap (should be about 30px total) between the headings and the previous grouping. Tried tweaking your css, but couldn't quite find a way.. 

Fifth, currently in the css, there's a `padding-bottom: 0` for `.zci--cheat_sheets .cheatsheet__data code`... please delete that line, so that the 2px padding-bottom can take effect.

abeyang:
Finally, two more things:

Can we make the footer more consistent with IAs? Our "show more" text seems a bit below and to the right of the icon...

screen shot 2015-05-27 at 10 16 53 am

And not sure how we usually do that horizontal line, but I'm assuming we don't typically use an `<hr>`? (If I were to guess, we just use the `border-top` CSS property.) Also, the color should be `rgba(0,0,0,0.15)`.

moollaza:
@abeyang alright, everything should be addressed. Should be live on my machine momentarily...

moollaza:
Alright, got the green light from the design team!

russellholt:
:+1: incredible this is done. On to v2! :)

zachthompson:
Nice job :+1:

abeyang:
:+1:

PR #1134, opened 2015-05-15
PeriodicTable.pm: Add `url` to the data.

jagtalon:
The title would link nowhere if we don't add this.

screen shot 2015-05-15 at 7 28 51 am


skywickenden:
Should the title be linking to the Wikipedia reference?

moollaza:
LGTM :+1:

PR #1139, opened 2015-05-19
International Protection Marking Codes [New]

mattr555:
I realized I had never made a `structured_answer` before!

Decodes IP Code markings.

These are a standard and simple for answers.

https://en.wikipedia.org/wiki/IP_Code

It's correct

`IP68`, `ip 42`

People working with machinery/parts that need to be protected from particle ingress (like roboticists)

[My own.](https://duck.co/ideas/idea/5227/ip-international-protection-marking-codes)

screenshot - 05192015 - 07 13 55 pm

## Checklist

```

    - Debian (Crunchbang)
        - [x] Chrome
```

### DIFF
--- a/lib/DDG/Goodie/IPCodes.pm
+++ b/lib/DDG/Goodie/IPCodes.pm
@@ -1,0 +1,61 @@
+package DDG::Goodie::IPCodes;
+# ABSTRACT: International Protection Marking Codes
+
+use DDG::Goodie;
+
+zci answer_type => "ipcodes";
+zci is_cached   => 1;
+
+name "IPCodes";
+description "International Protection Marking Codes";
+primary_example_queries "IP68", "IP4X";
+category "reference";
+topics "special_interest";
+code_url "https://github.com/duckduckgo/zeroclickinfo-goodies/blob/master/lib/DDG/Goodie/IPCodes.pm";
+attribution github => ["mattr555", "Matt Ramina"],
+            twitter => "mattr555";
+
+my $regex = qr/^ip\s?([0-6x])([0-8x]|6k|9k)$/;
+triggers query_lc => $regex;
+
+my %particles = (
+	"0" => "None",
+	"1" => "Back of hand",
+	"2" => "Fingers",
+	"3" => "Tools",
+	"4" => "Wire",
+	"5" => "Dust protected",
+	"6" => "Dust tight",
+	"X" => "Unspecified"
+);
+
+my %water = (
+	"0" => "None",
+	"1" => "Dripping water",
+	"2" => "Dripping water, object tilted",
+	"3" => "Spraying water",
+	"4" => "Splashing water",
+	"5" => "Water jets",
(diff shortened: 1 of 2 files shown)


MrChrisW:
@mattr555 Looking good, hope you don't mind I just did a small commit to simplify the var assignment and return :+1:

lib/DDG/Goodie/IPCodes.pm
```diff
@@ -0,0 +1,60 @@
+package DDG::Goodie::IPCodes;
+# ABSTRACT: International Protection Marking Codes
+
+use DDG::Goodie;
+
```
mintsoft: Can you add `use strict;` to keep Github in Perl5 mode?

lib/DDG/Goodie/IPCodes.pm
```diff
@@ -0,0 +1,60 @@
...
+description "International Protection Marking Codes";
+primary_example_queries "IP68", "IP4X";
+category "reference";
+topics "special_interest";
+code_url "https://github.com/duckduckgo/zeroclickinfo-goodies/blob/master/lib/DDG/Goodie/IPCodes.pm";
+attribution github => ["mattr555", "Matt Ramina"],
+            twitter => "mattr555";
+
+my $regex = qr/^ip\s?([0-6x])([0-8x]|6k|9k)$/;
+triggers query_lc => $regex;
```
mintsoft: It's a shame it needs a regex trigger, but short of listing all the IPs out as word triggers I can't think of an alternative

mintsoft:
Nice LGTM aside from the `use strict` @abeyang any design feedback, it's using the standard structured answer

mattr555:
@MrChrisW :+1: 

@mintsoft agreed on both accounts.

lib/DDG/Goodie/IPCodes.pm
moollaza: @mattr555 Any chance we could make this into a Fathead then? Seems like its a perfect use-case?

moollaza: Alternatively, we should be able to pre-generate all the combinations right? We could dump it into a JSON/YAML file and then load the keys as triggers -- very similar to how we handle the Transit IAs.

If we're going to generate all the combinations though, then a Fathead would be best as the triggering and answering would be handled internally. You just need to provide the data file :)

In this case I think it would be small enough to leave in the repo if needed.

mintsoft:
@mattr555 @moollaza are we going fathead for this? (btw, can fatheads be tested in `duckpan` yet?)

moollaza:
@mintsoft I believe a Fathead would be ideal here!

>(btw, can fatheads be tested in duckpan yet?)

Sadly, not yet, but it's definitely something we're working towards!

mattr555:
Good idea guys, totally forgot about fatheads. Continued over on duckduckgo/zeroclickinfo-fathead#131

PR #830, opened 2014-12-11
Add support for computing binomial coefficients

MichaelBurge:
# Goodie Pull Request Template

Using this template will help us better understand your Instant Answer and assist you when necessary.  Simply copy and paste the markdown below into the description of your GitHub pull request and complete as appropriate.

**What does your Instant Answer do?**
It computes the Binomial Coefficients, or elements of Pascal's Triangle.

**What problem does your Instant Answer solve (Why is it better than organic links)?**
The instant answer only triggers on queries that appear to want a specific calculated result, and in these cases the user gets a result sooner.

**What is the data source for your Instant Answer? (Provide a link if possible)**
I own a copy of Wolfram Mathematica 9.0, which has the built-in Binomial calculation. The test table was computed using the following code:
Table[ Binomial[x, y], {x, 1, 10}, {y, 1, 10} ]

and can be reproduced using Wolfram Alpha:
http://www.wolframalpha.com/input/?i=Table%5B+Binomial%5Bx%2C+y%5D%2C+{x%2C+0%2C+10}%2C+{y%2C+0%2C+10}+%5D

**Why did you choose this data source?**
I consider Mathematica to be a trustworthy software.

**Are there any other alternative (better) data sources?**
Wikipedia has a table for the first several rows here:
https://en.wikipedia.org/wiki/Binomial_coefficient

And Google can also do this:
https://www.google.com/#q=7+choose+2

**What are some example queries that trigger this Instant Answer?**
binomial(7, 1)
binomial(7,1)
7 choose 1

**Which communities will this Instant Answer be especially useful for? (gamers, book lovers, etc)**
It makes DDG slightly more useful as a calculator, which benefits everyone

**Is this Instant Answer connected to a DuckDuckHack [Instant Answer idea](https://duck.co/ideas)?**
I didn't find one when I searched.

**Which existing Instant Answers will this one supercede/overlap with?**
I didn't see any that calculated this exact function. The Calculator.pm could be generalized to have a more robust formula parser, which would make this module along with several others(Fibonacci.pm, BinaryLogic.pm) redundant.

**Are you having any problems? Do you need our help with anything?**

**Where did you hear about DuckDuckHack? (For first time contributors)**
I searched for this Instant Answer, and when it didn't come up remembered that DuckDuckHack existed. I first heard about it several months ago, although I don't remember what lead me to it.

**What does the Instant Answer look like? (Provide a screenshot for new or updated Instant Answers)**

## Checklist
Please place an 'X' where appropriate.

```
[...]

PR #1140, opened 2015-05-20
Computing binomial coefficients IA

javathunderman:
Continuation of #830 

cc @mintsoft

### DIFF
--- a/lib/DDG/Goodie/Binomial.pm
+++ b/lib/DDG/Goodie/Binomial.pm
@@ -1,0 +1,48 @@
+package DDG::Goodie::Binomial;
+# ABSTRACT: Calculates binomial coefficients
+
+use DDG::Goodie;
+use List::Util qw/ product /;
+use strict;
+
+zci answer_type => "binomial";
+zci is_cached   => 1;
+
+name "Binomial";
+description "Computes Binomial coefficients";
+primary_example_queries "binomial(7, 1)", "7 choose 1";
+category "calculations";
+topics "math";
+code_url "https://github.com/duckduckgo/zeroclickinfo-goodies/blob/master/lib/DDG/Goodie/Binomial.pm";
+attribution github => ["MichaelBurge", "Michael Burge"];
+            github => ["javathunderman", "Thomas Denizou"];
+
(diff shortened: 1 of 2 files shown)


lib/DDG/Goodie/Binomial.pm
```diff
@@ -0,0 +1,48 @@
...
+category "calculations";
+topics "math";
+code_url "https://github.com/duckduckgo/zeroclickinfo-goodies/blob/master/lib/DDG/Goodie/Binomial.pm";
+attribution github => ["MichaelBurge", "Michael Burge"];
+            github => ["javathunderman", "Thomas Denizou"];
+
+my $choose_regex = qr/(\d+) choose (\d+)/i;
+my $binomial_regex = qr/binomial\((\d+),\s*(\d+)\)/i;
+
+triggers query_raw => $choose_regex, $binomial_regex;
```
mintsoft: This can be replaced with a triggers start and a triggers any I think. with "binomal" and "choose"; it'd be much faster

lib/DDG/Goodie/Binomial.pm
```diff
@@ -0,0 +1,48 @@
...
+
+sub binomial_response {
+    my ($n, $k) = @_;
+    my $result = compute_binomial($n, $k);
+    return undef unless defined($result);
+    return "binomial($n, $k) = $result"
+}
+
+handle query_raw => sub {
+    my ($query) = @_;
```
mintsoft: Then move the regexy bit into here to extract the right bits

moollaza:
Seems like this functionality already exists in the calculator?

https://duckduckgo.com/?q=7+choose+4&t=ffab&ia=answer

javathunderman:
Oh, ok. I saw this labeled as "abandoned", so I assumed this wasn't live. I'll close this then, just unlabel that old PR.

moollaza:
@javathunderman I don't think the other PR was ever implemented but I'm just noticiing that we seem to have some of the functionality already. If we can add support for `binomial(x, y)` that would be great, however a new Math Spice is currently being developed :)

javathunderman:
Yeah, since the math Spice is in the works, I'm not going to bother with getting this PR merged.

PR #1142, opened 2015-05-20
Fix line ending. Stop warning from DDG lib

moollaza:
/cc @MrChrisW


MrChrisW:
@moollaza :+1: Sorry missed that!

PR #1146, opened 2015-05-23
Added pod documentation for unicode goodie

anotherkabab:
Added pod documentation for unicode goodie


moollaza:
@psatish thanks a lot for taking the time to document this code! I understand that POD is the typical way to document Perl, but we tend to use inline comments to explain functions. Would you mind switching the POD to comments?

psatish:
@moollaza Thanks for reviewing

Is there a reason we don't use POD? I was hoping we'd be able to pod all modules to have a searchable developer reference, which could even be hosted on a website.

moollaza:
@psatish there isn't a specifc reason we're not using it, initially all code was documented with inline comments and that method has stuck. Pulling out all the POD and hosting it as a reference would be neat, but I don't feel there's much need for that, or any big benefits.

However, if we had out own open-soure Perl/JS Utils library, I would totally be down to document those in some manner such that we could pull it out for the documentation.

We are considering open sourcing some of our Handlebars helpers for example, and I could see some use in a Perl utils lib, but right now we haven't need to share much Perl between IA's except for the functionality exported by the DuckDuckGo library, which is documented with POD, however we're not really doing much with it ;)

psatish:
@moollaza Thanks for your reply, makes sense.

I have now updated the commit to have inline comments

moollaza:
@anotherkabab so sorry, I just saw this update! Not sure how I missed the notification. 

This LGTM :+1: 

anotherkabab:
@moollaza thank you

Issue #1138, opened 2015-05-19
Merge all package tracking IA's into 1, implement new design


moollaza:
Currently, we have several Package Tracking IA's, each of which is implemented as their own Perl Package. Really, none of these IA's really need to be on their own and should probably be merged so we can continue to add more carriers to the list.

We should be able to create a single "PackageTracking" Goodie which loads a data file that specifies the name, word triggers and regex pattern for their tracking codes. Using this, the Goodie should be able to create triggers and regexes that map to carriers.

In the case that only one carrier works, we can use the new Icon template to produce a result. If multiple providers are valid, we can use the list template to show the matching providers.

Anyone interested in tackling this?
##### New designs

Icon view:
![trackpackage-logos-v1 1](https://cloud.githubusercontent.com/assets/873785/7708631/f9057d52-fe24-11e4-8574-c2f8f24ea61d.jpg)

(This is meant to demonstrate all the logos -- we'd only show one of these depending on the matched carrier)

List view: 
![trackpackage-list-v1 1x3](https://cloud.githubusercontent.com/assets/873785/7708626/ec084558-fe24-11e4-8612-3de315571d99.jpg)

abeyang:
Here are the assets:

http://cl.ly/3q1y1x3P0f2V

mintsoft:
If no-one else has picked this up in a few days I can take it on

dronov:
I would like to fix this issue. So I will made the PR as fast as I can.

mintsoft:
@dronov Excellent! Any problems, just open a PR and someone'll swing in to lend a hand :+1:

Mailkov:
@abeyang @moollaza the icons on  http://cl.ly/3q1y1x3P0f2V must contain background color because background colors of logos are different  IMHO

abeyang:
@Mailkov not sure I understand what you mean. The background color in the mockups should be generated from html/css.

Mailkov:
@abeyang  This is the result when I use icons on http://cl.ly/3q1y1x3P0f2V (in this case BRT and DHL)
![answerpackagetracking](https://cloud.githubusercontent.com/assets/10259499/11871297/e27ebb74-a4ce-11e5-9dd3-3cf949d06a72.png)
[...]

PR #1150, opened 2015-05-26
Merge all package tracking IA's into 1 (Work in progress #1138 )

dronov:
This is a start for fixing #1138 
I've made code structure. I added just Hongkong Post, Post Norden and Russian Post without any assets. I'll upload it later. 

It could be great If someone gives me any feedback looking at my code. It based on HKDK and one another module and I'm interested to make it clear and readable.

UPD: of course this is not a final code. I has to be reviewed.


share/goodie/package_tracking/responses.yml
```diff
@@ -0,0 +1,15 @@
+HongKong:
+  name: "Hongkong Post"
+  tracking_url: http://app3.hongkongpost.com/CGI/mt/genresult.jsp?tracknbr=
+  short_code: hk$
```
moollaza: nit: flip this and the previous line for consistency :)

moollaza:
@dronov thanks a lot for tackling this! We really apprecaite it :smile: 

This is definitely a step in the right direction. I think you'll want to adjust the Perl so that you collect all the tracking companies that succesfully matched the given code and then if we had 2+ matches, you'll want to adjust the template you're going to use.

Also, this should now be updated to use the Goodie templates. We *just* launched some documention outlining how this works for Goodies: https://duck.co/duckduckhack/goodie_displaying

If we have 2+ matches you'll want to set the template group to `list` otherwise it'll need to be the icon template and the icons etc will have to be set accordingly.

dronov:
@moollaza thank you for feedback! I'm glad that I had chosen the right way to tackle this :) 

This is an interesting task cause many postal services have a unique tracking codes. But currently I;m working on use cases with one service matches with user's query (e.g. I've added the Lasership tracking). When I'll meet use cases with 2+ matches tracking companies I'll change the main logic of code and make the correct layout.

moollaza:
Sounds good! Let us know when you're ready for more review :+1:

dronov:
@moollaza I didn't forget about my PR. I had not a lot of free time on last week but I'll back to PR tomorrow.

dronov:
Hello again, @moollaza! 

I need an advice. As I looked at the code again and again I realized that I need a little bit complicated solution for detecting is that the users query a real postal service or not. Many of classes have an different algorythms to check matched user's query for correct tracking number (e.g. Fedex, DHL, UPS, USPS). I think that writing one big perl package named PackageTracking wouldn't be the good solution. Based on this fact I think that the implementing of Strategy Pattern would be better: we have some classes (packages) for every Postal Service with methods to decide is that the user's query a real tracking code or not. In base strategy we could handle the common words like 'tracking, track' and codes like 'XX12356789YY'. After that we could create a concrete Strategy of some Postal Service which checks correctness of the code.

What do you think about this? If my thoughts make a sense could I make some directory like lib/DDG/Goodie/PackageTracking to implement the Strategy Pattern and hold necessary packages there?

PS. I thought about the simpler solutions, but all of them had a lot of code with difficult readable structure.

moollaza:
@dronov sorry for the delay! Let me think on this a bit and get back to you!

dronov:
@moollaza ping :) 
I still would like to get an advice...

moollaza:
@dronov very sorry for the delay!

I'm not familiar with the Strategy Pattern, but your approach sounds reasonable. I had hoped that the checks required to determine which postal service the input belongs to could be greatly simplified (ideally just a pattern match), so that we wouldn't need to have several modules.

Such is not that case ;) so we'll have to handle things a little differently. Your suggested approach means that the IA will hopefully be extensible, so new sub-modules can be added for new postal services and ideally the main controller won't have to change much, or at all.

I'm confident whatever you arrive at will be a sound solution, so please go ahead and get started. Feel free to ping me if you need more feedback!

Thanks again for contributing @dronov, and sorry about the delay!

mintsoft:
@dronov Just checking in, are you still interested in finishing this off?

PR #1154, opened 2015-05-28
Send metadata for cheat sheet json to SERP

zachthompson:
The CheatSheets IA serves multiple "child" cheat sheets via json.  The metadata associated with it should be selected dynamically from the json.

Before release, the three existing cheat sheets need to have their perl_module updated to DDG::Goodie::CheatSheets on each IA page.


moollaza:
LGTM :+1:

PR #1156, opened 2015-05-29
Remove old cheatsheets

moollaza:
Removing all the old cheat sheets now replaced by the new CheatSheets.pm

I've also updated the Tmux, Regex and Vim IA pages.


zachthompson:
Where does [SOURCE](https://duck.co/ia/view/tmux_cheat_sheet) come from and how do we change that?

moollaza:
@zachthompson I believe that's a question for @MariagraziaAlastra -- AFAIK we can't directly change it. It comes from the original PR that's attached to the IA?

PR #1159, opened 2015-05-29
Add regex cheat sheet aliases

zachthompson:
As examples of how to do aliases/alternates for cheat sheet names, added some for the regex cheat sheet.  These will handle:

   * regular expression cheat sheet
   * regular expressions cheat sheet
   * regexp cheat sheet

They should also handle irregular/multiple spacing between words.


moollaza:
LGTM :+1:

PR #1167, opened 2015-06-02
CheatSheet Parsing update

moollaza:

Thanks to @zekiel's Git CheatSheet, he brought a bunch of edge cases to light that I hadn't accounted for, mainly escaped brackets, that aren't inside literal brackets.

The fix is to replace the escaped brackets before checking for any (unescaped) brackets.

Also, I needed to switch the replacements to regex so we replace all occurrences.


gautamkrishnar:
:+1:

PR #1169, opened 2015-06-02
Capitalize GitHub properly

moollaza:
/cc @zekiel


zekiel:
damn hahah I always do that

PR #1172, opened 2015-06-03
Created ScreenResolution, to view your resolution

mattr555:
Displays the screen resolution to the user in a succinct readout

Reduces the need to clickthrough to whatismyscreenresolution.com or similar

`screen resolution`, `what is my display resolution` (any more potential triggers?)

Computer users

[My own](https://duck.co/ideas/idea/5096/screen-resolution)

The wiki summary for "display resolution"

screenshot - 06022015 - 08 57 13 pm

On devices where the pixel density is doubled, it is noted as such:
mobile

## Checklist

```

    - Crunchbang (Debian)
        - [x] Google Chrome

    - iOS (iPhone)
        - [x] Safari Mobile
```

IA Page: https://duck.co/ia/view/screen_resolution

### DIFF
--- a/lib/DDG/Goodie/ScreenResolution.pm
+++ b/lib/DDG/Goodie/ScreenResolution.pm
@@ -20,14 +20,14 @@ triggers startend => "screen resolution", "display resolution", "resolution of m
 handle remainder => sub {
     return unless /^((what\'?s|what is)?\s?(the|my|current))?$/;
 
-    return 'Javascript required', structured_answer => {
+    return structured_answer => {
         id => 'screen_resolution',
         name => 'Screen Resolution',
         data => {
             title => "Your screen resolution is [Loading...]"
         },
         templates => {
-            group => 'text',
+            group => 'icon',
             item => 0,
             options => {
                 moreAt => 0
--- a/lib/DDG/Goodie/ScreenResolution.pm
+++ b/lib/DDG/Goodie/ScreenResolution.pm
@@ -15,10 +15,10 @@ code_url "https://github.com/duckduckgo/zeroclickinfo-goodies/blob/master/lib/DD
 attribution twitter => 'mattr555',
             github => ['mattr555', 'Matt Ramina'];
 
-triggers startend => "screen resolution", "display resolution";
+triggers startend => "screen resolution", "display resolution", "resolution of my screen";
 
 handle remainder => sub {
-    return unless $_ eq '' || $_ eq 'what is my';
+    return unless /^((what\'?s|what is)?\s?(the|my|current))?$/;
 
     return 'Javascript required', structured_answer => {
         id => 'screen_resolution',
--- a/lib/DDG/Goodie/ScreenResolution.pm
+++ b/lib/DDG/Goodie/ScreenResolution.pm
@@ -1,0 +1,39 @@
+package DDG::Goodie::ScreenResolution;
+# ABSTRACT: Return the current screen resolution using javascript
+
+use DDG::Goodie;
+use strict;
+
+zci answer_type => "screen_resolution";
+zci is_cached   => 1;
+
+name "Screen Resolution";
+description "Displays the current screen resolution using javascript";
+primary_example_queries "screen resolution";
+secondary_example_queries "what is my display resolution";
+code_url "https://github.com/duckduckgo/zeroclickinfo-goodies/blob/master/lib/DDG/Goodie/ScreenResolution.pm";
+attribution twitter => 'mattr555',
+            github => ['mattr555', 'Matt Ramina'];
+
+triggers startend => "screen resolution", "display resolution";
+
+handle remainder => sub {
+    return unless $_ eq '' || $_ eq 'what is my';
+
+    return 'Javascript required', structured_answer => {
+        id => 'screen_resolution',
+        name => 'Screen Resolution',
+        data => {
+            title => "Your screen resolution is [Loading...]"
+        },
+        templates => {
+            group => 'text',
+            item => 0,
+            options => {
+                moreAt => 0
+            }
+        }
+    };
+};
+
+1;
--- a/share/goodie/screen_resolution/screen_resolution.js
+++ b/share/goodie/screen_resolution/screen_resolution.js
@@ -1,11 +1,13 @@
 DDH.screen_resolution = DDH.screen_resolution || {};
 
 DDH.screen_resolution.build = function(ops){
-    var res = 'Your screen resolution is ' + window.screen.width + 'x' + window.screen.height;
+    var title = window.screen.width + ' × ' + window.screen.height,
+        sub = ['Your Screen Resolution'];
     if (window.devicePixelRatio && window.devicePixelRatio > 1){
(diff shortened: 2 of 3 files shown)


lib/DDG/Goodie/ScreenResolution.pm
```diff
@@ -0,0 +1,39 @@
...
+triggers startend => "screen resolution", "display resolution";
+
+handle remainder => sub {
+    return unless $_ eq '' || $_ eq 'what is my';
+
+    return 'Javascript required', structured_answer => {
+        id => 'screen_resolution',
+        name => 'Screen Resolution',
+        data => {
+            title => "Your screen resolution is [Loading...]"
```
moollaza: you don't actually have to define this here :)

You could skip passing along a `data` object, and then in your JavaScript `build` you can add to the `ops` object, meaning you can set the `title` there, once you're in the JS.

share/goodie/screen_resolution/screen_resolution.js
```diff
@@ -0,0 +1,13 @@
+DDH.screen_resolution = DDH.screen_resolution || {};
+
+DDH.screen_resolution.build = function(ops){
+    return {
+        onShow: function(){
```
moollaza: As mentioned above, you don't need to actually do this work in the `on_show` -- I think you can do this all in your `build` fn and simply add to the `data` so that the template will handle the displaying and you won't have to look through the DOM and update the content.

moollaza: Sorry to clarify I think you'll want to do something like..

```js
DDH.screen_resolution.build = function(ops){

    // figure out screen resolution here
    // save it

     return {
        data: {
            title: "Your screen resolution is " + screenRes
        }
    }
}
```

moollaza: or I think this might even work...

```js
DDH.screen_resolution.build = function(ops){

    // figure out screen resolution here
    // save it
   ops.data.title = "Your screen resolution is " + screenRes
   return ops;
}
```

mattr555:
@moollaza Hmm, it seems like it is required to pass the `data` object, or else `build()` won't run. However I was able to modify the title without having to use jquery :+1:

mintsoft:
Cool idea!

jarmokivekas:
This is a cool IA. I tested it with four different resolution on Ubuntu using firefox. The results were correct.

Queries I think could safely trigger this answer (my first 3 attempts at triggering):

- whats my screen resolution
- what's my screen resolution
- what is the resolution of my screen

Also allow for question marks and 'display' instead of 'screen' in the examples

I think `[screen|display] resolution` on it's own should *not* trigger this. It will likely cause the IA to trigger to people who are not interested in the resolution of their own screen.

lib/DDG/Goodie/ScreenResolution.pm
```diff
@@ -0,0 +1,39 @@
...
+primary_example_queries "screen resolution";
+secondary_example_queries "what is my display resolution";
+code_url "https://github.com/duckduckgo/zeroclickinfo-goodies/blob/master/lib/DDG/Goodie/ScreenResolution.pm";
+attribution twitter => 'mattr555',
+            github => ['mattr555', 'Matt Ramina'];
+
+triggers startend => "screen resolution", "display resolution";
+
+handle remainder => sub {
+    return unless $_ eq '' || $_ eq 'what is my';
```
mintsoft: I'd probably make it work with:
* `current screen resolution`
* `my screen resolution`
and then also the ones suggested by @jarmokivekas

mattr555:
All of these suggestions should be addressed.

> I think [screen|display] resolution on it's own should not trigger this. It will likely cause the IA to trigger to people who are not interested in the resolution of their own screen.

I disagree here; a query like `screen resolution` allows the user to discover that such an IA exists, and the vertical space is pretty small.

moollaza:
@abeyang this one is ready for review :+1:

mintsoft:
> I disagree here; a query like screen resolution allows the user to discover that such an IA exists, and the vertical space is pretty small. 

I agree, it costs nothing, isn't annoying and would probably increase the number of accidental discoveries

lib/DDG/Goodie/ScreenResolution.pm
```diff
@@ -0,0 +1,39 @@
...
+code_url "https://github.com/duckduckgo/zeroclickinfo-goodies/blob/master/lib/DDG/Goodie/ScreenResolution.pm";
+attribution twitter => 'mattr555',
+            github => ['mattr555', 'Matt Ramina'];
+
+triggers startend => "screen resolution", "display resolution", "resolution of my screen";
+
+handle remainder => sub {
+    return unless /^((what\'?s|what is)?\s?(the|my|current))?$/;
+
+    return 'Javascript required', structured_answer => {
```
moollaza: We don't actually *need* to return a plaintext answer. In this case lets just `return structured_answer => ...`

lib/DDG/Goodie/ScreenResolution.pm
moollaza: Sorry I see now, after digging into the code, that there's an expectation of a defined `data` obj in order for the build function to execute. I think this was an oversight at the time and I've made a task for the team to fix. We shouldn't *need* a `data` obj for IA's like this.

If anything I would set this string to empty, `title => ""`.

I've made an issue for the team to fix this so that `data` isn't required.

abeyang:
Hi, can we use the icon template and utilize the header/subheader? So, it should say:

> **2560 x 1440**
> Your Screen Resolution

and

> **1440 x 900**
> Your Screen Resolution | Pixel Ratio: x2

Finally, can we use `&times;` instead of `x` for the screen resolution? We can use `x` for the pixel ratio, though.

mattr555:
Cool!

screenshot - 06242015 - 11 51 51 am

One thing I noticed is that the resolution itself is a clickable link, is there anyway to fix that?

moollaza:
>One thing I noticed is that the resolution itself is a clickable link, is there anyway to fix that?

That's actually a bug I noticed recently as well. Will make an internal issue.

lib/DDG/Goodie/ScreenResolution.pm
```diff
@@ -0,0 +1,39 @@
...
+            github => ['mattr555', 'Matt Ramina'];
+
+triggers startend => "screen resolution", "display resolution", "resolution of my screen";
+
+handle remainder => sub {
+    return unless /^((what\'?s|what is)?\s?(the|my|current))?$/;
+
+    return structured_answer => {
+        id => 'screen_resolution',
+        name => 'Screen Resolution',
```
moollaza: Ah missed this. Let's change this to "Answer"

abeyang:
Sweet! Visually, it looks good! :+1:

moollaza:
@mattr555 congrats! It's live: https://duckduckgo.com/?q=screen+resolution&ia=answer

Sorry for the delay!

mattr555:
Yay :tada: I can't wait for the tweet from a user excitedly discovering the new functionality :)

Issue #1163, opened 2015-05-30
Convert HTML Goodies to use templates


moollaza:
Many legacy Goodies return HTML strings for rendering. This was previously acceptable as we did not have access to the Handlebars templates used in every Spice IA.

Now that we have templates, we should convert all Goodies returning HTML into templated Goodie that ideally won't need to actually declare any HTML! We have many built-in templates and most IA's can leverage these by specifying only the data and letting our template system do the rest of the work.

In some cases the use of a Handlebars sub-template is necessary and acceptable. In rare cases a custom template is required and we now support that (but please check with us first to make sure its absolutely necessary!)

Let's cleanup our Goodies and hopefully simplify them by moving to our new templates!

Details on template usage can be seen here: https://duck.co/duckduckhack/goodie_displaying

Note: Goodies also support JS and CSS now, which means we should remove any inline scripts and styles and if necessary, move them to separate CSS and JS files.

Here's a list of the existing IA's which return HTML:
- [x] Average.pm
- [x] BashPrimaryExpressions.pm
- [x] BinaryLogic.pm
- [x] BloodDonor.pm
- [x] CalendarToday.pm
- [x] Chess960.pm
- [x] ChineseZodiac.pm
- [x] ColorCodes.pm
- [x] ConvertLatLon.pm
- [x] CrontabCheatSheet.pm
- [x] CryptHashCheck.pm
- [x] CurrencyIn.pm
- [x] Cusip.pm
- [x] Dessert.pm
- [ ] Dewey.pm
- [x] Dice.pm
- [ ] FenViewer.pm
- [x] FIGlet.pm
- [ ] Frequency.pm
- [ ] FrequencySpectrum.pm (this one sorely needs conversion! It should move the styles and scripts into a CSS file and a [x] JavaScript file)
- [x] GoldenRatio.pm
- [x] GuitarChords.pm
- [x] HTMLEntitiesDecode.pm
- [x] HTMLEntitiesEncode.pm
- [ ] IDN.pm
- [x] IndependenceDay.pm
- [ ] IPS.pm
- [ ] JsKeycodes.pm ( there is a Spice PR which will convert this to use JS -- technically it could remain a Goodie and do [x] this...)
- [ ] KernelTaint.pm
- [ ] Latex.pm
- [ ] Loan.pm
- [ ] NameDays.pm
- [ ] NLetterWords.pm
- [ ] Palindrome.pm
- [x] PercentError.pm
[...]

PR #1197, opened 2015-06-20
use record template in UnixPermissions goodie

jarmokivekas:
Moved the UnixPermissions goodie to use a template as part of #1163. If there is a more suitable template let me know. 

Tests still need amending to match the changes.

### DIFF
--- a/lib/DDG/Goodie/UnixPermissions.pm
+++ b/lib/DDG/Goodie/UnixPermissions.pm
@@ -60,19 +60,20 @@ handle query => sub {
     if ($attributes ne '') {
         $octal = "$attributes$octal";
         $plain_output = "Attributes: $attributes{$1}\n";
+        my $tmp_ttributes = $attributes;
         my @symbolic = split '', $symbolic;
-        if ($attributes >= 4) {
-            $attributes -= 4;
+        if ($tmp_ttributes >= 4) {
+            $tmp_ttributes -= 4;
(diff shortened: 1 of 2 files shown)


jarmokivekas:
This is how it looks before and after:

before:
before

after:
after


That last 'Attributes'  key is in the "Show more". Not all answers have that key, though (since it's not always relevant), in which case there is no "Show more" and the answer is displayed in its entirety.

mintsoft:
@jarmokivekas This is excellent, looks much better!

mintsoft:
Although, I think it probably makes more sense to never have the show more in this situation?

jarmokivekas:
I wasn't able to find how to disable the cut. Is there a key that can be set somewhere like in `structured_answer => templates => options`?

jarmokivekas:
I just noticed that in this case the "Show more" automatically disappears (the last item is shown) when the browser window is wider than 850 pixels.

mintsoft:
@jarmokivekas I think you might be right; I can't see an obvious way to 'disable' that behaviour so I feel it might be part of the responsive nature of the templates. @moollaza any clarification?

moollaza:
@jarmokivekas @mintsoft right now there's no way to disable or set the cut-off. It's actually based on height, not # of elements. 

This is great for now. If we find we need to configure the cutoff in other places I'll talk to the team about making it flexible.

@jarmokivekas thanks a ton!

Issue #1175, opened 2015-06-05
USPS: Make sure the query only contains the package number when matching


moollaza:
If we match on the regex pattern of a number, it should be the only thing in the query.

![37346365253153_eur_to_us_at_duckduckgo](https://cloud.githubusercontent.com/assets/873785/8016243/ef9672ca-0bae-11e5-816d-66a81478c9ea.png)

mintsoft:
@moollaza Abandoned?

moollaza:
whoops!

mintsoft:
Fixed by @marianosimone in https://github.com/duckduckgo/zeroclickinfo-goodies/pull/1199

PR #1199, opened 2015-06-20
Do not trigger USPS goodie when query contains other chars besides expected

marianosimone:
add some more tests, and make regexes allow whitespaces but not other chars

Should solve #1175


mintsoft:
LGTM! Thanks for the fix @marianosimone

marianosimone:
:+1: thanks for the quick merge ;)

moollaza:
:+1: Thanks a lot guys!

PR #1200, opened 2015-06-21
Trigger Pi goodie with π

### LINKED ISSUE #1037: Pi: Make it work with π

marianosimone:
as discussed with @mintsoft, adding `use utf8` did the trick.

Also added some new tests for the trigger :dart: 

How it looks before the fix:

before

And after:

after

should fix #1037


mintsoft:
Thanks again for the fix @marianosimone

PR #1227, opened 2015-07-02
Independence Day: Convert to icon template, display country flag

### LINKED ISSUE #1148: Independence Day: Convert to icon template, display country flag
This Goodies should converted to use the [Icon template group](https://duck.co/duckduckhack/templates_overview#icon-template-group) and should display the flag for the given country.

The result should look like this:

independence-day-v1 1-belgium
independence-day-v1 1-romania

**Note:** We'll likely need to use `Locale::Country` to get the 2 letter Country Code and pass that along to the `data` property. We'll then retrieve the country code in a new JavaScript file, that we can use to determine the icon size (based on whether or not we're dealing with a mobile device) and build the URL for the icon asset we need to show using `DDG.settings.region.getSmallIconURL` and `DDG.settings.region.getLargeconURL`

------
IA Page: http://duck.co/ia/view/independence_day

cskksc:
fixes #1148

### DIFF
--- a/lib/DDG/Goodie/IndependenceDay.pm
+++ b/lib/DDG/Goodie/IndependenceDay.pm
@@ -5,6 +5,7 @@ use strict;
 use DDG::Goodie;
 use JSON;
 use utf8;
+use Locale::Country;
 
 zci answer_type => "independence_day";
 zci is_cached   => 1;
@@ -78,9 +79,20 @@ handle query_clean => sub {
 
     return $text,
       structured_answer => {
-        input     => [],
-        operation => $prolog,
-        result    => $date_str
+        id => 'independence_day',
+        templates => {
+            group => "icon",
+            item => 0,
+            elClass => {
+                iconTitle => "t-xxl",
+                iconSubtitle => "tx--15"
+            }
+        },
+        data => {
+            country_code => country2code($country_key),
+            date => $date_str,
+            info => $prolog
+        }
       };
     
 };
--- a/lib/DDG/Goodie/IndependenceDay.pm
+++ b/lib/DDG/Goodie/IndependenceDay.pm
@@ -83,15 +83,15 @@ handle query_clean => sub {
         templates => {
             group => "icon",
             item => 0,
-            elClass => {
-                iconTitle => "t-xxl",
-                iconSubtitle => "tx--15"
+            variants => {
+              iconTitle => 'large',
+              iconImage => 'large'
             }
         },
         data => {
             country_code => country2code($country_key),
-            date => $date_str,
-            info => $prolog
+            title => $date_str,
+            subtitle => $prolog
         }
       };
     
--- a/share/goodie/independence_day/independence_day.css
+++ b/share/goodie/independence_day/independence_day.css
@@ -1,0 +1,3 @@
+.zci__body{
+    display: inline-block;
+}
--- a/share/goodie/independence_day/independence_day.css
+++ b/share/goodie/independence_day/independence_day.css
@@ -1,3 +1,0 @@
-.zci__body{
-    display: inline-block;
-}
--- a/share/goodie/independence_day/independence_day.js
+++ b/share/goodie/independence_day/independence_day.js
@@ -1,19 +1,10 @@
 DDH.independence_day = DDH.independence_day || {};
-
 DDH.independence_day.build = function(ops) {
     return {
         normalize: function(item) {
-            var isMobile = ($('.is-mobile').length > 0) ? true : false;
-            var info = {
-                title: ops.data.date,
-                subtitle: ops.data.info
+            return {
+                image: DDG.settings.region.getLargeIconURL(ops.data.country_code)
             };
-            if (isMobile) {
-                info.image = DDG.settings.region.getSmallIconURL(ops.data.country_code);
-            } else {
-                info.image = DDG.settings.region.getLargeIconURL(ops.data.country_code);
-            }
-            return info;
         }
     };
 };
--- a/share/goodie/independence_day/independence_day.js
+++ b/share/goodie/independence_day/independence_day.js
@@ -1,9 +1,12 @@
 DDH.independence_day = DDH.independence_day || {};
 DDH.independence_day.build = function(ops) {
+    var flagUrl = DDG.settings.region.getLargeIconURL(ops.data.country_code);
+    // makes flag icon url point to a 64 px image.
+    flagUrl = flagUrl.replace(/\/\d+\//, "/64/");
     return {
         normalize: function(item) {
             return {
-                image: DDG.settings.region.getLargeIconURL(ops.data.country_code)
+                image: flagUrl
             };
         }
     };
--- a/share/goodie/independence_day/independence_day.js
+++ b/share/goodie/independence_day/independence_day.js
@@ -1,0 +1,19 @@
+DDH.independence_day = DDH.independence_day || {};
+
+DDH.independence_day.build = function(ops) {
+    return {
+        normalize: function(item) {
+            var isMobile = ($('.is-mobile').length > 0) ? true : false;
+            var info = {
+                title: ops.data.date,
+                subtitle: ops.data.info
+            };
+            if (isMobile) {
+                info.image = DDG.settings.region.getSmallIconURL(ops.data.country_code);
+            } else {
+                info.image = DDG.settings.region.getLargeIconURL(ops.data.country_code);
+            }
+            return info;
+        }
+    };
+};
--- a/t/IndependenceDay.t
+++ b/t/IndependenceDay.t
@@ -16,15 +16,15 @@ sub answer {
         id => "independence_day",
         data => {
             country_code => country2code($_[0]),
-            date => $_[1],
-            info => "Independence Day of " . $_[0]
+            title => $_[1],
+            subtitle => "Independence Day of " . $_[0]
         },
         templates => {
             group => "icon",
             item => "0",
-            elClass => {
-                iconTitle => "t-xxl",
-                iconSubtitle => "tx--15"
+            variants => {
+              iconTitle => 'large',
+              iconImage => 'large'
             }
         }
     });
--- a/t/IndependenceDay.t
+++ b/t/IndependenceDay.t
@@ -4,149 +4,50 @@ use strict;
 use warnings;
 use Test::More;
 use DDG::Test::Goodie;
+use Locale::Country;
 
 zci answer_type => "independence_day";
 zci is_cached   => 1;
 
+sub answer {
+    my $prolog;
+    $prolog = "Independence Day of " . $_[0] . " " . $_[1];
+    test_zci($prolog, structured_answer => {
+        id => "independence_day",
+        data => {
+            country_code => country2code($_[0]),
+            date => $_[1],
+            info => "Independence Day of " . $_[0]
+        },
+        templates => {
+            group => "icon",
+            item => "0",
+            elClass => {
+                iconTitle => "t-xxl",
+                iconSubtitle => "tx--15"
+            }
+        }
+    });
+}
+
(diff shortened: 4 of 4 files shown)


t/ColorCodes.t
```diff
@@ -49,7 +49,7 @@ ddg_goodie_test(
...
.../color/ff0000' title='Tints, information and similar colors on color-hex.com' class='tx-clr--dk2'>Info</a><span class='separator'> | </span><a href='/?q=color%20picker%20%23ff0000' class='tx-clr--dk2'>Picker</a></p></div><div class='column2 tx-clr--dk2'><div class='complementary'><div class='cols_column'><a href='/?q=color%20picker%20%2300FFFF' class='mini-color circle' style='background: #00FFFF'></a></div><div class='desc_column'><p class='no_vspace'>Complementary #:</p><p class='no_vspace'><a onclick='document.x.q.value="#00FFFF";document.x.q.focus();' href='javascript:' class='tx-clr--lt'>00FFFF</a></p></div></div><div><div class='cols_column'><a href='/?q=color%20picker%20%23ff8000' class='mini-color circle' style='background: #ff8000'> </a><a href='/?q=color%20picker%20%23ff0080' class='mini-color circle' style='background: #ff0080'> </a></div><div class='desc_column'><p class='no_vspace'>Analogous #:</p><p class='no_vspace'><a onclick='document.x.q.value="#FF8000";document.x.q.focus();' href='javascript:' class='tx-clr--lt'>FF8000</a>, <a onclick='document.x.q.value="#FF0080";document.x.q.focus();' href='javascript:' class='tx-clr--lt'>FF0080</a></p></div></div></div></div>)
```
cskksc: Better diff:
```html
<a href='http://labs.tineye.com/multicolr/#colors=ff0000;weights=100;
```
vs
```html
<a href='http://labs.tineye.com/multicolr#colors=ff0000;weights=100;
```
https://github.com/duckduckgo/zeroclickinfo-goodies/blob/master/lib/DDG/Goodie/ColorCodes.pm#L105

t/FedEx.t
```diff
@@ -9,28 +9,28 @@ zci answer_type => 'fedex';
...
-        '9612804882227378545377' => test_zci(
-        	"9612804882227378545377",
-        	heading => 'FedEx Shipment Tracking',
-        	html => qq(Track this shipment at <a href="http://fedex.com/Tracking?tracknumbers=9612804882227378545377&action=track">FedEx</a>.)
-        )
+    [qw( DDG::Goodie::FedEx )],
+    'fedex 9241990100130206401644' => test_zci(
+        "9241990100130206401644",
+        heading => 'FedEx Shipment Tracking',
+        html => qq(Track this shipment at <a href="https://www.fedex.com/apps/fedextrack/?action=track&tracknumbers=9241990100130206401644&action=track">FedEx</a>.)
```
cskksc: https://github.com/duckduckgo/zeroclickinfo-goodies/blob/master/lib/DDG/Goodie/FedEx.pm#L118

t/ScreenResolution.t
```diff
@@ -8,14 +8,14 @@ use DDG::Test::Goodie;
...
-my @answer = test_zci('Javascript required', structured_answer => {
+my @answer = test_zci('Your screen resolution is [Loading...]', structured_answer => {
     id => 'screen_resolution',
     name => 'Screen Resolution',
     data => {
         title => "Your screen resolution is [Loading...]"
     },
     templates => {
-        group => 'text',
+        group => 'icon',
```
cskksc: https://github.com/duckduckgo/zeroclickinfo-goodies/blob/master/lib/DDG/Goodie/ScreenResolution.pm#L30

lib/DDG/Goodie/IndependenceDay.pm
```diff
@@ -78,9 +79,16 @@ handle query_clean => sub {
 
     return $text,
       structured_answer => {
-        input     => [],
-        operation => $prolog,
-        result    => $date_str
+        id => 'independence_day',
+        templates => {
```
cskksc: Should all the templating be done in JS like for spice IA? But if I do that, I get an error, 
```
Can't use an undefined value as an ARRAY reference at /home/codio/perl5/perls/perl-5.18.2/lib/site_perl/5.18.2/App/DuckPAN/Web.pm line 346
in App::DuckPAN::Web::request at /home/codio/perl5/perls/perl-5.18.2/lib/site_perl/5.18.2/App/DuckPAN/Web.pm line 346
  343:                                                 last;
  344:                                         }
  345:                                         else{ # auto-template
  346:                                                 my $input_size = @{$structured->{input}};
  347:                                                 die "Auto-templating with input of size $input_size called for '$structured->{operation}', "
  348:                                                         . "while only sizes up to 2 are supported" if $input_size >= 3;
  349:                                                 my $template_name = 'goodie_'.$input_size.'_inputs';
```
So have kept it here. Am I missing something ?

cskksc:
This is how it looks,
final

moollaza:
@cskksc it looks like you mixed in updates to other IA's in here? Let's keep this PR to Independence day changes, please.

lib/DDG/Goodie/IndependenceDay.pm
moollaza: We brought templating to Goodies so that pure Perl IA's could leverage templates without the need for a separate JS file.

Spice used templates first because they require JS to handle the API response but Goodies do not.

Now that we've added template support to Goodies they're able to define their own data and templates making them much more capable of producing quality IA's that meet our design goals.

lib/DDG/Goodie/ScreenResolution.pm
```diff
@@ -20,7 +20,8 @@ triggers startend => "screen resolution", "display resolution", "resolution of m
 handle remainder => sub {
     return unless /^((what\'?s|what is)?\s?(the|my|current))?$/;
 
-    return structured_answer => {
+    return "Your screen resolution is [Loading...]",
```
moollaza: This should really be in another PR...

We actually don't want to return a plain text result for this IA. The plaintext result is used by the non-JS version of the site and the Instant Answer API. Seeing as this IA requires JS to provide your screen resolution we can't/shouldn't provide a plain text result as it's basically useless ;)

lib/DDG/Goodie/IndependenceDay.pm
moollaza: >Should all the templating be done in JS like for spice IA? But if I do that, I get an error, 

Nope. We can define the templates in the Perl so we might as well. If necessary, the `structured_answer` hash can be modified in the JS (which means templates could be specified) but in this case we don't really need to.

As long as you define your `data` hash though, you should be able to define the templates in the JS.

moollaza:
@cskksc btw, if you move the "fixes #1148" into the description of your PR (rather than the title, which should ideally be a little more descriptive), GitHub will automatically close the specified issue(s) when this PR is merged :smile:

lib/DDG/Goodie/ScreenResolution.pm
cskksc: Yeah that makes sense. Ill remove it. But what about changes to fedex and color_codes tests ? The existing tests don't pass. There are some minor differences in the returned html. That should be done in another PR ?

cskksc:
@moollaza Reverted the commit. Had made the changes mostly because ```duckpan test``` was failing on my environment. But may be travis test strategy is different, since it passed. Now should I rebase and delete the commit entirely ? Or is it ok to have a revert commit in the log history.

share/goodie/independence_day/independence_day.js
```diff
@@ -0,0 +1,19 @@
+DDH.independence_day = DDH.independence_day || {};
+
+DDH.independence_day.build = function(ops) {
+    return {
+        normalize: function(item) {
+            var isMobile = ($('.is-mobile').length > 0) ? true : false;
```
moollaza: there's actually a JS variable, `is_mobile` that will be `1` when we're dealing with a moile device

lib/DDG/Goodie/IndependenceDay.pm
```diff
@@ -78,11 +79,22 @@ handle query_clean => sub {
...
+            group => "icon",
+            item => 0,
+            elClass => {
+                iconTitle => "t-xxl",
+                iconSubtitle => "tx--15"
+            }
+        },
+        data => {
+            country_code => country2code($country_key),
+            date => $date_str,
```
moollaza: might as well just set these here as the `title` and `subtitle`. Then we don't need to bother with them in the JS

share/goodie/independence_day/independence_day.js
moollaza: so we can remove this line and just check `if (is_mobile) { ... }` beloq

lib/DDG/Goodie/IndependenceDay.pm
```diff
@@ -78,11 +79,22 @@ handle query_clean => sub {
...
     return $text,
       structured_answer => {
-        input     => [],
-        operation => $prolog,
-        result    => $date_str
+        id => 'independence_day',
+        templates => {
+            group => "icon",
+            item => 0,
+            elClass => {
```
moollaza: Let's handle this with `variants` instead:
variants => {
    iconTitle => 'large',
    iconImage => 'large'
}
```

Should looks good
```perl

File: share/goodie/independence_day/independence_day.css [hunk 9]
```diff
@@ -0,0 +1,3 @@
+.zci__body{
```
moollaza: Please remove. Setting the size on the icon fixes the layout issue you were solving with this.

share/goodie/independence_day/independence_day.js
```diff
@@ -0,0 +1,19 @@
...
+            var info = {
+                title: ops.data.date,
+                subtitle: ops.data.info
+            };
+            if (isMobile) {
+                info.image = DDG.settings.region.getSmallIconURL(ops.data.country_code);
+            } else {
+                info.image = DDG.settings.region.getLargeIconURL(ops.data.country_code);
```
moollaza: It looks like the 32px icon is too small. If we grab the `large` icon and `replace("32", "64")` it looks much better and should work for both mobile and desktop.

moollaza:
@cskksc this is great! Thanks a lot :+1: 

With my suggested fxes, the code should be a bit simpler, and the result looks a bit better:

independence_day_of_canada_at_duckduckgo

cskksc:
Yeah. Had made quite a few mistakes there. I've made the changes. About the logo size, when we take the ```large``` icon, it already returns a ```64px``` icon. So we don't have to do anything different for mobile and PC right ? We can just use the same ```large``` icon ?

screen shot 2015-07-08 at 12 06 29 am

Sloff:
This will also fix part of #1163

moollaza:
>Yeah. Had made quite a few mistakes there. I've made the changes. About the logo size, when we take the large icon, it already returns a 64px icon. So we don't have to do anything different for mobile and PC right ? We can just use the same large icon ?

@cskksc yes! Sorry for the delay. That should be fine :+1:

moollaza:
@cskksc it looks like we're still showing the `32px` image?

canada_independence_day_at_duckduckgo

cskksc:
@moollaza Sorry I couldn't do this sooner. Have done a regex replace to convert the url to point to 64px image. But its strange. On my codio environment, ```getLargeIconURL``` already returned a ```64px``` image. 
```getSmallIconURL``` returns ```40px```  and ```getXSmallIconURL``` returns ```32px```.

moollaza:
@cskksc sorry for the late response! Thanks for updating, I'll take a look shortly.

moollaza:
@cskksc LGTM :+1:

moollaza:
Thanks a lot!

cskksc:
Thanks :smiley: .

PR #1229, opened 2015-07-04
Name Days: Set max-width on table

### LINKED ISSUE #1083: NameDays: Table shouldn't extend past organic link width
Setting a `max-width` of `44em` for the `<table>` or the container should make things look better!

altern8tif:
Fixes #1083

### DIFF
--- a/lib/DDG/Goodie/NameDays.pm
+++ b/lib/DDG/Goodie/NameDays.pm
@@ -37,18 +37,18 @@ my %dates = share('preprocessed_dates.txt')->slurp(iomode => '<:encoding(UTF-8)'
 sub parse_other_date_formats {
     # Quick fix for the date formats not supported by parse_datestring_to_date.
     # If parse_datestring_to_date will be improved, you can remove some of the following code.
-    
+
     # US date format ("month/day")
     if (/^([0-1]?[0-9])\s?\/\s?([0-3]?[0-9])$/) {
         # Suppress errors for invalid dates with eval
         return eval { new DateTime(year => 2000, day => $2, month => $1) };
     }
-    
+
     # Polish date format ("day.month")
     if (/^([0-3]?[0-9])\s?\.\s?([0-1]?[0-9])$/) {
         return eval { new DateTime(year => 2000, day => $1, month => $2) };
     }
-        
+
     # Polish month names
     s/\b(styczeń|stycznia)\b/Jan/i;
     s/\b(luty|lutego)\b/Feb/i;
@@ -62,7 +62,7 @@ sub parse_other_date_formats {
     s/\b(październik|października)\b/Oct/i;
     s/\b(listopad|listopada)\b/Nov/i;
     s/\b(grudzień|grudnia)\b/Dec/i;
-    
+
     # Czech month names
     s/\b(leden|ledna)\b/Jan/i;
     s/\b(únor|února)\b/Feb/i;
@@ -76,13 +76,13 @@ sub parse_other_date_formats {
     s/\b(říjen|října)\b/Oct/i;
     s/\b(listopad|listopadu)\b/Nov/i;
     s/\b(prosinec|prosince)\b/Dec/i;
-    
+
     # Parse_datestring_to_date uses the current year if the year is not specified, so
     # it will not parse "29 Feb" in a non-leap year. Fix this problem here.
     if (/^29\s?(?:th)?\s*(Feb|February)/ || /(Feb|February)\s*29\s?(?:th)?$/) {
         return new DateTime(year => 2000, day => 29, month => 2);
     }
-    
+
     return parse_datestring_to_date($_);
 }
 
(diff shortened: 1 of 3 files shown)


share/goodie/name_days/name_days.css
```diff
@@ -56,3 +56,7 @@ tr:last-of-type .name-days-country, tr:last-of-type .name-days-dates {
         width: 100%;
     }
 }
+
+table.name-days {
```
moollaza: oh man, I didn't realize none of the CSS for this was properly namespaced. That first line will affect everything on the page.

Mind adding namespacing (`.zci--name_days`) to that? Alternatively, converting it to use a template (so we're not dumping HTML onto the page) would be greatly appreciated :)

altern8tif: Sure, will convert it to a template since that is preferred.  Should be a `list` template?

moollaza: @altern8tif I believe so, but it may be a little difficult as we don't have the ability to create columns with the list template + `record` sub-template (it only supports a key and value per row).

You may need to use a custom sub-template with the List template group -- does that make sense?

altern8tif: @moollaza it does (somewhat), but I'll probably have to re-read the documentation again.  Will ping you if I need any help.

altern8tif: @moollaza Ok the Perl code is a bit over my head.  Realised that [preprocess.pl](https://github.com/duckduckgo/zeroclickinfo-goodies/blob/master/share/goodie/name_days/preprocess.pl) was dumping HTML table tags into the dates and names `txt` file which will probably mess with DDG's `list` template.  Kinda gave up a bit after that.

I've removed the `.zci__body` class in css – it's not affecting anything anyway.  And decided to namespace the css instead.

Guess someone else has got to do the heavy-lifting to convert this to a template :disappointed_relieved: 

Let me know if this works.

altern8tif:
@moollaza ping.

altern8tif:
@moollaza ping again?

share/goodie/name_days/name_days.css
```diff
@@ -1,58 +1,58 @@
...
 
-.name-days-country, .name-days-dates {
+.zci--name_days .name-days-country, .zci--name_days .name-days-dates {
     border-bottom: solid 1px #DDD;
     padding-top: 0.3em;
     padding-bottom: 0.3em;
 }
 
-tr:last-of-type .name-days-country, tr:last-of-type .name-days-dates {
+tr:last-of-type .zci--name_days .name-days-country, tr:last-of-type .zci--name_days .name-days-dates {
```
moollaza: These rules should be reversed. The `.zci--name_days` should be before the `tr:last-of-type` and let's put each rule on it's own line for legibility.

It looks like the rule isn't working right now as the last element in the table has a bottom border. In production it's working correctly though.

share/goodie/name_days/name_days.css
```diff
@@ -1,58 +1,58 @@
...
     padding-left: 0.4em;
 }
 
-.name-days-country .flag-sm {
+.zci--name_days .name-days-country .flag-sm {
     vertical-align: middle;
 }
 
-.name-days-country, .name-days-dates {
+.zci--name_days .name-days-country, .zci--name_days .name-days-dates {
```
moollaza: again, let's split up selectors onto newlines

lib/DDG/Goodie/NameDays.pm
```diff
@@ -123,22 +123,22 @@ handle remainder => sub {
...
                                         '</td><td class="name-days-dates">'  . $2 . '</td></tr>'@ge;
-        
+
         $header = 'Name days on <b>' . html_enc($query) . '</b>';
     }
-    
+
     # Add the header
     $html = '<span>' . $header . '</span>' .
-        '<div class="zci__content"><table>' .
```
moollaza: We actually want the `zci__content` as it provides some top spacing between the name of the IA and the table. Right now they're a little too close. Take a look at the live IA to see the difference.

Really the `zci__name_days` class should be applied higher up in the DOM (where right now we're applying a `zci__answer` class, which wraps *all* of the content. If anything, the `<div class="zci--name_days">` should be the first element so that it contains the `<span>` containing the title.

altern8tif:
@moollaza Sorry for the delay! Ready for review...

moollaza:
@altern8tif one last fix...Looks like the new IA setting the colour of the text to `#666` instead of `#333`. Let's fix that, and then this is good for merge :+1:

altern8tif:
@moollaza OK. Reinstated the original ``color: #333`` property to the css.

altern8tif:
@moollaza ping

moollaza:
@altern8tif sorry for the delay! This LGTM :+1: 

Thanks for the awesome work :smiley:

PR #1233, opened 2015-07-05
new added top footballer.json

vishwakulkarni:
created a cheat sheet for top footballer.

### DIFF
--- a/share/goodie/cheat_sheets/footballer.json
+++ b/share/goodie/cheat_sheets/footballer.json
@@ -1,0 +1,36 @@
+{
+    "id": "top_footballer", 
+
+    "name": "top footballers",
+
+    "description": "List of top footballer", 
+    
+    "metadata": { 
+        "sourceName": "topfootballer",
+        "sourceUrl": "http://www.goal.com/en-india/teams/spain/barcelona/2017" 
+    },
+ "section_order" : ["A","B"],
+    "sections": {
+        "A": [
+            {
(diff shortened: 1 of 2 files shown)


MariagraziaAlastra:
Hi @vishwakulkarni, thanks for the contribution! Do you mind adding the [the Goodies PR template](https://github.com/duckduckgo/zeroclickinfo-goodies/blob/master/pull_request_template_goodie.md) filled in, in your first comment?

zachthompson:
@vishwakulkarni Don't these rankings vary over time?  I believe this would become outdated pretty quickly if I understand the source correctly.

altern8tif:
I think this may be more useful as a Spice – using footballer's names as queries to show their rankings.

Cheatsheets should be more for information that rarely changes, exhaustive, more of a source for reference etc.  May be worthwhile to consider certain guidelines for what are good candidates for cheat sheets.

zachthompson:
Thanks, but needs to update dynamically to reflect changes.

moollaza:
@vishwakulkarni thanks a lot for contributing, sorry we can't use this PR though. We'd love for you to stick around though and continue to contribute. There's many issues in the repo's that need to be addressed and there are also many Ideas on or Forum that are awaiting a developer.

If you're going to work on a new Instant Answer please reach out to use first to make sure the idea has been approved so that we won't run into anything like this again.

Thanks again!

PR #1244, opened 2015-07-06
Password: Set minimum password length

### LINKED ISSUE #1145: Password: Implement minimum password length
The password generator is giving incorrect reports about password strength: 

screenshot-6

I'm pretty sure a one character long password isn't "high strength". 

IA: https://duck.co/ia/view/password

altern8tif:
Fixes #1145 

screen shot 2015-07-06 at 4 03 44 pm
screen shot 2015-07-06 at 4 03 59 pm
screen shot 2015-07-06 at 4 08 03 pm

### DIFF
--- a/lib/DDG/Goodie/Password.pm
+++ b/lib/DDG/Goodie/Password.pm
@@ -23,7 +23,6 @@ triggers start => 'password', 'random password', 'pw', 'random pw', 'pwgen';
 
 use constant MAX_PWD_LENGTH => 64;
 use constant MIN_PWD_LENGTH => 8;
-use constant DEFAULT_PWD_LENGTH => MIN_PWD_LENGTH;
 
 my %look_alikes = map { $_ => 1 } qw(l O I);    # Exclude alphabet characters which can be easily visually confused.
 my %averages = map { $_ => 1 } (2 .. 9);        # 0,1 missing for the same reasons as above.
@@ -59,7 +58,7 @@ handle remainder => sub {
     my @q_words = map { lc $_ } grep { defined } ($+{'fw'}, $+{'sw'});
 
     my $pw_length = first { looks_like_number($_) } @q_words;
-    $pw_length = ($pw_length) ? max(MIN_PWD_LENGTH, $pw_length) : DEFAULT_PWD_LENGTH;
+    $pw_length = ($pw_length) ? max(MIN_PWD_LENGTH, $pw_length) : MIN_PWD_LENGTH;
 
     return if ($pw_length > MAX_PWD_LENGTH);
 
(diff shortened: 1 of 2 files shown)


lib/DDG/Goodie/Password.pm
```diff
@@ -22,7 +22,8 @@ zci is_cached   => 0;
 triggers start => 'password', 'random password', 'pw', 'random pw', 'pwgen';
 
 use constant MAX_PWD_LENGTH => 64;
-use constant DEFAULT_PWD_LENGTH => 8;
+use constant MIN_PWD_LENGTH => 8;
+use constant DEFAULT_PWD_LENGTH => MIN_PWD_LENGTH;
```
MrChrisW: Don't think there's any need to do this, apart from making the code more readable you can just use the `MIN_PWD_LENGTH` const on line 62.

altern8tif: Haha ok... wasn't thinking clearly when I did that...

MrChrisW: haha all good :smile:

moollaza:
@altern8tif this is great. If they requests a password length that's below our minimum thoguh, we should just not show anything -- just like how we handle passwords that are too long.

altern8tif:
@moollaza I was actually contemplating returning a message that says "A secure password has a minimum of 8 characters" or something to that effect.  It's easier to just return `undef` I guess!

On a separate note, it seems funny that we give the option of allowing the user to specify the password strength – wouldn't most people prefer to not have a weak password?

I was actually toying with the idea of incorporating this feature ([How Secure Is My Password](https://howsecureismypassword.net/)) or having it as a separate IA.  But also not so sure if people are comfortable keying in their passwords into a search engine (even one as privacy-focused as DDG!).

Thoughts?

mintsoft:
@altern8tif I tend to agree with moollaza here, I think returning `undef` is wiser as there's quite a discussion involved in what actually constitutes a "secure" password

moollaza:
@altern8tif I agree that requesting a "low strength" password seems weird. I think it may be better in V2 of this IA to have more a of UI that let's you choose options that would determine the security, like allowing non-alphanumeric character, etc

MrChrisW:
@moollaza V2 should move to generating passwords using JS, allowing options similar to http://strongpasswordgenerator.com/ (example)? Sounds like a good idea.

mintsoft:
@MrChrisW I like that idea; also helps with the privacy concerns of "are DDG storing my passwords!"

jagtalon:
@altern8tif wouldn't we still want "random password 4" but not "random password 4 strong"?

mintsoft:
@jagtalon do you mean that it should output a 4 character password?

altern8tif:
@jagtalon in the related issue, we were discussing whether to [implement a minimum password length](https://github.com/duckduckgo/zeroclickinfo-goodies/issues/1145#issuecomment-105338421) (and to set it as a default).  

This PR is mainly to address that – so users are limited to 8-character (min length) or 64-character (max length) passwords.  So long as a number not in the 8-64 character range is specified then IA will return 
`undef`.

moollaza:
This LGTM :+1:

PR #1251, opened 2015-07-07
Improving contributing file to reflect contributor guide

talsraviv:
Changing to reflect the messaging in the contributing guide https://duck.co/help/contributing/create-instant-answer

[Same as PR on spice](https://github.com/duckduckgo/zeroclickinfo-spice/pull/2003)

cc @zekiel

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,49 +1,54 @@
-# Create an Instant Answer
+# Contribute to Goodie Instant Answers
 
-Creating an Instant Answer is the most exciting way to contribute to the community.
+There are several options for contributing Goodie Instant Answers, explained below.
 
-Whether you want to improve your personal search results, bring exposure to a source of great data, or simply enjoy collaborating with smart people around the world, your live code will help millions of searchers every day.
+You can always find the [Instant Answer Documentation here](https://duck.co/duckduckhack/ddh-intro). Also, if you have any questions at any point, feel free to ask on one of our community channels:
 
-Learn the [big picture of creating Instant Answers](https://duck.co/duckduckhack/ddh-intro).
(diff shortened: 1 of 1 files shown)


zekiel:
@talsraviv  I like it but wouldn't want this to exist as documentation --just a list of steps. Where to look, what to do, etc. "A launchpad" 

Would recommend the following:

1. Section order should be: 
 --Quick Start
 --Create New
 --Improve Current

2. Would replace Cheat Sheets w/ QuickStart goodie 

3. Take out language that makes it seem like each section is necessary before moving to the next. 

3. Each section should have a 1,2,3 list to get started. For example, the Create New section should be something like: 

1. Get an idea
   -grab one from the top voted or think of your own
2. Let us know what you're working on by emailing open@duckduckgo.com
3. Start coding here: duck.co/duckduckhack/.../

talsraviv:
Makes sense. Can you clarify number 2?

talsraviv:
@zekiel whatchoo think

zekiel:
@talsraviv  on #2 --> I meant suggesting a Cheat Sheet over the QuickStart bc it's actually valuable. It seems like we're advocating people to make QuickStart IAs since it's in contributing.md. To me, this should show people where to make an impact so would give them something real like CheatSheets

talsraviv:
@zekiel I thought about this for a bit and realized they're both valuable for different user goals. Check out what I did.

My thinking if it's valuable:
- If *learning the goodie framework* is the goal, the quick start is the best (because cheat sheets don't touch on the PM file or tests or the full loop - and they require doing some content research)
- If *contributing quickly* is the goal, then cheat sheets are better, for those same reasons.

talsraviv:
Reminder to self... if this is cool to merge I'll do the same for Spice repo

zekiel:
@talsraviv  nailed it! LGTM

moollaza:
LGTM :+1:

Issue #1202, opened 2015-06-22
UnixPermissions: Should trigger on "unix file permission", "unix permission", etc


moollaza:
While testing this, I realized there were some missing triggers for this IA.

It looks like it triggers for these queries, but the logic doesn't expect "unix" or "unix file" to precede the trigger word. This IA should be adjusted to handle that.

PR #1256, opened 2015-07-07
Allow "unix" and "files?" in UnixPermissions trigger

marianosimone:
As reported in #1202, if "unix", "file" or "files" was in the query, the IA was triggered but then discarded:

notfixed

This fixes it:

now

### DIFF
--- a/t/UnixPermissions.t
+++ b/t/UnixPermissions.t
@@ -6,6 +6,28 @@ use DDG::Test::Goodie;
 zci answer_type => 'unix_permissions';
 zci is_cached => 1;
 
+sub _expected_result {
+    return {
+            id => 'UnixPermissions',
+            description => 'Unix file permission',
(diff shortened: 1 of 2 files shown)


t/UnixPermissions.t
```diff
@@ -6,18 +6,8 @@ use DDG::Test::Goodie;
...
-    )],
-    'chmod 755' => test_zci(
-'755 (octal)
-rwxr-xr-x (symbolic)
-User: read, write and execute
-Group: read and execute
-Others: read and execute
-',
-        structured_answer => {
+sub _expected_result {
```
marianosimone: most of the expected result is the same, so we can reuse it and make it easier to write new tests :tada:

moollaza:
LGTM :+1: 

Thanks @marianosimone !

PR #1260, opened 2015-07-08
BloodDonor: Switched to record template

mintsoft:
Same as before really; just using structured answer:

image


mintsoft:
Just noticed this: 
image
is the column width adjustable somehow?

javathunderman:
@mintsoft I'm not sure about the column width, but I'm guessing that you can somehow overwrite the CSS...

ping @moollaza

moollaza:
@mintsoft @javathunderman the CSS route is the only method we have for now, perhaps later it could be a property of the record template data, but the CSS route should be OK for now.

lib/DDG/Goodie/BloodDonor.pm
```diff
@@ -36,43 +36,64 @@ sub table_data {
...
+            push(@criticalResults, $donorType . '-');
         }
-
-        $html .= '</table>';
-        return $output, html => $html, heading => "Donors for blood type ".uc($_);
     }
-    return;
+    
+    my $output = '';
+    my $html = "<table class='blooddonor'>";
```
moollaza: looks like we can remove this :)

lib/DDG/Goodie/BloodDonor.pm
```diff
@@ -36,43 +36,64 @@ sub table_data {
...
+
+    sub to_text
+	{
+		my ($data, $keys) = @_;
+		return join "\n", map {"$_: $data->{$_}";} @{$keys};
+	}
+
+    return to_text(\%record_data, \@record_keys), structured_answer => {
+            id => 'BloodDonor',
+            description => 'Returns available donors for a blood type',
```
moollaza: Is there a good place we can send them with a more at link? Perhaps Wikipedia?

moollaza: If so, let's add a `meta: {}` block

mintsoft:
@moollaza all sorted. One this that struck me as odd is that the properties in the meta hash are camelCase, whilst the ones in templates are snake case: (https://duck.co/duckduckhack/goodie_displaying#setting-display-properties-in-a-goodies-perl)

moollaza:
>One thing that struck me as odd is that the properties in the meta hash are camelCase, whilst the ones in templates are snake case:

Hahaha I never noticed that! I guess it shows the API has been changing a bit. I believe we're gravitating towards CamelCase but will check. I'm sure one day the API will have unified casing ;)

moollaza:
LGTM :+1:

share/goodie/blood_donor/blood_donor.css
```diff
@@ -1,20 +1,3 @@
...
-{
-	border-top: 1px solid rgba(155,155,155,.15);
-}
-
-.zci--answer .blooddonor tr:nth-of-type(1)
-{
-	border-top: 0px;
-}
\ No newline at end of file
+.record .record__cell--key {
```
moollaza: Whoops just noticed this isn't namespaced with `.zci--blood_donor`

mintsoft:
@moollaza Well spotted!

jagtalon:
Excellent. :+1: 

burns

Issue #1243, opened 2015-07-05
Calculator does not support multiplication symbol


Asherehsa:
The calculator solves 3 * 3 (three asterisk three) and 3 x 3 (three ex three) but not when you use the proper multiplication symbol 3 × 3

RomainLG:
Dammit, I've been trying but I'm not sure you can use the plaintext symbol, maybe I should use it's encoding numbers 

https://en.wikipedia.org/wiki/Multiplication_sign#In_computer_software

Anyway, who types it that way into their browsers?! ^^'

Sloff:
Hi @mintsoft, I think this issue can be closed as it was resolved in #1264

mintsoft:
@Sloff It was indeed; thanks for reminded me

PR #1262, opened 2015-07-09
support of multiplication unicode

RomainLG:
Sorry for the lack of description it's #1243 fix

Will need to add support to the minus unicode if we go that way.

But the whole thing is unnecessary IMO

### DIFF
--- a/lib/DDG/Goodie/Calculator.pm
+++ b/lib/DDG/Goodie/Calculator.pm
@@ -24,29 +24,30 @@ triggers query_nowhitespace => qr<
         ^
        ( what is | calculate | solve | math )?
 
-        [\( \) x X * % + / \^ \$ -]*
+        [\( \) x X \x{00D7} * % + / \^ \$ -]*
 
         (?: [0-9 \. ,]* )
         (?: gross | dozen | pi | e | c | squared | score |)
-        [\( \) x X * % + / \^ 0-9 \. , _ \$ -]*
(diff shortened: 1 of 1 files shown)


moollaza:
@RomainLG I believe you should be OK to put the actual Unicode character in the code, but you'll need to add `use utf8;` to the package.

> But the whole thing is unnecessary IMO

I agree it's not very likely to occur, but it doesn't hurt to support this. Makes it seem way more "magical" when it actually works without the user expecting it to :smile:

moollaza:
@RomainLG it looks like another contributor has made a PR wich adds this functionality (though they did it after you!)

Would you like to update your PR accordingly, or shall I close this and merge theirs?

See #1262

RomainLG:
@moollaza 

Merge his! He did it better than I did!

I've got some question though, he used the unicode `×` directly whereas I used the corresponding code `\x{00D7}` because upon testing the unicode wasn't working.

Any idea why it's that way? Is it better to use the unicode?

What is the `*.t` file use? It's so the instant answer can get tested automatically?

mintsoft:
@RomainLG Sorry for the delay. Personally I prefer to `use utf8` and have the symbols in the source as it's easier to see what's happening at a glance rather than having to look up codes. 

The `*.t` file is so we can write tests to ensure that we don't break existing functionality when we improve or extend instant answers!

mintsoft:
@RomainLG I'm going to close your PR for the time being in favour of https://github.com/duckduckgo/zeroclickinfo-goodies/issues/1264 Thanks for the contribution anyway!

PR #1264, opened 2015-07-09
Calculator: Added multiplication symbols and obelus along with tests

phylum:


moollaza:
@phylum thanks a lot for contributing! We really appreciate it :)

Another PR (#1262) was actually made before yours which adds some of these changes, but not all of them.

I've asked the dev there to either add in the changes you've made here, or close his PR in favour of yours. Sound good?

lib/DDG/Goodie/Calculator.pm
```diff
@@ -18,25 +18,26 @@ name 'Calculator';
...
-attribution github  => ['https://github.com/duckduckgo', 'duckduckgo'];
+attribution github  => ['https://github.com/duckduckgo', 'duckduckgo'],
+            github  => ['https://github.com/phylum', 'Daniel Smith'];
 
 triggers query_nowhitespace => qr<
         ^
        ( what is | calculate | solve | math )?
 
-        [\( \) x X * % + / \^ \$ -]*
+        [\( \) x X × ∙ ⋅ * % + ÷ / \^ \$ -]*
```
moollaza: I believe we should add `use utf8;` to the package now that these exist in the source.

phylum:
@moollaza sounds good to me.

lib/DDG/Goodie/Calculator.pm
phylum: Do you want me to add that to my version or wait till the other dev responds?

moollaza:
:+1:

RomainLG:
I was first but small player :)

Merge this one!

lib/DDG/Goodie/Calculator.pm
phylum: @moollaza, when I added <code>use utf8;</code> it broke my code and when I commented it out it worked fine again.  I'm not sure why it broke my code. Any insight would be helpful unless you want me to leave the code out.  Thanks.

mintsoft: @phylum I would guess that the character that was physically in the file wasn't utf8. How exactly did it break the code?

phylum: @mintsoft all tests using Unicode symbols display, "Sorry, no hit on your instant answer" when <code>use utf8;</code> is added.  Using all other symbols, such as an asterick instead of the multiplication symbol, work as expected.  However, the unicode symbols are processed correctly when I leave out <code>use utf8;</code>.
I used <code>Crtl+Shft+U xxxx</code> to input the symbols.  Should I have added them differently?

lib/DDG/Goodie/Calculator.pm
```diff
@@ -166,7 +171,7 @@ sub spacing {
     my ($text, $space_for_parse) = @_;
 
     $text =~ s/\s{2,}/ /g;
-    $text =~ s/(\s*(?<!<)(?:[\+\-\^xX\*\/\%]|times|plus|minus|dividedby)+\s*)/ $1 /ig;
+    $text =~ s/(\s*(?<!<)(?:[\+\-\^xX×∙⋅\*\/÷\%]|times|plus|minus|dividedby)+\s*)/ $1 /ig;
```
moollaza: @phylum @mintsoft It seems like the problem is that you're using the actual unicode character in the regex instead of its unicode encoding:
http://stackoverflow.com/a/1573804/819937

https://duckduckgo.com/?q=unicode+%E2%88%99&t=ffcm&ia=answer

phylum: @moollaza should I put each character in it's Unicode encoding instead?

moollaza: @phylum I believe so. I'd suggest you test with one to confirm

phylum: @moollaza I tested the Unicode encoding and it was successful with <code>use utf8;</code>.  I'm changing over all the Unicode symbols to their Unicode encoding so it will work properly.

moollaza: Sweet! :+1:

moollaza:
Hmm I'm noticing that it's not working in DuckPAN Server but the tests do pass and it works in duckpan query. The failure is likely a DuckPAN utf-8 bug.

mintsoft:
@moollaza I've never had that problem with anything else using unicode (`Hiragana.pm` for example), is this an outstanding issue?

moollaza:
>I've never had that problem with anything else using unicode (Hiragana.pm for example), is this an outstanding issue?

I was pretty sure DuckPAN was handling UTF-8 correctly, but I too haven't had any recent problems with it. Still trying to figure out where the issue lies or why it's not working in this case.

moollaza:
@mintsoft think I found the problem :smile: 

https://github.com/duckduckgo/p5-app-duckpan/blob/b70aa74b72f1373f8bb6298065590f218ae1df29/lib/App/DuckPAN/Web.pm#L241

We're encoding the input to DDG::Request for DuckPAN Server but not DuckPAN Query.

moollaza:
@phylum @mintsoft 

Aaalright. After more digging I realized that this PR can revert to having the actual unicode chars in the regex and should have `use utf8`. In that state, it works correctly in `duckpan server`. However it doesn't in `duckpan query`.

I've discovered this is a duckpan bug and have made a PR to fix it.

@phylum sorry for the back and forth, but can you revert the most recent commit and make sure `use utf8` is there?

phylum:
@moollaza @mintsoft I reverted it back to the previous version and added <code>use utf8;</code>.

mintsoft:
@phylum looks like the tests are failing: https://travis-ci.org/duckduckgo/zeroclickinfo-goodies/builds/71588519#L3906

phylum:
@mintsoft I know the code doesn't work in <code>duckpan query</code> once I add <code>use utf8;</code> which is probably the reason it failed the tests.  @moollaza had me revert back to my original commit and add <code>use utf8;</code> since it still worked in <code>duckpan server</code>.  I'll be glad to make corrections if this is not a <code>duckpan</code> issue or a work around is needed.

mintsoft:
Cool; this LGTM @moollaza

moollaza:
@phylum @mintsoft the updated version of DuckPAN has been released so this should now work in both cases. Will double check and then merge :+1:

moollaza:
Works perfectly! Thanks again @phylum and @mintsoft

PR #1270, opened 2015-07-11
Fixing Tests in Travis for EmailValidator

mintsoft:
@moollaza @zachthompson this seems to fix the tests for EmailValidator; the stderr redirection was hiding the issue; however there's now a lot of noise in the output due to:
```
Use of uninitialized value in subroutine entry at /home/travis/perl5/perlbrew/perls/5.16/lib/site_perl/5.16.3/File/ShareDir.pm line 329.
```
It looks like autodeps wasn't pulling in `Net::Domain::TLD`; added it into EmailValidator.pm and voila.

I'm not that familiar with dzil, so I guess this isn't solving some root problem somewhere, however at least it shows up the problem!

Secondly, how can we resolve the ShareDir issue; it looks like problem is within the module itself.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,4 +16,4 @@ language: perl
 perl:
   - 5.16
 script:
-  - dzil smoke --release --author 2> /dev/null
+  - dzil smoke --release --author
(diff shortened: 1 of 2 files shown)


zachthompson:
Indeed, Email::Valid needs that module for certain functionality.  See https://github.com/duckduckgo/zeroclickinfo-goodies/issues/1276 for the warnings.  Thanks for looking into this @mintsoft.

PR #1275, opened 2015-07-11
Add Open Era to tennis cheat sheet

zachthompson:
The Open Era dividing line is a significant event in tennis history.


moollaza:
LGTM :+1:

Issue #1261, opened 2015-07-09
DaysBetween: Seems to have weird date parsing now


mintsoft:
![image](https://cloud.githubusercontent.com/assets/978048/8591756/54f1328a-2622-11e5-8d88-f391e83cd418.png)

![image](https://cloud.githubusercontent.com/assets/978048/8591772/6c18df08-2622-11e5-83e7-18150bf26d37.png)

Bujiraso:
I started to dig into this a bit and noticed that this is actually affecting another Goodie, WorkdaysBetween, in a similar way.
![ddg workdays](https://cloud.githubusercontent.com/assets/1975771/8635415/fb21ca58-27f7-11e5-9f92-1da31f28da8b.png)

What was surprising to me was that WeekdaysBetween is unaffected by the problem.
![ddg weekdays](https://cloud.githubusercontent.com/assets/1975771/8635418/0d0fda2a-27f8-11e5-94dc-4daec4ecccc4.png)

Still hacking at it a bit but I thought that was noteworthy.

Bujiraso:
The offending line appears to be:
`my ($date1, $date2) = parse_all_datestrings_to_date($1, $2);`
This line is present in both of the malfunctioning Goodies where WeekdaysBetween instead parses the dates individually like so:
`my ($start, $end) = (parse_datestring_to_date($1), parse_datestring_to_date($2));`

So it seems the Dates role may be overwriting the start date, so I'm gonna have a hack at that.
I've already added in a test into DaysBetween.t in order to catch relative date errors in future changes, but I'm a bit short on ideas for what to test in WorkdaysBetween, since it seems a more complicated case to drum up two relative dates with a fixed answer for -- if anyone has a suggestion that works, fire it my way and I'll add it in.

Bujiraso:
In the Dates role, it appears that the `parse_datestring_to_date` function isn't reacting properly when given a base. 
If I substitute out
````perl
my $date_object = ($dates_to_return[0]
                        ? parse_datestring_to_date($date, $dates_to_return[0])
                        : parse_datestring_to_date($date)
                    );
````

With the simpler line `my $date_object = parse_datestring_to_date($date);`, the problem is muted and the correct answer is returned for all of the above affected Instant Answers.
So it seems the case where a base is included as an argument for `parse_datestring_to_date` might be causing the issue.

Bujiraso:
[...]

PR #1277, opened 2015-07-12
Fixed Dates role basing relative dates like "today" on another date

Bujiraso:
The Dates role was re-using the base date given to it to calculate a new date, and thereby destroying the base date. I also added in a couple tests to hopefully catch bugs with this in the future.
This should fix some broken functionality reported in issue #1261.


mintsoft:
@Bujiraso thanks for the detailed exploration of this issue. I'm really surprised we didn't notice this sooner as it's basically completely broken. 

> Should the first date even be used as a base, at all? If I typed "June 10 to tomorrow", I don't expect it to reference the 10th, and common sense lets you know that you already told it the date distance (since tomorrow basically means "add one day", next week means "add seven days", and so forth).

So the behaviour around the relative date parsing was supposed to be to handle 'naked months' relative to each other (https://github.com/duckduckgo/zeroclickinfo-goodies/pull/972/files#diff-7360d2d81b9f3f7c544ff758a2985107R195)

That said, "tomorrow, today, yesterday" should *always* be tomorrow relative to the actual current time. If by implementing the relative base times I broke that then that was absolutely a mistake.

Bujiraso:
Ok, cool. The "yesterday", "today", and "tomorrow" are currently broken, as far as I could tell, so that needs some fixing.
For now, this PR could still be useful as it solves the related but separate issue of re-using the base date and thereby destroying it.

Bujiraso:
After thinking on your comment, my revised question is this: are there any cases where a base date would be used with a relative term? Should "Jan 22 to next month" be expected to say a relative number from Jan 22, like 9 (days till Feb 1) or 30/31 (days till Feb 22)? I would imagine the expected behaviour is "days between Jan 22 and the next month from today", but I know not everyone necessarily thinks the same.

Going with the case that my expected behaviour is correct, by changing the line I just sent a PR for (` my $tmp_date = $base_time->clone();`) in `parse_all_datestrings_to_date` to simply get the current time instead of using a base, this seems to fix all issues that I can come up with, including ones with absolute dates with relative dates, bare months, etc.

In that case where my expected behaviour is the correct one, this PR should likely not be merged, because I can send another PR with the better changes that I currently have. I'm not comfortable making the decision on my own since I'm too new, so let me know what you think!

Bujiraso:
Oh, I guess that push I made to my repo is now changing this PR...
New stuff I am learning!

mintsoft:
> are there any cases where a base date would be used with a relative term? 

Absolutely, to give you context to the changes around the date handling here. The original base date behaviour came out of: https://github.com/duckduckgo/zeroclickinfo-goodies/issues/971 andhttps://github.com/duckduckgo/zeroclickinfo-goodies/issues/969 resulting in this PR: https://github.com/duckduckgo/zeroclickinfo-goodies/pull/972

I think; the PR you've done results in the behaviour that I would expect/was agreed at the time although it might be worth seeing what @MrChrisW and @moollaza think

Bujiraso:
Ah, yes. I noticed this strange behaviour to jump to the next year when I made a change (not committed) so I did something different to preserve the more logical behaviour. In that case, this change should only be an improvement, since that feature is preserved.

t/DaysBetween.t
```diff
@@ -114,6 +114,22 @@ ddg_goodie_test(
...
     ),
+    'days between today and tomorrow' => test_zci(
+        qr/^There are 1 days between.+ and.+\.$/,
+        structured_answer => {
+            input     => '-ANY-',
+            operation => 'Days between',
+            result    => 1
+        },
+    ),
+    'days between last year and today' => test_zci(
```
moollaza: I actually assumed this would consider "last year" as "the first day of last year". I think most people would know that a year back from a point, or a month, is roughly 365 and 30 days respectively.

I checked to see how Wolfram Alpha handles this and they agree with my assumption:
https://www.wolframalpha.com/input/?i=days%20between%20last%20year%20and%20today

Bujiraso: Yes. I was thinking the same but didn't want to make big changes if someone had designed it to work like that originally. 
Should that feature improvement be added to this PR? 
I can always make another PR so it doesn't affect my personal workflow too much, but the "0 days between" bug will remain for a while until I can improve that handling.

moollaza: >Should that feature improvement be added to this PR?

Sure :+1: 

>I can always make another PR so it doesn't affect my personal workflow too much, 

Perfect, thanks a lot!

>but the "0 days between" bug will remain for a while until I can improve that handling.

I thought that's been fixed here?

Bujiraso: Oh I just meant everyone else won't see the fix until it's merged. No worries, though, I'll work on the improved handling.

Bujiraso:
I should make some note here of the changes I"ll be implementing so that I can be time efficient in making the correct changes.

Year is a settled discussion: if someone mentions a year (this, next, last), it should point to January 1st.

For months, also using Wolfram Alpha as a base, the first day of the month will be used. As an example, "this month" will refer to the first day of the current month, where "last month" refers to the first day of last month.

For weeks, Wolfram Alpha has a bit of a strange response to questions like "days from today until next week" 
https://www.wolframalpha.com/input/?i=days+between+today+to+next+week
Is there any input which makes most sense to guess for? The first day of the week is notoriously argued over, should we just make a decision and go for it, or should we keep somewhat useless 7 days answer?

For statements where the relative term's duration is greatly emphasized (e.g. between today and 2 months ago), should we keep the somewhat less useful answer? It seems suitable enough for me, and Wolfram Alpha does about as much
https://www.wolframalpha.com/input/?i=days+between+today+and+two+months+ago

mintsoft:
@Bujiraso IIRC `2 months ago`, was implemented as (for example 14th May when run on the 14th July), with round-down behavior so that that a month ago on the 31st of March is the 28th/29th Feb. I thought that'd already been sorted & tested, however I may be wrong!

Bujiraso:
Hm. Lines 354 and 367 of `00-roles.t` do not look right at all. Should this have never passed or am I reading completely wrong:
````PERL
2014-10-08T00:00:00Z' => {
                'a month ago'       => '01 Dec 1999',
````

mintsoft:
@Bujiraso You are absolutely right; it looks like I totally ballsed up the handling of relative dates here. That should never have passed at all.

Bujiraso:
@mintsoft haha, no worries!
I'm confused at how it managed to sneak by, but either way -- it's gone now! Many of the tests will have changed to fully test the modified relative handling.
I'm relatively new to both duckduckhack and PERL, are there any recommended docs to show me how to trace or debug variables while implementing this? For instance, in BASH, I'm quite fond of `bash -x` since it shows me everything that happens.

mintsoft:
@Bujiraso I've just changed that test in my branch of master and it passes fine. I think that block of tests is broken completely

mintsoft:
@Bujiraso AH! Actually stupidity on my behalf completely that test is never executed as it's replaced in the hash by the same key on line 367

Bujiraso:
@mintsoft  Oh, neat, that's how it works? Good to know, then!

mintsoft:
@Bujiraso Can you add these tests in: https://github.com/duckduckgo/zeroclickinfo-goodies/blob/c23a8027050b614392211f18e080ae345b52bfdf/t/DaysBetween.t#L129

Bujiraso:
I agree. I'll put in those tests, push, and then I'll work on the year thing for a new PR

mintsoft:
@Bujiraso I'm going to pull your changes into another branch to kick things around a bit. I might merge this in or I might get someone else to merge the new branch; I'll keep you posted.

mintsoft:
@Bujiraso I've included your changes in #1294 along with some tests. There's definitely some funky behaviour here.

Bujiraso:
Awesome : )
So I should close this PR if that one merges? Either way, I'll wait on cue for further action.

mintsoft:
I'll/the person who merges it will close it for you @Bujiraso if it gets merged

Bujiraso:
: 3 thanks

mintsoft:
@Bujiraso Sorry for the delay on this, I'm just waiting for https://github.com/duckduckgo/zeroclickinfo-goodies/pull/1294

Bujiraso:
@mintsoft no problem : )

PR #1283, opened 2015-07-13
Cheat Sheet Dark Theme Fix

sdougbrown:
Directly declaring colors in custom CSS is usually discouraged because it doesn't work great in dark themes - this just brings the cheat sheet css up to speed.

to @bsstoner 
cc @abeyang


moollaza:
Damn, that's my bad. Should have thought about that!

LGTM :+1:

bsstoner:
:+1:

abeyang:
This just changes the headers, right? If so, then :+1: from me, too!

PR #1284, opened 2015-07-13
ColorCodes: add support for queries like <color> to <type>

### LINKED ISSUE #1048: ColorCodes: Improve triggering, should support "#ccc to rgb"
https://duck.co/ideas/idea/5047/better-triggering-for-color-conversion-instant-ans

We should support queries of the form, "X to Y", for each of the various colour code formats:

- "#ccc to rgb"
- "#ccc to hsl"
- "#ccc to cmyk
- etc.

https://duck.co/ia/view/color_codes

mattr555:
Fixes #1048

Would like some extra eyes on this one, it's kind of hacky :eyes:

### DIFF
--- a/lib/DDG/Goodie/ColorCodes.pm
+++ b/lib/DDG/Goodie/ColorCodes.pm
@@ -45,6 +45,7 @@ triggers query_raw => qr/^
         \#?([0-9a-f]{6})|\#([0-9a-f]{3})                # handles #00f, #0000ff, etc
     )
     (?:(?:'?s)?\s+(inverse|negative|opposite))?
+    (?:\sto\s(?:$typestr))?
     $/ix;
 
 zci is_cached => 1;
@@ -61,7 +62,7 @@ attribution  cpan   => 'CRZEDPSYC',
              github => ['http://github.com/mintsoft', 'Rob Emery'];
 
 my %trigger_invert = map { $_ => 1 } (qw( inverse negative opposite ));
-my %trigger_filler = map { $_ => 1 } (qw( code ));
+my %trigger_filler = map { $_ => 1 } (qw( code k b ));
 
 my $color_mix = Color::Mix->new;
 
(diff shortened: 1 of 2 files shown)


moollaza:
@mintsoft care to review this one? I believe you've written most of the code here ;)

mintsoft:
I'll have a thorough look through this evening hopefully

lib/DDG/Goodie/ColorCodes.pm
```diff
@@ -61,7 +62,7 @@ attribution  cpan   => 'CRZEDPSYC',
              github => ['http://github.com/mintsoft', 'Rob Emery'];
 
 my %trigger_invert = map { $_ => 1 } (qw( inverse negative opposite ));
-my %trigger_filler = map { $_ => 1 } (qw( code ));
+my %trigger_filler = map { $_ => 1 } (qw( code k b ));
```
mintsoft: @mattr555 I'm not sure why this is needed here? Shouldn't the alterations to the regex and the `s//` further down be all that's needed?

mattr555: It seems like the regex is too greedy in some places. Take for example the query `#aaa to cmyk`.

I observed through Data::Dumper where the matches were found, and the relevant part of the regex for this query is actually (when you substitute `$typestr` in)

```perl
(.*?)(cmy|cmyk|cmyb|...)\s*:?\s*\(?\s*(.+?)\s*\)?
```

The matches are `"#aaa to "`, `"cmy"`, and `"k"`. This is why I need to include "k" (and "b") in the list of strings to ignore as "colors". The part I added to the regex is never actually used in this example. I guess this part of the regex is too greedy?

mintsoft: Hmm, I think if we switch the `.*` and `.+` to `[^\s]*` and `[^\s]+` it'll prevent the matches from chomping through so much?

mattr555:
Adding `[^\s]*` to the offending section allows us to get rid of the hacky part of restricting "k" and "b"

mintsoft:
Excellent, LGTM then

PR #1289, opened 2015-07-14
DatesRole: This test would fail if it was ever executed. 

mintsoft:
It shouldn't be here at all; it looks like one I left in when I was test driving it however it gets replaced in the test hash with a correct test anyway (https://github.com/duckduckgo/zeroclickinfo-goodies/pull/1289/files#diff-7360d2d81b9f3f7c544ff758a2985107L367).


mintsoft:
This was noticed by @Bujiraso in https://github.com/duckduckgo/zeroclickinfo-goodies/pull/1277 (thanks!)

moollaza:
LGTM :+1:

Issue #1282, opened 2015-07-13
Replace legacy "url escape" image with newer Cheat Sheet


moollaza:
We currently show an image in the results for the query [url escape](https://duckduckgo.com/?q=url+escape).

We should showcase this information with a cheat sheet instead. Let's ditch the image!

gokul1794:
@moollaza I made a cheatsheet for the url escape, but the `url escape` triggers the URLEncode IA as `url escape` as one of the triggers, What to do?

mintsoft:
@gokul1794 It's probably best to guard the `handle` on the urlencode to just `return` if the remainder matches `/cheat[\s]?sheet/`. If you open a PR with the cheatsheet on we can check out the triggering of them all together :)

gokul1794:
@mintsoft I will make a PR and then can you can help me out?

PR #1293, opened 2015-07-15
url escape cheatsheet

gokul1794:
In the process of fixing #1282 
Currently this will trigger the URL Encode IA as it has url escape as a trigger word.
//cc @mintsoft I'm a little confused as to what to do in the `handle remainder`

### DIFF
--- a/lib/DDG/Goodie/URLEncode.pm
+++ b/lib/DDG/Goodie/URLEncode.pm
@@ -27,7 +27,7 @@ attribution twitter =>      ['nshanmugham', 'Nishanth Shanmugham'],
 handle remainder => sub {
     my $in = $_;
 
-    return unless $in;
+    return unless $in; #/cheat[\s]?sheet/
 
     my $encoded_url = encodeURIComponent($in);
 
--- a/share/goodie/cheat_sheets/url-escape.json
+++ b/share/goodie/cheat_sheets/url-escape.json
@@ -1,0 +1,118 @@
+{
+    "id" : "url_escape_cheat_sheet",
+    "name" : "Url Escape",
+    "description" : "Escape Sequence",
+    "metadata" : {
+        "sourceName" : "ASCIIEncodingReference",
+        "sourceUrl" : "http://www.w3schools.com/tags/ref_urlencode.asp"
+    },
+    "section_order" : "Character",
+    "sections": {
+        "Character": [
+            {
+                "key": "Space",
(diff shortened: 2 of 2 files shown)


lib/DDG/Goodie/URLEncode.pm
```diff
@@ -27,7 +27,7 @@ attribution twitter =>      ['nshanmugham', 'Nishanth Shanmugham'],
 handle remainder => sub {
     my $in = $_;
 
-    return unless $in;
+    return unless $in; #/cheat[\s]?sheet/
```
moollaza: There's actually a lot more triggers than just `cheat sheet` -- I think we should instead make sure that the remainder is a valid URL. That would be a much better fix :)

moollaza: Actually, we don't really need to enforce proper URL's here. Hmm. We might need to stick with a regex that checks for Cheat Sheet triggers: 

https://github.com/duckduckgo/zeroclickinfo-goodies/blob/master/lib/DDG/Goodie/CheatSheets.pm#L11

But that list could grow...

moollaza: On second thought, I'm not sure we need this. I believe we can handle this on the backend. Will discuss with the team shortly...

gokul1794: Sure! @moollaza let me know if I need to do anything :)

moollaza:
@gokul1794 I just had a chat with the team and we agreed that there's probably a better way to represent this data both on the backend (JSON) and front (Cheat sheet UI). There are some similar instant answers we'd like to support better like the [html colors](https://duckduckgo.com/?q=html+colors) table and [ASCII table](https://duckduckgo.com/?q=ascii+codes)

For now I'm going to put this one on hold and speak to the team to figure out how we want to handle these kinds of "reference tables". It's likely that we'll end up with a new template of sorts!

gokul1794:
@moollaza sounds good to me :+1:

moollaza:
@gokul1794 a lot has since changed and we now have a "reference" template for cheat sheets. Please checkout the docs on cheat sheets to see how to use it. 

This PR will also need to pull in changes to master that have since been made. For example we've moved the cheat sheets to a new `/json` subdir inside `/share/goodie/

moollaza:
@gokul1794 ping! Just checking if you're still interested in finishing this up?

PR #1294, opened 2015-07-15
DateRole: Tests and fixes

mintsoft:
Includes changes by @Bujiraso (thanks!) from here:
https://github.com/duckduckgo/zeroclickinfo-goodies/pull/1277https://github.com/duckduckgo/zeroclickinfo-goodies/pull/1277

And tests within the role to confirm behaviour.


mintsoft:
I've also discovered a bug I think around the behaviour here: https://github.com/duckduckgo/zeroclickinfo-goodies/pull/1294/files#diff-7360d2d81b9f3f7c544ff758a2985107R207

The dates are what I expected when I wrote the test, however the role currently behaves differently; I'm interested to see what the concensus of other people about how this should behave is @mattr555 @MrChrisW @moollaza @jagtalon

jagtalon:
@mintsoft nice! What do you mean by "the role currently behaves differently"?

mintsoft:
@jagtalon So the ambiguity arises when `parse_all_datestrings` is used, the idea is that this should be given all the dates so that it can make decisions around DD/MM/YY vs MM/DD/YY etc.

Originally one of the features was when provided with 
`2015-01-01` and `september` that the output dates would be `2015-Jan-01` and `2015-Sep-01`; however this bug is that all dates were relative to the first date (so `today` would be `2015-Jan-01`).

So the question is, given the input `1990-06-13`  and `last december`, would you expect this to be parsed as:

* Current implemented: `1990-Jun-13` and `1989-Dec-01`
* What I expected: `1990-Jun-13` and `2000-Dec-01`

mintsoft:
@MrChrisW @jagtalon @moollaza ping!

mattr555:
> What I expected: 1990-Jun-13 and 2000-Dec-01

I'm confused as to why you would expect that, do you mean `2014-Dec-01`? It seems like the functionality of finding a relative date is useful, but not while finding many dates in batch. Maybe split this functionality out into a different function?

mintsoft:
@mattr555 sorry yes, I mean't that I expected `2014-Dec-01` for `last december`

mintsoft:
Actually, I think I'll deal with the "bug" in a separate PR. It'll give me a chance to play around with how it's used and make sure the behaviour I think is correct is actually sensible.

mintsoft:
@jagtalon @moollaza Any chance one of you can merge this in? :)

moollaza:
@mintsoft apologies for the delay, this LGTM :+1:

PR #1291, opened 2015-07-15
Emacs Cheat Sheet text wrapped in braces 

Sloff:
Fixes #1281 

**Before**
before

**After**
after

PR #1296, opened 2015-07-16
Emacs Keys wrapped and commas added for Consecutive Keys

### LINKED ISSUE #1281: Emacs CheatSheets: wrap text snippets in angle brackets; convert to code block
The Emacs cheat sheet has a few "keys" which include a key, then a text entry, and then another key:

`M-x query-replace RET`

This should be converted to:

`[M-x] [<query-replace>] [RET]`

The result should look like this:
emacs_help_at_duckduckgo

------

Sloff:
Replaces #1291 and fixes #1281 

**Before**
before

**After**
after

### DIFF
--- a/share/goodie/cheat_sheets/emacs.json
+++ b/share/goodie/cheat_sheets/emacs.json
@@ -29,13 +29,13 @@
 	}],
         "Quick Answers": [{
 	    "val": "find file",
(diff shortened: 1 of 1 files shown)


moollaza:
@Sloff LGTM :+1: 

Thanks a lot!

Issue #1286, opened 2015-07-14
TimezoneConverter: wrong MSK timezone


sjoubert:
The MSK (Moscow Time) is wrongly configured to UTC+4 while it should be UTC+3

This makes all conversion involving MSK to be off by one hour.
Example: https://duckduckgo.com/?q=0+MSK+in+UTC

See the following links for references:
https://en.wikipedia.org/wiki/Moscow_Time
http://www.timeanddate.com/time/zones/msk

moollaza:
@sjoubert thanks a lot for reporting!

I feel like this is a good reason to leverage the TimeAndDate API for thse kinds of queries rather than trying to maintain a static Goodie which can clearly become outdated!

PR #1305, opened 2015-07-19
Fix MSK and consolidate timezone lists

cwallen:
Fixes https://github.com/duckduckgo/zeroclickinfo-goodies/issues/1286 and since I noticed two lists of timezones in the repo, I made TimezoneConverter use the list from Dates.

### DIFF
--- a/lib/DDG/Goodie/TimezoneConverter.pm
+++ b/lib/DDG/Goodie/TimezoneConverter.pm
@@ -28,7 +28,7 @@ triggers any => qw(in into to);
 zci is_cached   => 1;
 zci answer_type => 'timezone_converter';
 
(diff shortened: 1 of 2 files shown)


mintsoft:
@cwallen Thanks for this! Instead of using the variable from the role directly can you add a `sub get_timezones` and use that instead?

mintsoft:
@cwallen This looks good to me, once I've seen the CI green I'll merge this in.

Related to this IA, it looks like it's one that could do with converting to a structured answer (https://github.com/duckduckgo/zeroclickinfo-goodies/issues/1163) if you're at all interested!

cwallen:
Sure, I'll give it a shot.

PR #1307, opened 2015-07-19
HTMLEntitiesEncode: Structured answer

mintsoft:
Notes: This is a weird IA to work on, it has single row output and guessed multi-line output, so I've used a basic structured_answer in the "one line" output case and the record template in the multi-line output.

Edge case queries to test:
* right angle brackets htmlencode
* how to encode an apostrophe in html ?
* htmlencode pound

New output (I really doubt many people actually see this record template):

image
image

Single lookup answer (much more likely for people to see I think!)
image

### DIFF
--- a/lib/DDG/Goodie/HTMLEntitiesEncode.pm
+++ b/lib/DDG/Goodie/HTMLEntitiesEncode.pm
@@ -5,6 +5,7 @@ use strict;
 use warnings;
 use DDG::Goodie;
 use Text::Trim;
+use HTML::Entities 'decode_entities';
 
 triggers any =>             'html', 'entity', 'htmlencode','encodehtml','htmlescape','escapehtml', 'htmlentity';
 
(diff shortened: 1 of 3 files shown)


moollaza:
@mintsoft thanks for this!

@abeyang any ideas on the design?

abeyang:
Looks decent. But with the list view, can we include a (small) title that says something like this:

> HTML Entity Encode for {search term}

So, an example:

> HTML Entity Encode for right angle brackets

mintsoft:
@moollaza I can't seem to find a way to add a title or something to the record template without creating a custom template. Have I missed something?

mintsoft:
image

@moollaza @abeyang

mintsoft:
related: https://github.com/duckduckgo/zeroclickinfo-goodies/issues/1163

abeyang:
@mintsoft looking good! Sorry, after seeing this the way you have it, can we lay it out with both title and subtitle, like this:

> **Pound**
> HTML Entity Encode

I think after that, it should be good to go!

mintsoft:
@abeyang we're here:

image

moollaza:
@mintsoft @abeyang the LGTM :+1: 

One thought though: the new version, for single results doesn't show what the rendered HTML entitty looks like. The previous result would should you the name of the entity and the actual symbol.

@abeyang Shall we render it in the title?

> &#36; | $
> HTML Entity Encode: Dollar

----

This is particularly useful because then I can copy the actual element from the Serp. I use it often for things like `em dash` which lets me grab the symbol: —

mintsoft:
@moollaza how about this: 
image

abeyang:
Looks good!

moollaza:
@mintsoft @abeyang the only thing I don't love about the parens is if the entity is a paren. I suppose we can write some additional logic to change the parens to something else in that case?

html___at_duckduckgo

Either way it's a minor edge case we can solve outside of this PR :+1:

PR #1314, opened 2015-07-21
Prime factors altered to use templates

Sloff:
Fixes part of #1163 

Before:
before1
before2
before3

After:
after1
after2
after3

Bug:
Is cached is ```Math::BigInt``` instead of ```0```, not sure why or if it matters
problem

### DIFF
--- a/lib/DDG/Goodie/PrimeFactors.pm
+++ b/lib/DDG/Goodie/PrimeFactors.pm
@@ -30,7 +30,7 @@ name 'PrimeFactors';
 topics 'math';
 category 'calculations';
 attribution github => [ 'austinheimark', 'Austin Heimark' ],
-                    ['https://github.com/Sloff', 'Sloff'];
+            github => ['https://github.com/Sloff', 'Sloff'];
 
 sub convert_to_superscripts (_) {
     my $string = $_[0];
--- a/lib/DDG/Goodie/PrimeFactors.pm
+++ b/lib/DDG/Goodie/PrimeFactors.pm
@@ -123,9 +123,10 @@ handle remainder => sub {
         @result = format_answer("$formatted is a prime number");
     } else {
         my ($text, $answer) = format_prime(@factors);
-        my $intro = "Prime factorization of $formatted";
+        my $subtitle = "$formatted - Prime Factors";
+        my $plaintext = "The prime factorization of $formatted is $text";
 
-        @result = format_answer($intro." is ".$text, $answer, $intro);
+        @result = format_answer($plaintext, $answer, $subtitle);
     }
     
     return @result;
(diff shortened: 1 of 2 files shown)


javathunderman:
The "is unsafe= 0," is alright, I see that a lot. @MrChrisW do you see anything wrong with this?

mintsoft:
@Sloff The `is_cached` is fine, it's just indicating that it's a Math::BigInt rather than a native int.

mintsoft:
I thought the `is unsafe` is for USFW content not html unsafe @moollaza ?

moollaza:
You're correct. It's a flag for IAs that are known to return adult content. Those IAs can only trigger when the user disables safe search.

lib/DDG/Goodie/PrimeFactors.pm
```diff
@@ -52,10 +60,10 @@ sub format_prime {
     my @html_result = ();
```
moollaza: I believe we can remove this now (and the related html code)

lib/DDG/Goodie/PrimeFactors.pm
```diff
@@ -71,6 +79,24 @@ sub commify {
...
+sub format_answer {
+    my ($plaintext, $title, $subtitle) = @_;
+    
+    return $plaintext,
+    structured_answer => {
+        id => 'prime_factors',
+        name => 'Answer',
+        data => {
+            title => $title || $plaintext,
+            subtitle => $subtitle
```
moollaza: Now that we're not dealing with proper sentences, lets' ditch the leading "The" on the subtitle.

moollaza:
@Sloff this is looking great.

@abeyang any thoughts? I suggested we drop "The " from the `subtitle` now that we're not giving an answer in sentence form.

Sloff:
Thanks @moollaza, PS changes made as suggested

abeyang:
@Sloff , looking good! I know we're not consistent with this yet, but I'd like our IAs to have this kind of format to the question, "What is the *type* of *specifics*?"

> **Answer**
> *Specifics* - *Type*

Thus, to the question, "What is the prime factor of 30?"

> **2 x 3 x 5**
> 30 - Prime Factor

Sloff:
Hi @abeyang, changes made as suggested:
after4

abeyang:
:+1:

mintsoft:
Thanks again @Sloff !

Sloff:
Pleasure!

moollaza:
@Sloff Thanks again for taking the time to contribute! We really appreciate it. Feel free to stick around and comment/help on other PR's & Issues or even submit more Instant Answers!

PR #1316, opened 2015-07-21
Chinese Zodiac Templates used and icons added

Sloff:
Fixes part of #1163 

Before:
before 1
before2

After:
after1
after2
after3
after4


moollaza:
@Sloff thanks a lot!

@abeyang thoughts on the design? I like the idea of the icons, but I'm not sure Wikipedia's icons work well here. Should we make our own?

abeyang:
@moollaza I agree -- we should make our own icons. We may even be able to utilize the elements, similar to the other Zodiac IA (v2).

@Sloff , while we work the icons out (thanks, btw!), can we make the following text changes:

> **羊 (Goat)**
> Feb 19, 2015 - Feb 7, 2016

And instead of "answer" can we input "Chinese Zodiac" in the tab bar?

abeyang:
Just for reference, this is what the new Zodiac will eventually look like:

zodiac-v1 4-capricorn earth

Sloff:
Hi @abeyang, changes made as suggested
after5

abeyang:
@Sloff Can we change the dates to use the format `Month Day, Year` ? Eg, `Feb 19, 2015`

Sloff:
Hi @abeyang, the date formats have been changed:
after6

abeyang:
Thanks @Sloff -- looks good! We're going to work on the icons internally, but for now, can you disable yours? That way we can ship this as-is, and add in the icons later. Thanks!

mintsoft:
@abeyang we were originally trying to keep the date output formats consistent between each IA by using the role (hence the 01 Jan 2014 format)

Sloff:
Hi all, I removed the icons, let me know about the date and I'll change it if needed.

cc @mintsoft @abeyang

Sloff:
I'm going to close and reopen this

PR #1322, opened 2015-07-23
Alias added to cheat sheets

Sloff:
This alters CheatSheet.pm to allow aliases to be assigned to cheat sheets, for example "microsoft windows cheat sheet" and "windows cheat sheet" will show the same cheat sheet:

### DIFF
--- a/lib/DDG/Goodie/CheatSheets.pm
+++ b/lib/DDG/Goodie/CheatSheets.pm
@@ -21,10 +21,43 @@ triggers startend => (
     'key bindings', 'keys', 'default keys'
 );
 
+# Accepts the json cheatsheet file name and an array of everything
+# that should be aliased with the json file
+sub build_aliases {
+    my ($json, @aliases) = @_;
+    my @result = ();
+    
+    foreach (@aliases) {
+        push (@result, $_);
+        push (@result, $json);
+    }
+    
+    return @result;
+}
+
+# The hash of all aliases
+my %alias = (
+    build_aliases('microsoft-windows.json', (
+        'windows', 'windows xp', 'windows vista', 'windows 7', 'windows 8',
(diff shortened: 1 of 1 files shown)


moollaza:
@Sloff thanks for this. I think the better approach would be to go through each JSON file, checking for an "aliases" property and if it exists, add it to a hash that maps the triggers to the JSON filename.

That way on initialization, we can build a hash of all the triggers that map to each IA.

@zachthompson do you have any thoughts on this? Is there a better approach? Perhaps have DZIL build a JSON/YAML file during the build process so that the Goodie doesn't have to do the work?

Or is there a metadata approach we can leverage here?

Sloff:
Sounds good @moollaza, I'll wait for @zachthompson's response before altering

zachthompson:
@Sloff Thanks for the idea.  We can already do aliases with [symlinks](https://github.com/duckduckgo/zeroclickinfo-goodies/blob/master/share/goodie/cheat_sheets/regular-expression.json), e.g. "ln -s cheat_sheet.json alias.json".  It's not ideal but works for now. We're also investigating moving cheat sheets to a [Longtail](https://github.com/duckduckgo/zeroclickinfo-longtail/pull/19) and adding [keywords](https://github.com/duckduckgo/zeroclickinfo-longtail/pull/19/files#diff-e1e8f712d7dbbe4ba8b1b023a7e45e73R5) to the json to allow for just this thing.  If that doesn't work we'd prefer to add the keywords to the json and pull those in somehow.  In any event,  we want to make these as non-technical as possible so updating the module with aliases would be a last resort.

moollaza:
@Sloff I had a chat with @zachthompson and we agreed that if you'd like to implement this functionality, it would be a net benefit (though we're investigating alternative ways of triggering Cheat Sheets)

So if you're itching to make it easier to trigger these I think we should aim for something along the lines of my approach which is to pre-process each JSON file and build up a trigger hash in `CheatSheets.pm`

PR #1323, opened 2015-07-23
Sublime cheatsheet keys altered

Sloff:
This just updates the keys in the sublime cheatsheet to the newest standard


moollaza:
@Sloff thanks! LGTM :+1:

PR #1326, opened 2015-07-23
Alphabetical Morse Code

ankushg07:
It gives out morse code for each alphabet and digit

[Morse Code](http://morsecode.scphillips.com/morse2.html)

It provides basic morse codes for each alphabet

programmers, encoders, security officials

YES

Yes, there exists an instant answer which convert the given string into morse code and so it does for this cheat sheet

Morse Code

-----
IA Page: https://duck.co/ia/view/morse_code_cheat_sheet


mintsoft:
@ankushg07 Thanks for this; looks useful. I'm just thinking that it'd be better to get more characters on screen by default.

moollaza:
>Thanks for this; looks useful. I'm just thinking that it'd be better to get more characters on screen by default.

@mintsoft we plan to allow different layouts and initial views for cheat sheets. For now the 6 items is OK, but I agree it doesn't always look great.

share/goodie/cheat_sheets/alphabetical-morse-code.json
```diff
@@ -0,0 +1,125 @@
+{
+    "id": "alphabetical_morse_code_cheat_sheet",
```
moollaza: Please change to "morse_code_cheat_sheet"

share/goodie/cheat_sheets/alphabetical-morse-code.json
```diff
@@ -0,0 +1,125 @@
...
+    "name": "Alphabetical Morse Code Cheat Sheet",
+    "description": "Morse Code help",
```
moollaza: Let's remove this description; It's not adding much value.

ankushg07:
@moollaza well there is a conflict  on name more_code_cheat sheet, as there exists an instant answer which converts the string "Morse code cheat sheet " into the morse code of "cheat sheet"  ie. [See This ](https://duckduckgo.com/?q=morse+code+cheat+sheet&ia=answer)
Hence I could not find any alternative besides changing the name

share/goodie/cheat_sheets/alphabetical-morse-code.json
ankushg07: @moollaza  Changes done!

ankushg07:
@moollaza @mintsoft any other issue with this cheat sheet??

mintsoft:
@ankushg07 The morse code IA can be changed so that it doesn't trigger when the phrase to be 'morse-coded' is "cheatsheet" here: https://github.com/duckduckgo/zeroclickinfo-goodies/blob/master/lib/DDG/Goodie/Morse.pm#L28

ankushg07:
@moollaza @mintsoft fixed the name conflict between IA and Cheat sheet.. works perfectly now !!

lib/DDG/Goodie/Morse.pm
```diff
@@ -26,7 +26,7 @@ handle remainder => sub {
     my $input = shift;
 
     return unless $input;
-
+    return if($input eq 'cheat sheet');
```
mintsoft: Cool; can you also add a test into `t/Moorse.t` that proves that Morse returns undef for this?

ankushg07: @mintsoft sorry but i dont know how to do that..could you please guide me?

mintsoft:
@ankushg07 Yeah sure.
Basically we want to add a test so that if someone goes into the `Morse.pm` and fiddles around with it then they don't accidentally break the change you've made.

So if we add something like:
```perl
'morse code cheatsheet' => undef,
'morse code cheat sheet' => undef,
```
Into the `Morse.t` file then it'll fail the tests if someone accidentally removes the functionality you've added!

ankushg07:
@mintsoft thank you for help ..done with that ..

mintsoft:
@ankushg07 actually it looks like the Morse code still fires: https://travis-ci.org/duckduckgo/zeroclickinfo-goodies/builds/73369231

ankushg07:
@mintsoft it seems all checks have been passed now ..!!

lib/DDG/Goodie/Morse.pm
```diff
@@ -26,7 +26,8 @@ handle remainder => sub {
...
+    return if($input eq 'cheatsheet');
```
mintsoft: you could do `return if($input ~= /cheat[\s]?sheet/);` or `return if($input eq 'cheat sheet' || $input eq 'cheatsheet');`

mintsoft:
@moollaza this LGTM

lib/DDG/Goodie/Morse.pm
moollaza: Hmm, there's actually many other triggers, like "help" for example or "reference". We're exploring multi-Goodie triggering, but for now we may have to give cheat sheets a different priority.

moollaza: Whoops, I forgot this change was included in this PR! I'll post a task to reign in the triggering of this IA because it seems a little too easy to trigger.

PR #1328, opened 2015-07-24
Binary Logic altered to use templates

Sloff:
Fixes part of #1163 

Before:
before 1
before 2

After:
after1
after2

### DIFF
--- a/lib/DDG/Goodie/BinaryLogic.pm
+++ b/lib/DDG/Goodie/BinaryLogic.pm
@@ -122,10 +122,19 @@ handle query_raw => sub {
     return if not defined $value_ref;
 
     my $text_output = ${$value_ref};
-    my $html_output = "<div>Result: <b>" . ${$value_ref} . "</b></div>";
-    my $heading = "Binary Logic";
 
-    return answer => $text_output, html => $html_output, heading => $heading;
+    return $text_output, structured_answer => {
+        id => 'binary_logic',
+        name => 'Answer',
+        data => {
+            title => $text_output,
+            subtitle => 'Binary Logic'
+        },
+        templates => {
+            group => 'text',
+            moreAt => 0
+        }
+    }
 };
 
 1;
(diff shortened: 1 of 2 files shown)


moollaza:
@abeyang any thoughts on the subtitle?

Wondering if we should have a `more at` link to a Wikipedia article on Binary Logic?

abeyang:
@moollaza is [boolean algebra](https://en.wikipedia.org/wiki/Boolean_algebra) a better description for this? /cc @Sloff 

Also, I think it would be better to actually show the binary numbers. So, if the query is "4 or 5", the result would be:

> **1**
> Boolean Algebra: 100 XOR 101

mintsoft:
@abeyang technically no; boolean algebra is the rearrangement/use (or whatever) of logical operations (true/false) whilst this is doing bitwise computation.

moollaza:
@Sloff @mintsoft @abeyang 

Let's describe it as "Bitwise Operation" then? https://en.wikipedia.org/wiki/Bitwise_operation

moollaza:
@Sloff once the subtitle is updated accordingly this should be good!

@abeyang I think if we show the binary version of the nubmer, then we should still show the original input?

>1
>Bitwise Operation: 100 (decimal 4) XOR 101 (decimal 5)

mintsoft:
@moollaza Sounds good

Sloff:
@moollaza @mintsoft @abeyang 

Hi all, changes made to the subtitle and also to the triggers to fire on `4 ⊕ 5` and `4⊕5`:
after1
after2
after3

One thing I'm not sure how to handle, an answer of `0` displays as follows:
bug1

abeyang:
@Sloff Overall, looks good! Just need to figure out how to get that "0" in there, and I think we're set!

lib/DDG/Goodie/BinaryLogic.pm
```diff
@@ -122,10 +141,19 @@ handle query_raw => sub {
     return if not defined $value_ref;
 
     my $text_output = ${$value_ref};
```
mintsoft: @Sloff if you switch the text output to always be a string, i.e. to this:
```perl
my $text_output = "${$value_ref}";
```
Then the "0" output will be displayed correctly.

Sloff: Thanks @mintsoft!

mintsoft:
@moollaza community + design sign off complete!

moollaza:
@mintsoft I'm on a train right now but you're welcome to merge!

PR #1330, opened 2015-07-24
Perimeter altered to use templates

Sloff:
Before:
before1
before2

After:
after1
after2

### DIFF
--- a/lib/DDG/Goodie/Perimeter.pm
+++ b/lib/DDG/Goodie/Perimeter.pm
@@ -24,20 +24,24 @@ handle query_lc => sub {
         my %polygons = ("pentagon" => 5, "hexagon" => 6, "octagon" => 8);
         my $shape = $1 ? "circle" : $3 || $5 || $8;
 
-        my $subtitle = "Perimeter";
+        my $subtitle = "Perimeter of $shape with sides of ";
         my $answer;
 
         if ($shape eq "square") {
+            $subtitle .= $4;
             $answer = $4 * 4;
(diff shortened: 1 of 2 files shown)


moollaza:
@Sloff thanks again. Code LGTM :+1: 

@abeyang any thoughts on design?

abeyang:
@Sloff @moollaza I think subtitle has to be slightly more descriptive, such as:

> Perimeter of hexagon with sides of 45

And

> Circumference of circle with radius of 1

A little verbose, but it's better to be clear. For example, I wasn't sure whether the circumference example was diameter or radius.

Sloff:
Hi @abeyang, changes made as suggested:

after3
after4

lib/DDG/Goodie/Perimeter.pm
```diff
@@ -24,26 +24,40 @@ handle query_lc => sub {
         my %polygons = ("pentagon" => 5, "hexagon" => 6, "octagon" => 8);
         my $shape = $1 ? "circle" : $3 || $5 || $8;
 
-        my $answerPrefix = "Perimeter of $shape: ";
+        my $subtitle = "Perimeter of $shape with sides of ";
         my $answer;
 
         if ($shape eq "square") {
+            $subtitle .= $4;
```
moollaza: It's not really necessary, but using named captures in the regex would make this much clearer. Referring to the capture groups by number isn't very clear

moollaza:
@Sloff this LGTM :+1: 
Thanks a lot!

PR #1332, opened 2015-07-25
Chinese Zodiac Templates used and icons added

Sloff:
Replaces #1316


mintsoft:
@abeyang are you happy to have the date format switched back to the one provided by the role for now? I'll make sure it's handled correctly once I've done the location aware role

moollaza:
@Sloff thanks a lot! In the future, please attach some snapshots so we can quickly glance at the changes.

abeyang:
:+1:

PR #1335, opened 2015-07-26
Factorial of a number

codenirvana:
Find the factorial of a given number

It calculate the factorial of a number so that user don't have to use any other 3rd party tool/website.

No

* 7 factorial
* 5 fact

Maths

Yes

None

Yes, I want to trigger the IA even if use search like **7!** but its not working. Maybe because of !bang.

img1

img2

## Checklist

```
- [X] Verified that instant answer adheres to code styleguide (https://duck.co/duckduckhack/code_styleguide)
- [X] Tested cross-browser compatibility

    - Windows 8
        - [X] Google Chrome
        - [] Firefox
        - [] Opera
        - [] IE 10

    - Mac OSX
        - [X] Google Chrome
        - [] Firefox
        - [] Opera
        - [X] Safari
```
-----
IA Page: https://duck.co/ia/view/factorial

### DIFF
--- a/lib/DDG/Goodie/Factorial.pm
+++ b/lib/DDG/Goodie/Factorial.pm
@@ -25,10 +25,7 @@ triggers any => "factorial", "fact";
 handle remainder => sub {
     my $n = $_;
     
-    #accept only numeric input
-    if ($n =~ m/[^0-9]/){
-        return;
-    }
+    return unless $n =~ /^\d+$/;
     
     $n = int($n);
     
--- a/lib/DDG/Goodie/Factorial.pm
+++ b/lib/DDG/Goodie/Factorial.pm
@@ -1,0 +1,54 @@
+package DDG::Goodie::Factorial;
+# ABSTRACT: n factorial
+
+use strict;
+use DDG::Goodie;
+
+use List::Util qw(reduce);
+use bignum;
+use Scalar::Util qw(looks_like_number);
+
+zci answer_type => "factorial";
+zci is_cached   => 1;
+
+name "Factorial";
+description "Returns Factorial of n";
+primary_example_queries "factorial 7", "7 factorial";
+secondary_example_queries "12 fact";
+topics 'math';
+category 'calculations';
+code_url "https://github.com/duckduckgo/zeroclickinfo-goodies/blob/master/lib/DDG/Goodie/Factorial.pm";
+attribution github => ["https://github.com/codenirvana", "Udit Vasu"],
+            twitter => "uditdistro";
+
+triggers any => "factorial", "fact";
+
+handle remainder => sub {
+    my $n = $_;
+    return if(looks_like_number($n) eq false);
+    
+    #check if given number is negative
+    my $isNegative = 0;
+    if( $n =~/^\-/ ){
+        $n = $n*(-1);
+        $isNegative = 1;
+    }
+    
+    my $fact = reduce { $a * $b } 1 .. $n;
+    
+    #if given number is negative then
+    if($isNegative){
+        $fact = '-'.$fact;
+        $n = '-'.$n;
+    }
+    
+    return "Factorial of $n is $fact",
+      structured_answer => {
+        input     => [$n],
+        operation => 'Factorial',
+        result    => $fact
+    };
+
+};
+
+1;
--- a/lib/DDG/Goodie/Factorial.pm
+++ b/lib/DDG/Goodie/Factorial.pm
@@ -32,7 +32,7 @@ handle remainder => sub {
     
     $n = int($n);
     
-    my $fact = 1;
+    my $fact = '1';
     $fact = reduce { $a * $b } 1 .. $n if ($n > 0);
     
     return "Factorial of $n is $fact",
--- a/lib/DDG/Goodie/Factorial.pm
+++ b/lib/DDG/Goodie/Factorial.pm
@@ -5,8 +5,7 @@ use strict;
 use DDG::Goodie;
 
 use List::Util qw(reduce);
-use bignum;
-use Scalar::Util qw(looks_like_number);
+use bigint;
 
 zci answer_type => "factorial";
 zci is_cached   => 1;
@@ -25,22 +24,13 @@ triggers any => "factorial", "fact";
 
 handle remainder => sub {
     my $n = $_;
-    return if(looks_like_number($n) eq false);
     
-    #check if given number is negative
-    my $isNegative = 0;
-    if( $n =~/^\-/ ){
-        $n = $n*(-1);
-        $isNegative = 1;
+    #accept only numeric input
+    if ($n =~ m/[^0-9]/){
+        return;
     }
     
     my $fact = reduce { $a * $b } 1 .. $n;
-    
-    #if given number is negative then
-    if($isNegative){
-        $fact = '-'.$fact;
-        $n = '-'.$n;
-    }
     
     return "Factorial of $n is $fact",
       structured_answer => {
--- a/lib/DDG/Goodie/Factorial.pm
+++ b/lib/DDG/Goodie/Factorial.pm
@@ -32,7 +32,8 @@ handle remainder => sub {
     
     $n = int($n);
     
-    my $fact = reduce { $a * $b } 1 .. $n;
+    my $fact = 1;
+    $fact = reduce { $a * $b } 1 .. $n if ($n > 0);
     
     return "Factorial of $n is $fact",
       structured_answer => {
--- a/lib/DDG/Goodie/Factorial.pm
+++ b/lib/DDG/Goodie/Factorial.pm
@@ -30,6 +30,8 @@ handle remainder => sub {
         return;
     }
     
+    $n = int($n);
+    
     my $fact = reduce { $a * $b } 1 .. $n;
     
     return "Factorial of $n is $fact",
--- a/t/Factorial.t
+++ b/t/Factorial.t
@@ -4,7 +4,7 @@ use strict;
 use warnings;
 use Test::More;
 use DDG::Test::Goodie;
-use BigInt;
+use bigint;
 
 zci answer_type => "factorial";
 zci is_cached   => 1;
--- a/t/Factorial.t
+++ b/t/Factorial.t
@@ -12,7 +12,7 @@ zci is_cached   => 1;
 ddg_goodie_test(
     [qw( DDG::Goodie::Factorial )],
     'fact 4' => test_zci(
-        'Factorial of 4 is 24.',
+        'Factorial of 4 is 24',
         structured_answer => {
             input     => ['4'],
             operation => 'Factorial',
@@ -20,7 +20,7 @@ ddg_goodie_test(
         }
     ),
     'factorial 6' => test_zci(
-        'Factorial of 6 is 720.',
+        'Factorial of 6 is 720',
         structured_answer => {
             input     => ['6'],
             operation => 'Factorial',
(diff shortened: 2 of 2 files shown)


lib/DDG/Goodie/Factorial.pm
```diff
@@ -0,0 +1,54 @@
...
+    my $isNegative = 0;
+    if( $n =~/^\-/ ){
+        $n = $n*(-1);
+        $isNegative = 1;
+    }
+    
+    my $fact = reduce { $a * $b } 1 .. $n;
+    
+    #if given number is negative then
+    if($isNegative){
```
mintsoft: Hmm this isn't right, factorial isn't defined for `n < 0`. Personally I'd return undef

lib/DDG/Goodie/Factorial.pm
```diff
@@ -0,0 +1,54 @@
...
+category 'calculations';
+code_url "https://github.com/duckduckgo/zeroclickinfo-goodies/blob/master/lib/DDG/Goodie/Factorial.pm";
+attribution github => ["https://github.com/codenirvana", "Udit Vasu"],
+            twitter => "uditdistro";
+
+triggers any => "factorial", "fact";
+
+handle remainder => sub {
+    my $n = $_;
+    return if(looks_like_number($n) eq false);
```
mintsoft: You could do `return unless(looks_like_number($n));`

lib/DDG/Goodie/Factorial.pm
```diff
@@ -0,0 +1,54 @@
+package DDG::Goodie::Factorial;
+# ABSTRACT: n factorial
+
+use strict;
+use DDG::Goodie;
+
+use List::Util qw(reduce);
+use bignum;
```
mintsoft: Probably better off with bigint; as factorial is only defined for integers

lib/DDG/Goodie/Factorial.pm
mintsoft: Also could do with protection from non-integers, as `looks_like_number` covers all numbers

codenirvana:
@mintsoft Its my first time using Perl, Thanks for pointing out the errors!

lib/DDG/Goodie/Factorial.pm
```diff
@@ -0,0 +1,44 @@
...
+            twitter => "uditdistro";
+
+triggers any => "factorial", "fact";
+
+handle remainder => sub {
+    my $n = $_;
+    
+    #accept only numeric input
+    if ($n =~ m/[^0-9]/){
+        return;
```
codenirvana: @mintsoft is it correct? or should I ```return undef```?

mintsoft: It'd  work, however I'd `return undef;`

t/Factorial.t
```diff
@@ -0,0 +1,33 @@
...
+    ),
+    'factorial 6' => test_zci(
+        'Factorial of 6 is 720.',
+        structured_answer => {
+            input     => ['6'],
+            operation => 'Factorial',
+            result    => 720
+        }
+    ),
+    'tell a factorial'    => undef,
```
mintsoft: I'd add some tests for non-integers and negative numbers that show it returns `undef`

lib/DDG/Goodie/Factorial.pm
codenirvana: @mintsoft I tried ```return undef;``` but then it shows a blank Answer!

t/Factorial.t
codenirvana: @mintsoft added some tests.

lib/DDG/Goodie/Factorial.pm
mintsoft: @codenirvana Sorry yes, you're absolutely right it should just be `return;` I have no idea what I was thinking :)

codenirvana:
@mintsoft added few more tests.

mintsoft:
@codenirvana Hmm; as `0!` = `1` by definition, you'll probably have to do something like:
```perl
my $fact = 1;
$fact = reduce { $a * $b } 1 .. $n if ($n > 0);
```

t/Factorial.t
mintsoft: Can you add one for `0!` too?

codenirvana:
@mintsoft tried many methods but still getting the same error.
Check the Structured Answer status:
```
internals: {
        answer              "Factorial of 0 is 1",
        answer_type         "factorial",
        caller              "DDG::Goodie::Factorial",
        is_cached           Math::BigInt,
        is_unsafe           0,
        structured_answer   {
            input       [
                [0] 0
            ],
            operation   "Factorial",
            result      Math::BigInt
        }
    }                 
```

The result should be **1** not ```Math::BigInt ```

mintsoft:
@codenirvana If you add `use BigInt;` in the test then it'll pass :)

codenirvana:
@mintsoft Still same problem!
no idea what's wrong..

mintsoft:
@codenirvana Sorry I meant `use bigint;` (i.e. the same as is in the `.pl`)

codenirvana:
@mintsoft I did that too, not working!
But its working now :) This is what I did" ```my $fact = '1';```
I think this is because of bigint. We need to assign value in quotes.
Did some tests and everything is working fine now.

mintsoft:
@codenirvana Looks like the tests are failing because there's a `.` on the end of the output that the tests aren't expecting https://travis-ci.org/duckduckgo/zeroclickinfo-goodies/builds/73113462#L4020

lib/DDG/Goodie/Factorial.pm
moollaza: Might be a little more straightforward to `return` unless we're only dealing with numbers?

```perl
return unless $n =~ m/\d+/;
```

moollaza:
Here's a pic of the failing tests:
build__4316_-_duckduckgo_zeroclickinfo-goodies_-_travis_ci

FYI @codenirvana , you can run the tests locally as well by running `duckpan test` or `prove -Ilib t/testName.t`

codenirvana:
@mintsoft @moollaza Fixed the test file errors!

lib/DDG/Goodie/Factorial.pm
mintsoft: @codenirvana Actually I'm with moollaza now I've thought about it, I'd use the `return unless $n =~ m/\d+/;` it's cleaner IMO

codenirvana: @mintsoft I think its not good to use ```$n =~ m/\d+/``` because its giving error for non-numeric inputs like '5a', 'n12' etc.

So what to do?
```if ($n =~ m/[^0-9]/){
        return;
    }``` Is working perfectly fine!

Comment (moollaza) [inline lib/DDG/Goodie/Factorial.pm]: @codenirvana whoop, we need to anchor my regex, and then it should work fine: `/^\d+$/`

Comment (moollaza) [inline lib/DDG/Goodie/Factorial.pm]: I guess `return if n =~ m/[^0-9]/` is just as good. I just prefer the `return unless` approach. We tend to use that quite a bit on the IA code

**Comment by codenirvana** [comment]
@moollaza Updated condition...
```return unless $n =~ /^\d+$/;``` working perfectly fine!

File: lib/DDG/Goodie/Factorial.pm [hunk 6]
```diff
@@ -0,0 +1,44 @@
...
+description "Returns Factorial of n";
+primary_example_queries "factorial 7", "7 factorial";
+secondary_example_queries "12 fact";
+topics 'math';
+category 'calculations';
+code_url "https://github.com/duckduckgo/zeroclickinfo-goodies/blob/master/lib/DDG/Goodie/Factorial.pm";
+attribution github => ["https://github.com/codenirvana", "Udit Vasu"],
+            twitter => "uditdistro";
+
+triggers any => "factorial", "fact";
```
moollaza: Later, I think we should add "calculate" as a trigger and check if the remainder matches `\d+!`. The currently interferes with some !bangs though so we'll need to wait.

moollaza:
@codenirvana this is looking great! I have on final suggestion, that we change "inf" to "Infinity" as the result.

@abeyang this one is ready for review. It's pretty straightforward :smile:

codenirvana:
@moollaza anything else to fix from my side?

abeyang:
Looks good! :+1:

moollaza:
@codenirvana I had a chat with the team about this IA -- is there any reason the existing calculator IA wasn't modified to add this functionality? It seems like that would be a more powerful approach because we'd be able to support any equation with a factorial.

lib/DDG/Goodie/Factorial.pm
```diff
@@ -0,0 +1,44 @@
...
+
+handle remainder => sub {
+    my $n = $_;
+    
+    return unless $n =~ /^\d+$/;
+    
+    $n = int($n);
```
moollaza: We should probably put a sane upper-bound on acceptable inputs. i.e `fact 1239123123` should just return "Infinity" rather then wasting cycles (and memory) computing it. I was able to produce an "out of memory" warning when I threw some big numbers at it. 

It looks like anything larger than `170` returns a result of Infinity right now.

codenirvana:
@moollaza Good idea to have factorial function in calculator IA... but what if user just search **15 factorial** will it still works?
If Yes then tell me where to make changes to add this factorial function?

lib/DDG/Goodie/Factorial.pm
moollaza: Also, we might as well `return if $num < 0` rather than waste any time on it :+1:

moollaza:
@codenirvana ideally it should work that way! Checkout [Calculator.pm](https://github.com/duckduckgo/zeroclickinfo-goodies/blob/master/lib/DDG/Goodie/Calculator.pm)

moollaza:
@codenirvana I just realized that `Math::BigInt` has it's own `bfac()` method which calculates factorials. I think it would be best to use that instead as it's a stronger implementation which seems to handle integers up to 7000. This seems to be the function it uses: https://metacpan.org/source/PJACKLAM/Math-BigInt-1.9997/lib/Math/BigInt/Calc.pm#L1614

mintsoft:
@moollaza I think the only issue with putting it into Calculator will be that we're `eval`ing the result; so we've got to be careful with the input into arbitrary (especially non-core!) functions like this

moollaza:
@mintsoft I was thinking we could look for `\d+!` in the query and replace it with the value, then pass it on to `eval()`?

moollaza:
I'm not 100% sure if that would work if we received a massive number though. Separately, myself and @zachthompson are investigating `Math::GMP` right now. It looks like if we switch to that there will be some potentially larger performance gains.

mintsoft:
@moollaza that's a good point, I forgot it's only got to work with integers

codenirvana:
@moollaza when I tried ```bfac()``` its giving some errors...
Have a look at the structured_answer:
 ```result      Math::BigInt``` is wrong right?

```
answer              "Factorial of 12 is 479001600", 
        answer_type         "factorial", 
        caller              "DDG::Goodie::Factorial",
        is_cached           Math::BigInt, 
        is_unsafe           0,
structured_answer   {
            input       [
                [0] 12
            ],
            operation   "Factorial",
            result      Math::BigInt
        }   
```

codenirvana:
@moollaza  checked Calculator.pm and looks very complex to me!
I am new to perl and my changes may broke the complete IA, so I think its good for me to stay away from that file :smile:

moollaza:
@codenirvana okay. I'll merge this for now, and open an issue for someone to port add functionality to the Calculator Goodie, and to handle the calculation using Math::BigInt

PR #1338, opened 2015-07-27
IRC Cheat Sheet

namandistro:
screenshot from 2015-07-27 18 27 24

-----
IA Page: https://duck.co/ia/view/irc_cheat_sheet

### DIFF
--- a/share/goodie/cheat_sheets/irc.json
+++ b/share/goodie/cheat_sheets/irc.json
@@ -12,82 +12,82 @@
     ],
     "sections": {
         "Main": [{
-            "key":"$server serverurl",
+            "key":"/server serverurl",
             "val":"Joining a Server"
           }, {  
-            "key":"$join #Channel",
+            "key":"/join #Channel",
             "val":"Joining a Channel"
           }, {  
-            "key":"$away <reason>",
+            "key":"/away <reason>",
             "val":"Setting yourself away"
           }, {  
-            "key":"$msg NickServ register PASSWORD email",
+            "key":"/msg NickServ register PASSWORD email",
             "val":"Registering your Name"
           }, {  
-            "key":"$nick NAME",
+            "key":"/nick NAME",
             "val":"Changing Your Name"
           }, {  
-            "key":"$msg NickServ identify PASSWORD",
+            "key":"/msg NickServ identify PASSWORD",
             "val":"Identifing to nickserv"
           },
              {  
-            "key":"$msg NickServ set password PASSWORD",
+            "key":"/msg NickServ set password PASSWORD",
             "val":"Changing your password (You have to repeat this for all of your linked nicks, which can be found with msg NickServ info YOUR NICK)"
           }, 
              {
-              "key":"$topic NEW TOPIC",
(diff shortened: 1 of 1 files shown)


mintsoft:
@namandistro Thanks for this! Sorry it's taken a while to get to.

One thing that'st just stood out to me at first glance is that I thought all commands were prefixed with `/` not `$`. I.e `/nick mintsoft` would change my nick not `$nick mintsoft` ?

namandistro:
@mintsoft All those command are the terminal one and i saw the git command starts with $ so i thought that symbol will make sense rather than this /

mintsoft:
@namandistro Ah, I see what the confusion is here. The `$` on the git ones is used to indicate that it's a shell command and that it should be executed as such (this is a fairly common thing within the *nix sphere). 

When you say "terminal ones" do you mean within a specific IRC client?

namandistro:
@mintsoft This command varies from client to client some gives us god gui some ask directly the channel name i made a cheat sheet to give all the command we can use from our shell to access it... So this are the basic commands you need to use for IRC from telnet a short video where you can understand what i am talking about https://www.youtube.com/watch?v=oZ81n_0M63g&noredirect=1

mintsoft:
AH I see, I didn't realise it was essentially telnet. Other IRC clients I've used always require `/` infront of  commands. Either way, I'm not sure if `$` is right for this. 

@moollaza @jagtalon @MrChrisW @mattr555  Any opinions?

namandistro:
Ya, But that thing depends on client to client but as default all this command will be used at backend...

MrChrisW:
@mintsoft I tend to agree with using the forward-slash format.

@namandistro I understand what you're saying, the **forward slash** standard is based on the **client-side** and not directly implemented nor enforced by the IRC protocol.

However its become the standard of almost all IRC clients (GUI and CLI), using the `/` to interpret a command and send a raw command to the IRC server. 

You'll also find this (forward slash) mentioned on any other help guides online.

+ http://www.ircbeginner.com/ircinfo/ircc-commands.html

+ http://web.cortland.edu/flteach/methods/obj2/irc-cmds.html

Therefore I think its appropriate to use `/` here as oppose to `$`. 

On a side note, I don't think the people using telnet or a raw socket connection to an IRC server would necessarily need a CheatSheet :smile:

mintsoft:
I suppose a compromise would be to not have anything at the start of each command, I.e. `nick mint`?

namandistro:
So @MrChrisW and @mintsoft What is the final recommendation is hould use $ or /. ?

mintsoft:
@namandistro I think `/`

namandistro:
@mintsoft i made the recommended changes

moollaza:
LGTM :+1:

moollaza:
@namandistro Thanks again for taking the time to contribute! We really appreciate it. Feel free to stick around and comment/help on other PR's & Issues or even submit more Instant Answers!

moollaza:
@namandistro @sagarhani congrats! It's live: https://duckduckgo.com/?q=irc+cheat+sheet&t=ffab&ia=cheatsheet

PR #1349, opened 2015-08-01
CheatSheets: Handle queries of form 'X for Y'

moollaza:
I just realized we don't support queries like "Cheat sheet for Firefox" or "key bindings for vim".

This adds that functionality.

Also. we need triggering tests for these!

/cc @zachthompson

### DIFF
--- a/lib/DDG/Goodie/CheatSheets.pm
+++ b/lib/DDG/Goodie/CheatSheets.pm
@@ -21,16 +21,27 @@ triggers startend => (
     'key bindings', 'keys', 'default keys'
 );
 
-handle remainder => sub {
+sub get_json_path {
+    my $name = lc shift;
+    return share(join('-', split /\s+/o, "$name.json"));
+}
 
-    # allow queries like "cheat sheet for vim"
-    s/\bfor\b// if $_ =~ m/^for .+/;
-    trim($_);
+handle remainder => sub {
 
     # If needed we could jump through a few more hoops to check
     # terms against file names.
-    my $json_path = share(join('-', split /\s+/o, lc($_) . '.json'));
-    open my $fh, $json_path or return;
+    my $json_path = get_json_path($_);
+    my $fh;
+
+    # try to open file, may need to cleanup candidate filename
+    unless (open $fh, $json_path){ 
+        # allow queries like "cheat sheet for vim"
+        s/^for\s+//;
+        trim($_);
+        $json_path = get_json_path($_);
+        open $fh, $json_path or return;
+    }
+
     my $json = do { local $/;  <$fh> };
     my $data = decode_json($json);
 
--- a/lib/DDG/Goodie/CheatSheets.pm
+++ b/lib/DDG/Goodie/CheatSheets.pm
@@ -3,7 +3,7 @@ package DDG::Goodie::CheatSheets;
 
 use JSON::XS;
 use DDG::Goodie;
-use DDP;
+use Text::Trim;
 
 zci answer_type => 'cheat_sheet';
 zci is_cached   => 1;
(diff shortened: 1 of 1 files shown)


lib/DDG/Goodie/CheatSheets.pm
```diff
@@ -22,6 +22,11 @@ triggers startend => (
 );
 
 handle remainder => sub {
+
+    # allow queries like "cheat sheet for vim"
+    s/\bfor\b// if $_ =~ m/^for .+/;
```
moollaza: this is a little naïve. My other idea was to simply add a whole new list of `start` triggers for most of the existing triggers.

zachthompson: No need to check then sub.  If the sub pattern fails, nothing happens:

    s/^for\s+//;

Of course, if any cheat sheets start with "for ", well, they're in a bad way.  Maybe we should have these types of subs after the first open fails.  That gives cheat sheets a chance to match with for before removing.

lib/DDG/Goodie/CheatSheets.pm
```diff
@@ -22,6 +22,11 @@ triggers startend => (
...
+    trim($_);
```
zachthompson: s/^\s*(.+?)\s*$/$1/;

mintsoft: We've used `Text::Trim` in most other Goodies, is there any particular reason you prefer a regex?

zachthompson: Less code, fewer dependencies, for something that could be, and is under the hood, handled by a regex.  But if we're already using the mod, no point.

lib/DDG/Goodie/CheatSheets.pm
moollaza: >No need to check then sub. If the sub pattern fails, nothing happens:

:+1:

moollaza:
@zachthompson alright this should be a little better now. Any thoughts?

lib/DDG/Goodie/CheatSheets.pm
```diff
@@ -21,11 +21,27 @@ triggers startend => (
...
     # terms against file names.
-    my $json_path = share(join('-', split /\s+/o, lc($_) . '.json'));
-    open my $fh, $json_path or return;
+    my $json_path = get_json_path($_);
+    my $fh;
+
+    # try to open file, may need to cleanup candidate filename
+    unless (open $fh, $json_path){ 
+        # allow queries like "cheat sheet for vim"
+        s/^for\s+//;
```
zachthompson: Now that I see this new sub, we could have done:
```Perl
$json_path =~ s/^for-//;
```
No need to do that twice.  The trim should be the first sub executed as well so the first file test gets the same benefit.

moollaza: :+1: Will fix. Thanks!

moollaza:
@zachthompson I think this is good now. Only generating the path once. Trimming it and lowercasing the name once as well.

lib/DDG/Goodie/CheatSheets.pm
```diff
@@ -21,11 +21,25 @@ triggers startend => (
...
     # terms against file names.
-    my $json_path = share(join('-', split /\s+/o, lc($_) . '.json'));
-    open my $fh, $json_path or return;
+    my $json_path = get_json_path($_);
+    my $fh;
+
+    # try to open file, may need to cleanup candidate filename
+    unless (open $fh, $json_path){ 
+        # allow queries like "cheat sheet for vim"
+        $json_path =~ s|/for-|/|;
```
mintsoft: Wouldn't this be better as `\bfor-` ?

zachthompson: get_json_path is unnecessary overhead.  Just get rid of it.   We don't want to call that repeatedly probably ever.  So if we need to do multiple subs in the future we should just be copying $json_path to a new var.

PR #1354, opened 2015-08-02
Fixed translations, grammar, spelling and typos in the cheat sheet for German.

Retzudo:
Fixed translations, grammar, spelling and typos in the cheat sheet for German.


mintsoft:
@Retzudo Thanks for this, it looks like the original was rather incorrect!

PR #1367, opened 2015-08-06
CheatSheets: removed plus sign between consecutive keys as per most recent syntax rules

moollaza:
Cmd used: `find . -type f -exec sed -i 's!\] + \[!] [!g' {} \;`

## Results

Blender:
screen shot 2015-08-05 at 22 55 59

DuckDuckGo:
screen shot 2015-08-05 at 22 59 31

Gimp
screen shot 2015-08-05 at 23 00 08

Gnome
screen shot 2015-08-05 at 23 03 33

Mac:
screen shot 2015-08-05 at 23 09 44

Vim:
screen shot 2015-08-05 at 23 10 41

to @zachthompson 
/cc @jagtalon

### DIFF
--- a/share/goodie/cheat_sheets/blender.json
+++ b/share/goodie/cheat_sheets/blender.json
@@ -37,7 +37,7 @@
             "key": "Mouse wheel"
         }, {
             "val": "Add object",
-            "key": "[Shift] + [a]"
+            "key": "[Shift] [a]"
         }, {
             "val": "Delete",
             "key": "x"
@@ -52,13 +52,13 @@
             "key": "n"
         }, {
             "val": "Save file",
-            "key": "[Ctrl] + [s]"
+            "key": "[Ctrl] [s]"
         }, {
             "val": "Render image",
             "key": "F12"
         }, {
             "val": "Render animation",
-            "key": "[Ctrl] + [F12]"
+            "key": "[Ctrl] [F12]"
         }, {
             "val": "Stop render",
             "key": "Esc"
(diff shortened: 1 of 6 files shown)


share/goodie/cheat_sheets/blender.json
```diff
@@ -223,56 +223,56 @@
             "key": "v"
         }, {
             "val": "Weight paint mode",
-            "key": "[Ctrl] + [Tab]"
+            "key": "[Ctrl] [Tab]"
         }, {
             "val": "Switching workspace",
-            "key": "[Ctrl] + [Left arrow] or [Ctrl] + [Right arrow]"
+            "key": "[Ctrl] ([←]/[→])"
```
zachthompson: This looks a bit weird. The parens almost make it look optional.  Space around the "/"?

share/goodie/cheat_sheets/gnome-keyboard.json
```diff
@@ -10,66 +10,69 @@
     "sections": {
         "Desktop Shortcuts": [{
             "val": "Switch between the Activities overview and desktop",
-            "key": "[Alt] + [F1] or [Super]"
+            "key": "[Alt] [F1]"
```
zachthompson: "Super" isn't right or...?

share/goodie/cheat_sheets/gnome-keyboard.json
```diff
@@ -10,66 +10,69 @@
...
+        },{
+            "val": "Switch between the Activities overview and desktop (alternative)",
+            "key": "Super"
```
zachthompson: No brackets?

share/goodie/cheat_sheets/mac-keyboard.json
```diff
@@ -9,148 +9,148 @@
...
         }, {
-            "key" : "[Cmd] + [Opt] + [V]",
+            "key" : "[Cmd] [Opt] [V]",
             "val" : "Move the copied files to current directory"
         }, {
-            "key" : "[Cmd] + [Del]",
+            "key" : "[Cmd] [Del]",
             "val" : "Delete selected folder or file"
         }, {
             "key" : "Return/Enter",
```
zachthompson: Brackets?

share/goodie/cheat_sheets/mac-keyboard.json
```diff
@@ -9,148 +9,148 @@
...
         }, {
-            "key" : "[Cmd] + [N]",
+            "key" : "[Cmd] [N]",
             "val" : "Open new Finder window"
         }, {
-            "key" : "[Ctrl] + [Tab]",
+            "key" : "[Ctrl] [Tab]",
             "val" : "Cycle through tabs"
         }, {
             "key" : "Spacebar",
```
zachthompson: Brackets?

share/goodie/cheat_sheets/mac-keyboard.json
```diff
@@ -9,148 +9,148 @@
...
             "val" : "Preview selected item using Quick Look"
         }], "Spotlight": [{
-            "key" : "[Cmd] + [Space]",
+            "key" : "[Cmd] [Space]",
             "val" : "Open Spotlight"
         }, {
             "key" : "Enter",
```
zachthompson: Ditto.

zachthompson: I guess it doesn't matter if there is only one value, right?

share/goodie/cheat_sheets/blender.json
moollaza: This is actually the syntax we (Abe and I) agreed on, and we reference it in the docs. We need the parens so it's clear that the alternates are the arrow keys and not the entire left side of the / vs the right.

share/goodie/cheat_sheets/gnome-keyboard.json
moollaza: Alternative key combos should be separate entries, so I've moved it below and added "(alternate)"

share/goodie/cheat_sheets/gnome-keyboard.json
moollaza: Single keys (any string without a space) get wrapped in a code block automatically. It seems this has confused a few people and I'm contemplating removing it. I did this when looking at the regex IA otherwise all its entries would need to be escaped but that may make more sense.

share/goodie/cheat_sheets/mac-keyboard.json
moollaza: yup, currently the brackets would be handled literally and end up in the code block

share/goodie/cheat_sheets/blender.json
zachthompson: Cool, I get the parens. What about spaces around the "/" to accentuate the alternates?

moollaza: We haven't been doing that elsewhere, but if you have a strong opinion on it, we can pass it by Abe. We'd need to updated existing IA's and the docs to follow suite.

The parser *doesn't* ignore white space currently, so alternatively we could update it to do that, so that the legibility would be better in the JSON but the output will still look correct.

zachthompson: No strong opinion really, just seems a bit inconsistent with how we do it elsewhere.

PR #1371, opened 2015-08-06
bash cheat sheet

synbit:
# Goodie Cheat Sheet Pull Request Template

Displays a bash cheat sheet.

1. whatever I've learnt so far 2. O'REILLY book (Shell Scripting by Arnold Robbins & Nelson H. F. Beebe) 3. https://www.gnu.org/software/bash/ 4. Various other little sources on the web which I don't remember

1. The easiest thing to start with 2. Nice book 3. Because it's the official site I suppose 4. For completeness

Maybe, I don't know. Please let me know if you do know.

I'm expecting Linux users to find this useful, especially new users (I wish I knew this a few years back).

I searched after I started working on this , and yes there is apparently:
https://duck.co/ideas/idea/5983/display-bash-keyboard-shortcuts

Basically, I need feedback because I am a little concerned with the size of this cheat sheet. Also, I am not sure if the categories I put in are correct/appropriate. I'd like to know what I've done wrong please.

DuckDuckHack was mentioned to me by Rob (one of your community leaders) a few months ago.

bashcheatsheet

bashcheatsheetshowmore

-----
IA Page: https://duck.co/ia/view/bash_cheat_sheet


share/goodie/cheat_sheets/bash.json
```diff
@@ -0,0 +1,513 @@
...
+            {
+                "key": "M-→",
+                "val": "Move forward by a word"
+            },
+            {
+                "key": "M-f",
+                "val": "Move forward by a word (alternative)"
+            },
+            {
+                "key": "C-xx",
```
mintsoft: Should this be `C-x, C-x`?

share/goodie/cheat_sheets/bash.json
```diff
@@ -0,0 +1,513 @@
...
+        "Here Documents",
+        "Auto-Completion",
+        "Backslash Escape Sequences",
+        "Misc"
+    ],
+    "sections": {
+        "Lingo": [
+            {
+                "key": "C",
+                "val": "Cotrol Key (Ctrl)"
```
mintsoft: I think this should be `Control key`

share/goodie/cheat_sheets/bash.json
```diff
@@ -0,0 +1,513 @@
...
+            {
+                "key": "C-p",
+                "val": "Get previous command"
+            },
+            {
+                "key": "C-n",
+                "val": "Get next command"
+            },
+            {
+                "key": "[C-r] [<pattern>]",
```
mintsoft: All the key shortcuts should be in square brackets like these are, i.e. `[C-n]` for the Get next command

share/goodie/cheat_sheets/bash.json
```diff
@@ -0,0 +1,513 @@
...
+            {
+                "key": "\"$*\"",
+                "val": "Command line arguments for current process, expanded to a single argument"
+            },
+            {
+                "key": "$0",
+                "val": "The name of the shell program"
+            },
+            {
+                "key": "$<N>",
```
mintsoft: It might be worth putting things to do with "nth arguments of the last command" and "Last argument of last command" together etc

share/goodie/cheat_sheets/bash.json
```diff
@@ -0,0 +1,513 @@
...
+            },
+            {
+                "key": "||",
+                "val": "Logical OR"
+            }
+        ],
+        "IO Redirection": [
+            {
+                "key": "cmd > file",
+                "val": "Send cmd output to file"
```
mintsoft: It's probably worth making it explicit that this truncates the `file` before adding the content in; maybe `truncate file and send cmd output to file` or `replace file content with cmd output`

share/goodie/cheat_sheets/bash.json
```diff
@@ -0,0 +1,513 @@
...
+            },
+            {
+                "key": "cmd >> file",
+                "val": "Append cmd output to file"
+            },
+            {
+                "key": "cmd < file",
+                "val": "Contents of file become arguments to cmd"
```
mintsoft: This should be `content of file is passed to cmd's stdin`

share/goodie/cheat_sheets/bash.json
```diff
@@ -0,0 +1,513 @@
...
+            },
+            {
+                "key": "cmd1 | cmd2",
+                "val": "Output of cmd1 is passed to cmd2"
```
mintsoft: I'd probably be more explicit and say `is passed to stdin of cmd2`

share/goodie/cheat_sheets/bash.json
```diff
@@ -0,0 +1,513 @@
...
+            }
+        ],
+        "Expansion And Other Operators": [
+            {
+                "key": "${var:-val}",
+                "val": "If $var exists and isn't NULL, return its value; otherwise return val"
```
mintsoft: I didn't think that bash had a concept of `NULL`?

share/goodie/cheat_sheets/bash.json
```diff
@@ -0,0 +1,513 @@
...
+                "key": "${var:?}",
+                "val": "If $var exists and isn't NULL, return its value; otherwise return 'var: parameter null or not set' and abord"
+            },
+            {
+                "key": "${var:+val}",
+                "val": "If $var exists and isn't NULL, return val; otherwise return NULL"
+            },
+            {
+                "key": "$((...))",
+                "val": "Arithmetic Expansion"
```
mintsoft: I'd probably clarify this with an explanation that arithmetic can be performed within the `$(())`

share/goodie/cheat_sheets/bash.json
synbit: That's right, fixed.

share/goodie/cheat_sheets/bash.json
```diff
@@ -0,0 +1,513 @@
...
+                "val": "String length operator"
+            }
+        ],
+        "Exit Statuses": [
+            {
+                "key": "0",
+                "val": "Success"
+            },
+            {
+                "key": ">0",
```
mintsoft: This should be `!=0` as `-1` can be used on some systems

share/goodie/cheat_sheets/bash.json
```diff
@@ -0,0 +1,513 @@
...
+                "val": "Here doc start (newline must follow)"
+            },
+            {
+                "key": "EOF",
+                "val": "Here doc end (must be at the beginning of the line)"
+            }
+        ],
+        "Auto-Completion": [
+            {
+                "key": "TAB",
```
mintsoft: `[TAB]`

share/goodie/cheat_sheets/bash.json
```diff
@@ -0,0 +1,513 @@
...
+            },
+            {
+                "key": "\\uHHHH",
+                "val": "Byte with Hex value HHHH (unicode)"
+            }
+        ],
+        "Misc": [
+            {
+                "key": "$(cmd)",
+                "val": "Execute cmd"
```
mintsoft: Just an empty line with `cmd` on it would execute cmd; doesn't this execute `cmd` and capture the output so it can be placed into a variable? i.e. it would normally be used as: `a=$(cmd)`

share/goodie/cheat_sheets/bash.json
```diff
@@ -0,0 +1,513 @@
...
+            },
+            {
+                "key": "`cmd`",
```
mintsoft: Same again

share/goodie/cheat_sheets/bash.json
synbit: "Control Key" or "Ctrl"?

mintsoft: Either, but definitely not `Cotrol`

share/goodie/cheat_sheets/bash.json
synbit: From the guide:
"Brackets, [ ], or braces, { }, are used to wrap key combinations in code blocks. For convenience, if you include no brackets or braces, the entire string will be shown in a code block."

I read that as "omitting them is just fine". I have seen examples which use square brackets as well as examples that don't. What I don't like about what I've done is that I have used inconsistently. Just let me know what is preferred.

share/goodie/cheat_sheets/bash.json
synbit: That's a good point, I'll look into this.

share/goodie/cheat_sheets/bash.json
synbit: fixed

share/goodie/cheat_sheets/bash.json
synbit: agree, fixed

share/goodie/cheat_sheets/bash.json
synbit: Open a terminal and type:
echo ${var:?}

This prints:
bash: var: parameter null or not set

I cannot analyse exactly what the difference between "not set" and "null" is, but this makes me believe that the "null" concept exists.

share/goodie/cheat_sheets/bash.json
synbit: Not sure I would, the description is pretty accurate. Unless people prefer to replace $((..)) with an example. So, make the key $((3+4)) and the value 7 or something along these lines. 

Thoughts?

share/goodie/cheat_sheets/bash.json
synbit: Is that the "all the shortcuts should have square brackets" problem? If so, see comment further up. Let me know and I'll change it accordingly.

share/goodie/cheat_sheets/bash.json
moollaza: Let's just use "Ctrl"

share/goodie/cheat_sheets/bash.json
synbit: That was very confusing. Changed it!

share/goodie/cheat_sheets/bash.json
synbit: Fixed

share/goodie/cheat_sheets/bash.json
moollaza: @mintsoft actually single keys (or "key"'s without spaces in them) don't need to be wrapped in `[]` -- I'm starting to think that may be confusing to some, but it seemed useful because there are a lot of single "word" keys and also cases where wrapping it in `[]` isn't mean to create a key, like the regex cheat sheet: `[:ALPHA:]`, etc otherwise all these `[]` would have to be escaped.

Ultimately I think forcing all `[]` to be escaped makes the most sense, but I'm worried it's going to create a lot of extra work for devs. In this case the about `${..}` would have to be `$\\{..\\}`.

Any thoughts on this?

share/goodie/cheat_sheets/bash.json
```diff
@@ -0,0 +1,513 @@
+{
+    "id": "bash_cheat_sheet",
+    "name": "bash",
+    "description": "Unix Shell",
```
moollaza: Can we make this a little more descriptive?

share/goodie/cheat_sheets/bash.json
synbit: fixed

share/goodie/cheat_sheets/bash.json
synbit: fixed

share/goodie/cheat_sheets/bash.json
mintsoft: To me, expansion means something different from execution; It doesn't "expand" arithmetical expressions, it computes them.

share/goodie/cheat_sheets/bash.json
mintsoft: @moollaza thanks for the clarification! It's definitely confused me; I think personally that escaping things is relatively intuitive whereas slightly inconsistent behaviour is more confusing.
That said, maybe we could do with a different representation for keys which doesn't 'clash' with more commonly used characters?

share/goodie/cheat_sheets/bash.json
mintsoft: Right I see, I think I've usually used "empty" to describe variables in those situations rather than `null` however it makes perfect sense

share/goodie/cheat_sheets/bash.json
```diff
@@ -0,0 +1,513 @@
...
+                "key": "${var:-val}",
+                "val": "If $var exists and isn't NULL, return its value; otherwise return val"
+            },
+            {
+                "key": "${var:=val}",
+                "val": "If $var exists and isn't NULL, return its value; otherwise set it to val and then return its value"
+            },
+            {
+                "key": "${var:?val}",
+                "val": "If $var exists and isn't NULL, return its value; otherwise return 'var: val' and abord"
```
mintsoft: abord?

share/goodie/cheat_sheets/bash.json
mintsoft: As @moollaza has clarified, it's not needed for single 'keys' to display correctly

share/goodie/cheat_sheets/bash.json
mintsoft: Bourne Again Shell?

share/goodie/cheat_sheets/bash.json
synbit: http://www.tldp.org/LDP/abs/html/arithexp.html

?

share/goodie/cheat_sheets/bash.json
synbit: "All you'll ever need"? I'm referring to Bash, not the cheat sheet, obviously... :)

Anyway, I like "Bourne Again Shell".

share/goodie/cheat_sheets/bash.json
synbit: I suggest the key to be changed to:
${var:?text}

As is, seems useless, but ${var:?bla and bla} will actually print "bla and bla" if $var is not set.

Thoughts?

synbit: btw when I say abord, I mean terminate execution and return 1 (aka fail).

share/goodie/cheat_sheets/bash.json
mintsoft: Fair enough if that's what they call it!

share/goodie/cheat_sheets/bash.json
mintsoft: The markup is obviously less important than the appearance, basically you only need [] around some sections if you want to be able to output multiple `code` blocks within the key.

synbit: I made a few changes regarding this. I think it looks OK as per \3ce2083\

share/goodie/cheat_sheets/bash.json
synbit: As I said above, it looks OK to me as per 3ce20837d9fb56fa32d5bd5050f303bbabed6ac0

Let me know what you think.

share/goodie/cheat_sheets/bash.json
synbit: Changed! see 714a562ee8ae7c205b3d3064f2ec0ce8635d2f2a

share/goodie/cheat_sheets/bash.json
synbit: The order seems sensible to me as per aa69dff98aa882cb023682a07d4baaccd3e88da0

share/goodie/cheat_sheets/bash.json
synbit: That makes sense to me. Fixed. Also added more stuff [1] in eb2ae15321844841d13fc36d69bac21419185ce4

[1] http://www.tldp.org/LDP/abs/html/exitcodes.html

share/goodie/cheat_sheets/bash.json
```diff
@@ -0,0 +1,565 @@
...
+                "key": "cmd >> file",
+                "val": "Send cmd output to file (append)"
+            },
+            {
+                "key": "cmd < file",
+                "val": "Contents of file are passed to cmd's stdin"
+            },
+            {
+                "key": "cmd << '<string>'",
+                "val": "Contents of the shell script up to a line identical to <string> become stdin for cmd (verbatim)"
```
mintsoft: I don't understand what this actually means?
* 'the shell script' => which shell script? 
* should it be used with `<` and `>` in it or is that being used as a placeholder?

share/goodie/cheat_sheets/bash.json
```diff
@@ -0,0 +1,565 @@
...
+                "key": "cmd << <string>",
+                "val": "Same as above but variables and arithmetic operations are processed"
+            },
+            {
+                "key": "cmd1 | cmd2",
+                "val": "Output of cmd1 is passed to cmd2's stdin"
+            },
+            {
+                "key": "cmd >&n",
+                "val": "Send cmd's output to file descriptor n"
```
mintsoft: I'd say `stdout` rather than output, as stderr is technically output

share/goodie/cheat_sheets/bash.json
```diff
@@ -0,0 +1,565 @@
...
+                "key": "cmd m>&n",
+                "val": "Same as above except that output that would normally go to file descriptor m, will also go to n"
+            },
+            {
+                "key": "cmd <&n",
+                "val": "File descriptor n becomes stdin for cmd"
+            },
+            {
+                "key": "cmd m<&n",
+                "val": "Same as above except that intput that would normally come from file descriptor m, will now come from n"
```
mintsoft: intput?

share/goodie/cheat_sheets/bash.json
```diff
@@ -0,0 +1,565 @@
...
+                "key": "0",
+                "val": "Success"
+            },
+            {
+                "key": "!=0",
+                "val": "Failure"
+            },
+            {
+                "key": "1",
+                "val": "Catchall fo rgeneral errors"
```
mintsoft: `fo regeneral` ?

share/goodie/cheat_sheets/bash.json
```diff
@@ -0,0 +1,565 @@
...
+            {
+                "key": "128+n",
+                "val": "Fatal error signal \"n\""
+            },
+            {
+                "key": "130",
+                "val": "Script terminated by Ctrl-C"
+            },
+            {
+                "key": "255*",
```
mintsoft: What's the `*` to indicate?

share/goodie/cheat_sheets/bash.json
```diff
@@ -0,0 +1,565 @@
...
+                "key": "a=$(cmd)",
+                "val": "$a now holds cmd's output"
+            },
+            {
+                "key": "a=`cmd`",
+                "val": "$a now holds cmd's output (alternative)"
+            },
+            {
+                "key": "cmd &",
+                "val": "Execute cmd in the background"
```
mintsoft: It might be worth adding the word `fork` into this as it's the term that's generally used

share/goodie/cheat_sheets/bash.json
synbit: What that means is that 

```
cmd <<'something'
...
something
```

is included within a script. That is the only way this makes sense.

share/goodie/cheat_sheets/bash.json
synbit: Agree, fixed.

share/goodie/cheat_sheets/bash.json
synbit: Replaced with stdin

share/goodie/cheat_sheets/bash.json
synbit: Thanks! Fixed!

share/goodie/cheat_sheets/bash.json
synbit: Added

share/goodie/cheat_sheets/bash.json
synbit: My only guess after reading the documentation is that they simply mean >255, but I left it like this because if it means something else, I wouldn't want to change their intentions.

What they say is that if a program exits like this:

```
...
exit 3809;
```

what will happen is this 3809%256. Any objections if I make it >255 instead?

mintsoft: I think it might be better to take it out completely as the `!=0` covers general failures. With regards to > 255 values, that's system dependent. Some *nix's have unsigned bytes for return codes (0->255), others have signed bytes (-128->127); some use signed 16 bit ints etc

share/goodie/cheat_sheets/bash.json
mintsoft: According to the manpage it's a here-doc being piped to stdin of `cmd`:

> Here Documents
>     
>       This type of redirection instructs the shell to  read  input  from  the
>       current source until a line containing only delimiter (with no trailing
>       blanks) is seen.  All of the lines read up to that point are then  used
>       as the standard input for a command.
>
>       The format of here-documents is:
>
>              <<[-]word
>                      here-document
>              delimiter
>
>       No  parameter  and variable expansion, command substitution, arithmetic
>       expansion, or pathname expansion is performed on word.  If any  charac-
>       ters  in  word are quoted, the delimiter is the result of quote removal
>       on word, and the lines in the here-document are not expanded.  If  word
>       is  unquoted, all lines of the here-document are subjected to parameter
>       expansion, command substitution, and arithmetic expansion, the  charac-
>       ter  sequence  \<newline>  is  ignored, and \ must be used to quote the
>       characters \, $, and `.
>
>       If the redirection operator is <<-, then all leading tab characters are
>       stripped  from  input  lines  and  the line containing delimiter.  This
>       allows here-documents within shell scripts to be indented in a  natural
>       fashion.

share/goodie/cheat_sheets/bash.json
synbit: Yes, by this logic some *nix's have no bash and you need you need to pipe into perl in order to do a loop. I agree with you, this is not a strict standard. However, 50% of what is presented by this cheat sheet is not standard. Yet, most of it (if not all) seems to tally with what is seen in RHEL based systems and Debian based systems (the 127+n status for instance). As the reference says, this is not a standard but it's a recommendation in an attempt to follow C/C++ standards, as the author says.

In some cases it's going to be correct, in some cases it's going to be wrong. Personally, since this is a reference I would keep it in, that's why I included it. If things change we can always update the cheat sheet. 

Thoughts?

share/goodie/cheat_sheets/bash.json
synbit: Any ideas on this? I am not convinced it's clear but I'm out of ideas. How can it be described in a better way?

synbit: Yes it's basically a here doc. Which makes me thing that changes should be made to the _Here Document_ section. Either delete it and include extra comments on this, or just delete the dummy ":" from the beginning...

share/goodie/cheat_sheets/bash.json
mintsoft: Well, that's a very different situation; the lack of a userland shell and differing return codes (kernel compilation option) are totally different beasts. It's fairly likely that the return codes could be compiled differently due to the architecture of the chips (ARM vs i386 for example) which is fairly widely used by Raspberry Pi users which will also have bash.

Most ANSI implementations only guarantee a single byte. I'd still just bomb it out; it has basically no value being there however it can be incorrect.  Personally I'd rather the information be correct, even if less explicit in this case.

synbit: In general, I agree. I'd rather have something that is correct. The simple thing is that you cannot prove this is wrong, in the same way I cannot prove it's right. I could claim I'm right because my default installations comply with it, and I am sure you can compile your own custom version in a way that makes more sense to you and in this case it **might** behave differently.

It might be me, but when I use some sort of reference I expect it to be accurate in **most** situations, **not** all.

In short, you are **absolutely** right, leaving just the two options (zero/non-zero) will be correct everywhere. Which is why these two cases are included on top of the section. More explicit exit statuses follow which **may** be useful to people. That is the idea behind it.

Is there a way to have more input for this? I am only 51% against deleting it, so having more opinions will help us make a better decision I think, even if it's not going to be the correct one.

mintsoft: Yeah there is, we can wait for other people to chime in,

@MrChrisW @moollaza @mattr555 Opinions?

moollaza: TBH my Bash knowledge is pretty weak so keeping that in mind, if it works for the general case (and the source is reliable) then it's OK.

We can always put a disclaimer/explanation in the description if necessary, or utilize the Cheat Sheet's Description to explain more about this cheat sheet and that certain things may vary based on the user's distro/architecture etc:

For example it would be helpful to indicate this:

>Yet, most of it (if not all) **seems to tally with what is seen in RHEL based systems and Debian based systems (the 127+n status for instance)**. As the reference says, this is not a standard but it's a recommendation in an attempt to follow C/C++ standards, as the author says.

As well, it's likely that once we announce this IA, there will be some feedback and we can see what the community has to say about it. Maybe we'll need to generalize it here, and break down the exit statuses for specific distros/architectures in a "bash exist status cheat sheet"?

Does that work?

share/goodie/cheat_sheets/bash.json
```diff
@@ -0,0 +1,565 @@
...
+                "key": "cmd1 | cmd2",
+                "val": "Output of cmd1 is passed to cmd2's stdin"
+            },
+            {
+                "key": "cmd >&n",
+                "val": "Send cmd's stdout to file descriptor n"
+            },
+            {
+                "key": "cmd m>&n",
+                "val": "Same as above except that output that would normally go to file descriptor m, will also go to n"
```
zachthompson: Really should have the short forms "&>" and "&>>" for doing both stdout/err.

share/goodie/cheat_sheets/bash.json
zachthompson: I vote for "0" and "not 0" only.  The rest are too overkill for a cheat sheet.

share/goodie/cheat_sheets/bash.json
```diff
@@ -0,0 +1,565 @@
...
+            {
+                "key": "130",
+                "val": "Script terminated by Ctrl-C"
+            },
+            {
+                "key": "255*",
+                "val": "Exit status out of range (example: -1)"
+            }
+        ],
+        "Here Documents": [
```
zachthompson: Why single this out from **Redirection** as it is in the man page?  You already give here-docs above as well.

share/goodie/cheat_sheets/bash.json
mintsoft: Good catch!

share/goodie/cheat_sheets/bash.json
synbit: That's exactly what I said further up. I think the **Here Documents** section needs to go. I'll do that then.

share/goodie/cheat_sheets/bash.json
synbit: The disclaimer option feels unnecessary to me; a bash cheat sheet presented by an instant answer as a result of a search cannot be seen as something that will always be correct and under all systems. That's why the mintsoft's argument was not convincing. The source has been triple-checked (two books and tldp.org website), but this doesn't mean anything really. The overkill argument is understandable; I don't think this is the first thing people are going to be looking at on this cheat sheet, even though it could be useful. But then again... "SEARCH BETTER" I suppose...

For different reasons the zero/not-zero option seems more popular. I can always be changed later on if you prefer, but I think it makes more sense to keep just these two options for now.

share/goodie/cheat_sheets/bash.json
zachthompson: Sort of problematic since the newlines matter and wrapping can mess with it.  Can we do embedded newlines in keys, @moollaza?  I tried a few ways and it didn't seem like it.

share/goodie/cheat_sheets/bash.json
synbit: Yeah, can do. I don't remember when was the last time I used a user-defined file descriptor, but the idea was that you can put whatever you want to the variable ```n```, not just ```stdout```/```stderr```. Very possibly I'm wrong here, but I think ```&>``` and ```&>>``` are both for ```stdout``` and ```stderr``` like this. There is also ```>&``` which I think is the same as ```&>```? 

Are you suggesting adding those in addition to what's already there or you think it makes more sense to have only those?

zachthompson: Yeah, please add them.  They are very concise alternatives to `>word 2>&1` and `>>word 2>&1.  `>&` is an alt but the man page prefers &> and it also doesn't exist for the append version afaik.  I would skip those.

share/goodie/cheat_sheets/bash.json
synbit: Done.

share/goodie/cheat_sheets/bash.json
synbit: Added.

share/goodie/cheat_sheets/bash.json
moollaza: @zachthompson nope not right now. We can/may add support but nothing right now.

mintsoft: Maybe a description such as "The content of the heredoc terminated with x is passed to stdin of the command" ?

zachthompson: If we want to keep this it should include something along the lines of "content of here-doc on a new line and finalized by \<string\> on its own line."  You could potentially put something like,
```
cmd <<EOF [enter] here-doc [enter] EOF
```
replacing `[enter]` with whatever is clearest but it's getting a bit long.

synbit: @zachthompson As suggested earlier, it would be best to be displayed like this:
```
cmd <<'string'
    content...
string
```
but, that's not possible as I understand it. Having a very long key like the one suggested above will probably force it to put things in new lines, but the problem is, I think, we have no control as to which bits go to which line (actually, as it currently stands it's confusing anyway. For me for example, the ```'<string>'``` portion of the key is displayed on the next line. When one copies it from the cheat sheet is fine, but looking at it can be confusing). 
Would it be better if we were to include a hint:
```Contents up to a line identical to <string> become stdin for cmd verbatim (here-document start)```
With regards to the key, it might be better to replace ```'<string>'``` with something shorter (the standard ```'EOF'``` maybe?) in an attempt to force it display as we want, if that works... Here is a screenshot:
here-doc
The ```cmd << <string>``` will be changed accordingly then (which I haven't done in the screenshot as you can see). Any thoughts?

Comment (synbit) [inline share/goodie/cheat_sheets/bash.json]: BTW, the plan is to make this complete so that the ```Here Documents``` section can go. Unless the opposite is preferred.

Comment (zachthompson) [inline share/goodie/cheat_sheets/bash.json]: I like the short EOF identifier but it's still not clear to me that the here-doc needs to start on its own line.  This is borderline too much potential for confusion and should be removed.  A cheat sheet doesn't need to be a comprehensive reference and this seems fairly obscure.  Have any of you used here-docs with bash regularly?  I use with Perl fairly often but have never needed it with bash.

Comment (synbit) [inline share/goodie/cheat_sheets/bash.json]: I often abuse it due to the lack of block comments. It's one of those things that it's nice to know it's there for the one case you are going to need it...

Comment (synbit) [inline share/goodie/cheat_sheets/bash.json]: Both keys have been removed.

File: share/goodie/cheat_sheets/bash.json (hunk 23 above)
Comment (synbit) [inline share/goodie/cheat_sheets/bash.json]: @zachthompson *Here Documents* section has been removed (as well as the respective keys from the *IO Redirection* section.

File: share/goodie/cheat_sheets/bash.json [hunk 24]
```diff
@@ -0,0 +1,518 @@
...
+            {
+                "key": "[<string>], [M-?]",
+                "val": "Display commands starting with <string>"
+            },
+            {
+                "key": "M-*",
+                "val": "Insert all possible completions"
+            }
+        ],
+        "Backslash Escape Sequences": [
```
synbit: Does anyone else feel this section is not as useful? The *Here Documents* section seems vital in comparison in my opinion... All one needs to know is that ```\``` is the escape character. Opinions?

zachthompson: Looking at the description in the man page certainly seems a bit arcane.

mintsoft: I concur, it seems basically pointless; although perhaps a separate "ANSI escape sequences" cheatsheet might be useful

synbit: That's now gone.

mintsoft:
@moollaza I think this is ready then now that everything looks resolved

moollaza:
@synbit @mintsoft thanks a lot guys, and sorry for the delay!

This LGTM :+1:

moollaza:
@synbit congrats! It's live: https://duckduckgo.com/?q=bash+cheat+sheet&ia=cheatsheet

PR #1372, opened 2015-08-06
Cheat sheets aliases added

Sloff:
Hi @moollaza, as discussed in #1322 I've added an alias functionality that uses an `aliases` array within the json files, let me know if you think I should do something differently.

### DIFF
--- a/lib/DDG/Goodie/CheatSheets.pm
+++ b/lib/DDG/Goodie/CheatSheets.pm
@@ -21,10 +21,43 @@ triggers startend => (
     'key bindings', 'keys', 'default keys'
 );
 
+sub getAliases {
+    my @files = File::Find::Rule->file()
+                                ->name("*.json")
+                                ->in(share());
+    my @results;
+    
+    foreach my $file (@files) {
+        open my $fh, $file or return;
+        my $json = do { local $/;  <$fh> };
+        my $data = decode_json($json);
+        
+        if ($data->{'aliases'}) {
+            foreach my $alias (@{$data->{'aliases'}}) {
+                push (@results, $alias);
+                push (@results, File::Basename::fileparse($file));
+            }
+        }
+    }
+    
+    return @results;
+}
+
+my %alias = (
+    getAliases()
+);
+
 handle remainder => sub {
     # If needed we could jump through a few more hoops to check
     # terms against file names.
-    my $json_path = share(join('-', split /\s+/o, lc($_) . '.json'));
+    my $json_path;
+    
+    if ($alias{lc($_)}) {
+        $json_path = share($alias{lc($_)});
+    } else {
+        $json_path = share(join('-', split /\s+/o, lc($_) . '.json'));
+    }
+    
     open my $fh, $json_path or return;
     my $json = do { local $/;  <$fh> };
     my $data = decode_json($json);
--- a/lib/DDG/Goodie/CheatSheets.pm
+++ b/lib/DDG/Goodie/CheatSheets.pm
@@ -28,21 +28,19 @@ sub getAliases {
     my %results;
     
     foreach my $file (@files) {
-        open my $fh, $file or return;
+        open my $fh, $file or warn "Error opening file: $file\n" and next;
         my $json = do { local $/;  <$fh> };
         my $data = decode_json($json);
         
-        my $filename = File::Basename::fileparse($file);
-        
-        my $defaultName = $filename;
+        my $defaultName = File::Basename::fileparse($file);
         $defaultName =~ s/-/ /g;
         $defaultName =~ s/.json//;
         
-        $results{$defaultName} = $filename;
+        $results{$defaultName} = $file;
         
         if ($data->{'aliases'}) {
             foreach my $alias (@{$data->{'aliases'}}) {
-                $results{$alias} = $filename;
+                $results{$alias} = $file;
             }
         }
     }
@@ -54,7 +52,7 @@ my $aliases = getAliases();
 handle remainder => sub {
     # If needed we could jump through a few more hoops to check
     # terms against file names.
-    open my $fh, share($aliases->{join(' ', split /\s+/o, lc($_))}) or return;
+    open my $fh, $aliases->{join(' ', split /\s+/o, lc($_))} or return;
     
     my $json = do { local $/;  <$fh> };
     my $data = decode_json($json);
--- a/lib/DDG/Goodie/CheatSheets.pm
+++ b/lib/DDG/Goodie/CheatSheets.pm
@@ -25,40 +25,37 @@ sub getAliases {
     my @files = File::Find::Rule->file()
                                 ->name("*.json")
                                 ->in(share());
-    my @results;
+    my %results;
     
     foreach my $file (@files) {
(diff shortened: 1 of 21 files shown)


lib/DDG/Goodie/CheatSheets.pm
```diff
@@ -21,10 +21,43 @@ triggers startend => (
     'key bindings', 'keys', 'default keys'
 );
 
+sub getAliases {
+    my @files = File::Find::Rule->file()
+                                ->name("*.json")
+                                ->in(share());
+    my @results;
```
mintsoft: It'd probably be clearer to switch this to `%results` and switch the pushes below to something like:
```perl
$results{$alias} = File::Basename::fileparse($file);
```

moollaza:
@Sloff thanks a lo! We'll take a look and give you some feedback. Overall I like this direction much more. I'd like to hear @zachthompson 's thoughts as well

lib/DDG/Goodie/CheatSheets.pm
zachthompson: :+1:

lib/DDG/Goodie/CheatSheets.pm
```diff
@@ -21,10 +21,43 @@ triggers startend => (
...
+    
+    foreach my $file (@files) {
+        open my $fh, $file or return;
+        my $json = do { local $/;  <$fh> };
+        my $data = decode_json($json);
+        
+        if ($data->{'aliases'}) {
+            foreach my $alias (@{$data->{'aliases'}}) {
+                push (@results, $alias);
+                push (@results, File::Basename::fileparse($file));
```
zachthompson: The alias should just map to the absolute path given by $file. Like @mintsoft suggested, just do a hash here.

lib/DDG/Goodie/CheatSheets.pm
```diff
@@ -21,10 +21,43 @@ triggers startend => (
...
+    my @files = File::Find::Rule->file()
+                                ->name("*.json")
+                                ->in(share());
+    my @results;
+    
+    foreach my $file (@files) {
+        open my $fh, $file or return;
+        my $json = do { local $/;  <$fh> };
+        my $data = decode_json($json);
+
```
zachthompson: Add an alias for the default query as well.

lib/DDG/Goodie/CheatSheets.pm
```diff
@@ -21,10 +21,43 @@ triggers startend => (
...
-    my $json_path = share(join('-', split /\s+/o, lc($_) . '.json'));
+    my $json_path;
+    
+    if ($alias{lc($_)}) {
+        $json_path = share($alias{lc($_)});
+    } else {
+        $json_path = share(join('-', split /\s+/o, lc($_) . '.json'));
+    }
+    
     open my $fh, $json_path or return;
```
zachthompson: If you add the default query, the lookup value for any query should be:
```Perl
open my $fh, $aliases->{join(' ', split /\s+/o, lc($_)} or return;
```
The if/else above isn't needed.

lib/DDG/Goodie/CheatSheets.pm
```diff
@@ -21,10 +21,43 @@ triggers startend => (
...
+                push (@results, $alias);
+                push (@results, File::Basename::fileparse($file));
+            }
+        }
+    }
+    
+    return @results;
+}
+
+my %alias = (
```
zachthompson: my $aliases = getAliases();

lib/DDG/Goodie/CheatSheets.pm
```diff
@@ -21,10 +21,43 @@ triggers startend => (
...
+        
+        if ($data->{'aliases'}) {
+            foreach my $alias (@{$data->{'aliases'}}) {
+                push (@results, $alias);
+                push (@results, File::Basename::fileparse($file));
+            }
+        }
+    }
+    
+    return @results;
```
zachthompson: return \%results;

Sloff:
Hi @zachthompson, changes made as suggested

mintsoft:
@moollaza @zachthompson Is there no tests for the general cheatsheet IA?

moollaza:
@mintsoft not quite yet. We should probably make some though! ;)

lib/DDG/Goodie/CheatSheets.pm
```diff
@@ -21,11 +21,41 @@ triggers startend => (
...
+        my $json = do { local $/;  <$fh> };
+        my $data = decode_json($json);
+        
+        my $filename = File::Basename::fileparse($file);
+        
+        my $defaultName = $filename;
+        $defaultName =~ s/-/ /g;
+        $defaultName =~ s/.json//;
+        
+        $results{$defaultName} = $filename;
```
zachthompson: How about mapping aliases here to `$file` instead of `$filename` so we don't need to call `share` below?  The path won't change.

Sloff: Hi @zachthompson, I tried that but the path I get is `/home/codio/workspace/...`

zachthompson: Not sure I follow.  You've just used `$file` on line 31 to actually open it so we know that's the right path. At this point so you should be able to store it and refer to it later, rather than re-deriving it via `share`.

lib/DDG/Goodie/CheatSheets.pm
```diff
@@ -21,11 +21,41 @@ triggers startend => (
...
 );
 
+sub getAliases {
+    my @files = File::Find::Rule->file()
+                                ->name("*.json")
+                                ->in(share());
+    my %results;
+    
+    foreach my $file (@files) {
+        open my $fh, $file or return;
```
zachthompson: Not quite sure `return` is the right response here.  If we can't open a file, we should probably `warn` then `next` so the entire IA isn't dead in the water due to a single file.

lib/DDG/Goodie/CheatSheets.pm
Sloff: Sorry @zachthompson, think something else was wrong last time I tried using `$file`, it's working now!

Sloff:
Hi @zachthompson, changes made as suggested

zachthompson:
Lookin' pretty good @Sloff.  Would you be interested in creating a test script for this IA as was suggested?

Sloff:
Will do @zachthompson! Should I add it to this pull request or should I create a new one?

zachthompson:
Let's just add it here.

moollaza:
@Sloff we've just merged a PR that adds new functionality to Cheat Sheets so there's a conflict now. Let's get this PR updated so that it's mergable again as I'd like to get it out soon.

@zachthompson in the interest of getting this out, shall we add the tests in another PR?

zachthompson:
@moollaza when are we planning to release next?

mintsoft:
@Sloff  can you rebase? There are conflicts on the merge

Sloff:
Hi all, conflicts resolved

moollaza:
@zachthompson I plan to release today. Deploy Friday or Monday

PR #1375, opened 2015-08-07
Added English Braille Cheat Sheet

chappers:

Add content to the cheatsheet instant answers.

The cheat sheet for english braille is based on the article on wikipedia. <https://en.wikipedia.org/wiki/English_braille>

It is easily accessible compared with digging through the standards which is in a pdf.

Alternative data sources could be from the International Council on English Braille (ICEB) <http://www.iceb.org/ueb> which has the same alphabet encoding on the "Rules of Unified English Braille 2013.pdf" on the same webpage. The encoding is on pg 41

English Braille Cheat Sheet

Reference, Special Interest

No

Not right now

From the stopwatch/timer portion of duck duck go

image

-----
IA Page: https://duck.co/ia/view/english_braille_cheat_sheet

### DIFF
--- a/share/goodie/cheat_sheets/english-braille.json
+++ b/share/goodie/cheat_sheets/english-braille.json
@@ -7,7 +7,8 @@
     },
     "section_order": [
         "Alphabet",
-        "Digit"
+        "Digit",
+        "Formatting Marks"
     ],
     "sections": {
         "Alphabet": [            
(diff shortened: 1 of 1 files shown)


moollaza:
@chappers thanks a lot for contributing, we really appreciate it! We'll be sure to review this shortly and give you some feedback!

mintsoft:
@chappers this looks pretty good to me, the only thing is there doesn't appear to be a difference between the numbers and the first letters of the alphabet. Is it context sensitive or are there markers to indicate when it's digits etc?

chappers:
@mintsoft There are formatting marks for letter, capitalization, and numbers, but they don't really have an English equivalent, so I'm not sure whats the best way to place it on the cheat sheet.

moollaza:
>There are formatting marks for letter, capitalization, and numbers, but they don't really have an English equivalent, so I'm not sure whats the best way to place it on the cheat sheet.

@chappers can you give an example? Maybe we should have examples as another section?

chappers:
I probably should have checked with you before actually changing it. I have some examples as shown here:

image


I switched `key` and `val`, which is similar to what [nato alphabet cheat sheet](https://duckduckgo.com/?q=nato+alphabet+cheatsheet&ia=cheatsheet) did, and added some examples on how it would work.

moollaza:
>I switched key and val, which is similar to what nato alphabet cheat sheet did, and added some examples on how it would work.

:+1: That seems fine

@chappers let's put these examples into a new section though so it's separated from the other two parts

moollaza:
@abeyang any thoughts on design?

abeyang:
@chappers If that screenshot that you've provided is the latest (with the formatting marks section), then it looks good! :+1:

chappers:
@abeyang Yes, that screenshot is the latest.

abeyang:
Then it's :+1: from me. Unless we're ready to try out the reference sheet -- @moollaza ?

moollaza:
LGTM :+1: Will add the reference template myself..

moollaza:
@chappers congrats! It's live: https://duckduckgo.com/?q=english+braille+cheat+sheet&ia=cheatsheet

PR #1379, opened 2015-08-09
New Cheat Sheet Templates

alohaas:
Hey @russellholt,

This is a first pass at adding template variants to cheat sheets.
I've modified the following existing cheat sheets to test the new templates:
- vim: keyboard-shortcuts
- git: terminal-snippets
- regex: code-snippets
- wu-tang: reference-sheet

I have some lingering questions that I hope we can find some time to discuss:
1. I removed the `<span>` tags surrounding the `<h6 class="cheatsheets__title">` because it seemed to be adding mysterious vertical space. Not sure if the `<span>` had a specific purpose...
2. There is a lot of repetition in each handlebars file for the templates. I wanted to discuss the best way to approach a refactor before paring it down. I'm not sure if there is an existing paradigm in IAs for extending templates from a base template or using macros to generate specific blocks of markup.
3. I'm passing an empty string as a third parameter to `{{#cheatsheets_ordered ...}}` in `detail.handlebars` in place of "template_type". I took the simplest approach to ensure that of the existing cheat sheets can still rely on the original template; however, I think it would be more elegant to allow for the third parameter to be unspecified. I went for the quickest solution, but would like to explore how to achieve this in `cheat_sheets.js`. An alternate approach I considered was to point all cheat sheets with unspecified template types to the new `code-snippets.handlebars` template, but I didn't want to shake things up too much in the early stages of this enhancement.

Looking forward to your feedback :)

https://duck.co/ia/view/cheat_sheets


russellholt:
tested, looking good so far. will. answer these specific questions later tonight, along with other feedback.

moollaza:
Hey @alohaas great job so far! Russell asked me to chime in so let me address your questions:

> 1 - I removed the `<span>` tags surrounding the `<h6 class="cheatsheets__title">` because it seemed to be adding mysterious vertical space. Not sure if the `<span>` had a specific purpose...

I don't recall why I added the `span` and it doesn't seem to be affecting the UI much, so removing it should be OK.

> 2 - There is a lot of repetition in each handlebars file for the templates. I wanted to discuss the best way to approach a refactor before paring it down. I'm not sure if there is an existing paradigm in IAs for extending templates from a base template or using macros to generate specific blocks of markup.

We can definitely reduce the repetition! If we utilize our  `include` helper we can pull in sub-templates and specify the context for them. This way the static parts like the `name`, `description`, `footer`, etc can be in the "base" template. Then, based on cheat sheet's template, we can include the right sub-template to render the dynamic part (the `table` or `ul`).

The helper is documented here: https://duck.co/duckduckhack/handlebars_helpers#codeincludecode
You can see an example usage in the Recipes IA: https://github.com/duckduckgo/zeroclickinfo-spice/blob/master/share/spice/recipes/recipes_item_detail.handlebars#L10

> 3 - I'm passing an empty string as a third parameter to {{#cheatsheets_ordered ...}} in detail.handlebars in place of "template_type". I took the simplest approach to ensure that of the existing cheat sheets can still rely on the original template; however, I think it would be more elegant to allow for the third parameter to be unspecified. I went for the quickest solution, but would like to explore how to achieve this in cheat_sheets.js. An alternate approach I considered was to point all cheat sheets with unspecified template types to the new code-snippets.handlebars template, but I didn't want to shake things up too much in the early stages of this enhancement.

I'm pretty sure all arguments are already optional? I just removed the `""` from the template and everything seems to be working fine? You should be able to set the `template_type` to a default if the parameter isn't defined. Alternatively, you can allow the `template_type` to be defined in the optional `options` hash that gets passed along:

```handlebars
{{!-- no template_type --}}
`{{#cheatsheets_ordered sections section_order}}`

{{!-- template_type --}}
`{{#cheatsheets_ordered sections section_order template_type="foo"}}`
```

russellholt:
Thanks @moollaza .

@alohaas What I think we should do now is look onto refactoring those templates. @abeyang will review the design details, and when that's all good we should launch it so everyone can start using it!

alohaas:
Thanks @moollaza and @russellholt —
I refactored, taking advantage of the {{include }} helper.
I still have some repetition in three table-based templates (code-snippets, terminal-snippets, keyboard-shortcuts). The three files are identical and I could consolidate them; however I see some semantic value in keeping these templates separate...

russellholt:
@alohaas We need to change the way we use color here, to handle themes.
See https://duckduckgo.com/styleguide  for how to use theme-friendly colors.

abeyang:
@alohaas hey, thanks for this -- it's looking really good! As @russellholt suggested, for colors (text and/or background), we'll need to use the special classnames so that it'll jive well with other themes. One thing issue you **might** bump into is the bullet-point colors... in any case, please give it a shot, and if you have any issues, give us a holler!

In the meantime, I'm going to provide some feedback for additional fixes; mostly just tweaks here and there. Good job!

abeyang:
First -- and this *may* be the hardest to implement -- can we fix the margins between the key and the description? The distance between the largest key and the desc should be **18px**, so it should look something like this:

screen shot 2015-08-10 at 1 59 09 pm

abeyang:
Next, I'm going to refer to my specs here for the margins inbetween the pairings and groupings:

cheatsheet v2 4-guides

Let's fix a few things: the boxes for the keys, the margins between the rows, and the margins between the groups:

``` css
/* fix the for shortcut keys and code snippets */
.zci--cheat_sheets code {
  padding: 3px 4px 1px;
}

/* fix margins between the rows */
.zci--cheat_sheets td {
  padding-bottom: 8px;
}

/* fix margins between groups and titles */
.zci--cheat_sheets .cheatsheet__title {
  padding: 16px 0 12px;
}
```

alohaas:
@russellholt and @abeyang thanks for the feedback. I can make those changes.

alohaas:
Hey @abeyang, I think I hit everything!
You suggested that adding an 18px margin between the key and the description might be the hardest part to implement...felt like it was too easy :grimacing: I may have misunderstood.

russellholt:
thanks @alohaas , looks good!

@abeyang ?

abeyang:
@alohaas , much better! And yes, I think there's a misunderstanding: for **each** grouping, the smallest distance between the largest key and the desc should be 18px. So here's an example based on your latest changes: note that there's still a large gap between the two:

pasted_image_8_11_15__10_57_am

Hopefully that makes more sense! Good luck!

alohaas:
@abeyang gotcha! Thanks for the clarification :+1:

russellholt:
screen shot 2015-08-11 at 3 02 10 pm

russellholt:
screen shot 2015-08-11 at 4 04 57 pm

russellholt:
LGTM :+1:

abeyang:
Awesome, looks good! :+1:

PR #1382, opened 2015-08-10
Week: Temporal test that's already covered with MockTime

mintsoft:
This is breaking the tests


moollaza:
@mintsoft shoot, I didn't notice that!

mintsoft:
@moollaza Yeah; it looks like someone (IIRC I) covered it with MockTime but didn't delete the original

PR #1388, opened 2015-08-12
Preserve cheat sheet state and restore it on page load

bsstoner:
Adds `iax=1` to the querystring when you click Show More on a cheat sheet, and restores it to the expanded state on page load if `iax=1` is in the querystring. It uses the same `env.history` API that we use internally to preserve IA state.

This will allows cheat sheets to be bookmarked and shared in the expanded state.

@zachthompson @russellholt @moollaza


moollaza:
LGTM :+1:

PR #1394, opened 2015-08-13
Move json files into their own directory

zachthompson:
Let's move the json files to their own subdirectory so we can better see what's a cheat sheet and what's an IA asset.


moollaza:
LGTM :+1:

PR #1397, opened 2015-08-14
Cheatsheet tests added

Sloff:
Test file for cheat sheets added.

cc @zachthompson @mintsoft @moollaza

### DIFF
--- a/lib/DDG/Goodie/CheatSheets.pm
+++ b/lib/DDG/Goodie/CheatSheets.pm
@@ -4,6 +4,7 @@ package DDG::Goodie::CheatSheets;
 use JSON::XS;
 use DDG::Goodie;
 use DDP;
+use File::Find::Rule;
 
 no warnings 'uninitialized';
 
--- a/t/CheatSheets.t
+++ b/t/CheatSheets.t
@@ -5,15 +5,41 @@ use warnings;
 use Test::More;
 use DDG::Test::Goodie;
 use JSON::XS;
+use File::Find::Rule;
 
 zci answer_type => "cheat_sheet";
 zci is_cached   => 1;
 
-sub build_answer {
-    my ($filename) = @_;
-    my $file = 'share/goodie/cheat_sheets/json/'.$filename;
+sub getTests {
+    my @files = File::Find::Rule->file()
+                                ->name("*.json")
+                                ->in('share/goodie/cheat_sheets/json/');
+    my %tests;
     
-    open my $fh, $file or return;
+    foreach my $file (@files) {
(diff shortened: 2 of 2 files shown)


lib/DDG/Goodie/CheatSheets.pm
```diff
@@ -4,6 +4,7 @@ package DDG::Goodie::CheatSheets;
 use JSON::XS;
 use DDG::Goodie;
 use DDP;
+use File::Find::Rule;
```
mintsoft: Is this needed here?

t/CheatSheets.t
```diff
@@ -0,0 +1,44 @@
...
+            item => 0,
+            options => {
+                content => "DDH.cheat_sheets.detail",
+                moreAt => 0
+            }
+        }
+    };
+}
+
+ddg_goodie_test(
```
mintsoft: Could/should we iterate over all the cheatsheets in `share` and make sure they all work instead? That way when a new user creates a PR it'd be automatically tested, hopefully it'd catch parse errors in the JSON etc

lib/DDG/Goodie/CheatSheets.pm
zachthompson: It is.

t/CheatSheets.t
zachthompson: Great idea.

Sloff:
Hi all, improvements made as suggested

t/CheatSheets.t
```diff
@@ -0,0 +1,67 @@
...
+        
+        my $defaultName = File::Basename::fileparse($file);
+        $defaultName =~ s/-/ /g;
+        $defaultName =~ s/.json//;
+        
+        $tests{$defaultName." cheat sheet"} = test_zci(build_answer($file));
+        
+        if ($data->{'aliases'}) {
+            foreach my $alias (@{$data->{'aliases'}}) {
+                $tests{$alias." cheat sheet"} = test_zci(build_answer($file));
```
zachthompson: Pass `$data` to build_answer so you don't need to open the file and decode again.

t/CheatSheets.t
```diff
@@ -0,0 +1,67 @@
...
+            }
+        }
+    }
+    return %tests;
```
zachthompson: Return a ref.

zachthompson:
@Sloff Ran it and it works very well! Just needs a couple of tweaks.

Sloff:
Hi @zachthompson, changes made as suggested

PR #1401, opened 2015-08-15
Added Boolean algebra Cheat Sheet 

gautamkrishnar:
It will display all basic equations of Boolean algebra as cheat sheet

(n/a)

(n/a)

(n/a)

Students

https://duck.co/ideas/idea/6230/logic-cheat-sheet

Just started contributing... Its my first pull request :)

When i was wandering online, I found it, I fell in love with it.

screenshot 423


share/goodie/cheat_sheets/json/logic.json
```diff
@@ -0,0 +1,172 @@
+{
+  "id": "logic_cheat_sheet",
+  "name": "logic",
```
moollaza: Let's change this to "Boolean Algebra" or "Boolean Algrebra Logic" -- something a little more descriptive/specific

share/goodie/cheat_sheets/json/logic.json
```diff
@@ -0,0 +1,172 @@
+{
+  "id": "logic_cheat_sheet",
```
moollaza: similarly, I think this should be `boolean_algebra_cheat_sheet`

share/goodie/cheat_sheets/json/logic.json
```diff
@@ -0,0 +1,172 @@
...
+  "name": "logic",
+  "description": "These are a set of operators and expressions used in the boolean algebra",
+  "metadata": {
+    "sourceName": "Image",
+    "sourceUrl": "http://i.imgur.com/OXdpVxx.png"
+  },
+  "aliases": [
```
moollaza: :+1:

share/goodie/cheat_sheets/json/logic.json
```diff
@@ -0,0 +1,172 @@
+{
+  "id": "logic_cheat_sheet",
+  "name": "logic",
+  "description": "These are a set of operators and expressions used in the boolean algebra",
+  "metadata": {
+    "sourceName": "Image",
+    "sourceUrl": "http://i.imgur.com/OXdpVxx.png"
```
moollaza: Any chance we can find a slightly better/more-usable source? Does https://en.wikipedia.org/wiki/Boolean_algebra cover everything?

share/goodie/cheat_sheets/json/logic.json
moollaza: We also now have support for a few CheatSheet templates. Let's try adding `"template_type": "reference"` which should make this easier to read. Alternatively the `code` template may also work well.

moollaza:
@gautamkrishnar thanks a lot for contributing! We really appreciate it :smile: 

@abeyang this one is a new edge case -- a lot of the "keys" don't really have a value or explanation. I've asked the dev to switch this to the `reference` template and I think the final result should still look OK.

Any thoughts on the content? I think we should probably space out the math a little so the letters and symbols aren't all squished together?

gautamkrishnar:
Added all changes...
Thanks for your support...

moollaza:
@gautamkrishnar whoops, it looks like you have conflicts on your PR.

Please update it so that the only file being added is the single .json file for your Cheat Sheet. We now have an `"aliases" : ...` property which you can use to provide an array of additional trigger words.

In this case for "boolean-algebra.json" you would want to add `"aliases": "boolean logic", "logic" ...`

gautamkrishnar:
Moved to new pull request:
https://github.com/duckduckgo/zeroclickinfo-goodies/pull/1424
@moollaza @abeyang

share/goodie/cheat_sheets/json/boolean_algrebra.json
```diff
@@ -0,0 +1,195 @@
...
+      },
+      {
+        "val": "Not",
+        "key": "¬"
+      }
+    ],
+	
+    "Basic Principles": [
+      {
+        "key": "X+0=X"
```
moollaza: I feel like we should space these out for legibility. We should put spaces around the `=` and `+` and other operators

PR #1404, opened 2015-08-17
Fixes over triggering of binary logic

### LINKED ISSUE #1403: BinaryLogic: Over triggering, throwing errors
Generic queries, like "what do number and letter codes in a lens name mean" are triggering this IA and throwing errors. We need to fix this asap!

Also, let’s remove the regex trigger in favour of a much more efficient words trigger:

```perl
triggers any => "and", "or", "xor", "⊕", "∧", "∨","not", "¬";
```

We can move the regex's into the `handle` function and use them to verify our inputs. The regexs need to be improved to prevent unexpected inputs and the captures from the regex need to be checked to ensure they are valid numbers.

/cc @Sloff

Sloff:
Hi @moollaza this will fix #1403

### DIFF
--- a/lib/DDG/Goodie/BinaryLogic.pm
+++ b/lib/DDG/Goodie/BinaryLogic.pm
@@ -99,13 +99,17 @@ sub BinaryLogic_Actions::do_not {
 }
 
 handle query_raw => sub {
+    my $input = $_;
+    
+    my $testError = $input;
+    $testError =~ s/(?:0x|0b|[\d\s]|and|or|xor|not|\(|\)|⊕|∧|∨|¬)//ig;
+    return undef if length $testError != 0;
+
     my $grammar = Marpa::R2::Scanless::G->new({ source => \$rules });
     my $recce = Marpa::R2::Scanless::R->new({
         grammar => $grammar,
         semantics_package => 'BinaryLogic_Actions'
     });
-
-    my $input = $_;
 
     # Substitute the unicode characters. The parser does not seem to
     # like unicode.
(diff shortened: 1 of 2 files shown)


moollaza:
@Sloff wow, you *just* beat me. I was about to submit a PR where I've moved the subtitle building loop below the parser runs. That way we catch errors and bail before formatting the subtitle.

moollaza:
I also added a bunch of failing tests. I think we should add in more "failing" tests to verify queries that could trigger but shouldn't return an answer, don't return anything:

https://github.com/duckduckgo/zeroclickinfo-goodies/compare/zaahir/fix-binary-logic

lib/DDG/Goodie/BinaryLogic.pm
```diff
@@ -99,14 +99,18 @@ sub BinaryLogic_Actions::do_not {
 }
 
 handle query_raw => sub {
+    my $input = $_;
+    
+    my $testError = $input;
+    $testError =~ s/(?:0x|0b|[\d\s]|and|or|xor|not|\(|\)|⊕|∧|∨|¬)//ig;
+    return undef if length $testError != 0;
```
moollaza: @Sloff you actually just want to `return` here, not `return undef`

moollaza:
@Sloff I've combined our PR's here: https://github.com/duckduckgo/zeroclickinfo-goodies/pull/1405

Thanks a ***lot*** for the very fast response.

Sloff:
Pleasure @moollaza!

PR #1409, opened 2015-08-17
CheatSheets: eval decode_json

zachthompson:
Wrap decode_json in eval so we don't bail entirely for one malformed cheat sheet.

### DIFF
--- a/lib/DDG/Goodie/CheatSheets.pm
+++ b/lib/DDG/Goodie/CheatSheets.pm
@@ -34,7 +34,7 @@ sub getAliases {
         open my $fh, $file or warn "Error opening file: $file\n" and next;
         my $json = do { local $/;  <$fh> };
         my $data = eval { decode_json($json) } or do {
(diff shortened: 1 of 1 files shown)


lib/DDG/Goodie/CheatSheets.pm
```diff
@@ -33,7 +33,10 @@ sub getAliases {
     foreach my $file (@files) {
         open my $fh, $file or warn "Error opening file: $file\n" and next;
         my $json = do { local $/;  <$fh> };
-        my $data = decode_json($json);
+        my $data = eval { decode_json($json) } or do {
+			warn "Failed to decode $fh: $@";
```
moollaza: Seems like `$fh` needs to be converted to a string?

1_____ssh_

zachthompson:
Whoops, needed to list the file and not the file handle in the error message.

russellholt:
:+1:

moollaza:
LGTM :+1:

PR #1415, opened 2015-08-19
move json into /json/ dir, switch to code template

moollaza:
@abeyang I think the "code" template works best here. The reference template looks a little odd because the bullets and braille dots are very similar.

Thoughts?

**Code Template**:
english_braille_cheat_sheet_at_duckduckgo

**Reference Template**:
english_braille_cheat_sheet_at_duckduckgo

### DIFF
--- a/share/goodie/cheat_sheets/json/english-braille.json
+++ b/share/goodie/cheat_sheets/json/english-braille.json
@@ -1,17 +1,18 @@
 {
     "id": "english_braille_cheat_sheet",
-    "name": "English Braille Cheat Sheet",
+    "name": "English Braille",
     "metadata": {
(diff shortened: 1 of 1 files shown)


abeyang:
@moollaza I agree -- let's use the code template for this!

russellholt:
could maybe break up the Alphabet section into A-L, M-Z to give the page more balance.

moollaza:
@russellholt good idea! I've split it in half:
english_braille_cheat_sheet_at_duckduckgo

abeyang:
@russellholt @moollaza Nice. I think eventually, we'll need to find a way to more organically split long columns in half.

russellholt:
@abeyang definitely

Issue #1433, opened 2015-08-25
XCode Cheat Sheet: Use ⌘ instead of "Command"


jagtalon:
It might look nicer to use ⌘ instead of "Command" in the cheat sheet.

![screen shot 2015-08-25 at 8 18 17 am](https://cloud.githubusercontent.com/assets/81969/9466307/e7153018-4b01-11e5-948a-1117e77835f8.png)

seisfeld:
I implemented this change on #1434 for everyone to review.

jagtalon:
Fixed by @seisfeld in #1434

PR #1434, opened 2015-08-25
changes "command" to "⌘", implements #1433

seisfeld:
As suggested here (#1433) i quickly made this change for everyone to see how it looks.

xcode-cheat-sheet


moollaza:
@seisfeld thanks a lot!

share/goodie/cheat_sheets/json/xcode.json
```diff
@@ -20,331 +20,331 @@
     "sections": {
         "Xcode menu command": [
```
moollaza: @seisfeld would you mind removing `comman` from all these headings? I just realized it's a little redundant. We should capitalize `Menu` as well for each.

seisfeld:
@moollaza I made your suggested changes.

jagtalon:
That was hella fast. :+1:

PR #1460, opened 2015-08-29
CheatSheets: htmlencode < >

### LINKED ISSUE #1414: Cheat Sheets: Need to encode angle brackets "<", ">" inside code blocks
Looks like the Emacs Cheat Sheet is slightly broken now. The "Search and Replace" section contains a key with `[<replace-string>]` in it. The angle brackets aren't being encoded and so a `<replace-string></replace-string>` element is being created on the page.

https://duckduckgo.com/?q=emacs+cheat+sheet&t=ffab&ia=cheatsheet&iax=1

-----

mattr555:
This seems to work... fixes #1414

/cc @moollaza @mintsoft

### DIFF
--- a/share/goodie/cheat_sheets/cheat_sheets.js
+++ b/share/goodie/cheat_sheets/cheat_sheets.js
@@ -36,7 +36,9 @@ DDH.cheat_sheets.build = function(ops) {
         var codeClass = typeof className === "string" ? className : "bg-clr--white";
         
         // replace escaped slashes and brackets
-        string = string.replace(/\\\\/, "<bks>")
+        string = string.replace(/\</g, '&lt;')
+                .replace(/\>/g, '&gt;')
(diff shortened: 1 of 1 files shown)


mintsoft:
LGTM

moollaza:
Seems OK. I think a more general `encode_entities` function will be needed so that we can encode others.

Something similar to underscore.js's `_.escape()` could work and could be added to our new open-source utils repo: https://github.com/duckduckgo/duckduckgo-utils/blob/master/util.js

We have a `DDG.strip_all_html()` actually which might work?

mintsoft:
@moollaza I looked at the `strip_all_html` doesn't that just remove it rather than encode it?

moollaza:
> I looked at the strip_all_html doesn't that just remove it rather than encode it?

@mintsoft looks like you're right. We should probably have an `encode_entities` method then :+1:

PR #1464, opened 2015-08-29
Fixes #1457 - updated code styles for terminal cheat sheet templates

### LINKED ISSUE #1457: Cheat Sheets Terminal Template: Long text causes columns to overlap
Looks like we'll need to adjust our CSS a bit to prevent this.

/cc @alohaas 

duckduckhack

alohaas:
Displaying the `<code>` elements as `inline-block` seemed to create a bubble of excess space in the right side of the column. Couldn't find a definitive answer as to why that was happening, but by moving a class assignment around in the `terminal.handlebars`, we can display `<code>` elements as `inline`, which solved the problem. CSS is weird :)

### DIFF
--- a/share/goodie/cheat_sheets/cheat_sheets.css
+++ b/share/goodie/cheat_sheets/cheat_sheets.css
@@ -55,26 +55,17 @@
 
 
 .zci--cheat_sheets .terminal td {
-  margin-bottom: 6px;
+  padding-bottom: 12px;
 }
 
 .zci--cheat_sheets .terminal code {
-  display: inline;
   width: 100%;
-  padding: 0;
-  margin: 0;
-  border: none;
-  background: none;
 }
 
 .zci--cheat_sheets .terminal td.cheatsheet__key {
   display: block;
   min-width: 100%;
-  padding: 3px 4px 1px;
-  line-height: 1.17;
-  border-radius: 2px;
-  word-break: break-all;
-  word-wrap: break-word;
+  padding-bottom: 1px;
 }
 
 .zci--cheat_sheets .terminal td.cheatsheet__value {
--- a/share/goodie/cheat_sheets/cheat_sheets.css
+++ b/share/goodie/cheat_sheets/cheat_sheets.css
@@ -30,7 +30,8 @@
     border-radius: 2px;
     font-size: 0.833em;
     white-space: normal;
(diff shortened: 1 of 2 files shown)


mintsoft:
@alohaas I'm not sure this is fixed in FF. I just hacked the perldoc cheatsheet to have content similiar to the mysql one in the screenie and it still appears to be problematic:

image

alohaas:
hm, let me see if I can replicate your content in the perldoc cheat sheet. I am working in FF, so I probably overlooked something.

mintsoft:
Looking at it, I think it might be a linewrap issue. I think the large `column1=value1` is forcing the box to overflow

alohaas:
Yeah I am seeing that too.

alohaas:
got it. `word-break` and `word-wrap`assignments need to happen to the parent `td.cheatsheet__key` not on the `code` element. Making this change.

alohaas:
@mintsoft seems to be better...will be submitting a new PR
screen shot 2015-08-29 at 11 30 33 am

mintsoft:
Cool, does that mean the original changes to the handlebars template were unnecessary or were they tweaking something that I've not noticed?

alohaas:
@mintsoft possibly! Let me take a look.

alohaas:
@mintsoft seemed to work without the template change :+1:

mintsoft:
Cool @moollaza I'll turn this over to you :)

moollaza:
@alohaas this LGTM :+1:

Thanks a lot for the awesome and quick fix!

alohaas:
You're welcome :grin:

moollaza:
@alohaas in working with more PR's I think I've discovered some visual regression from these changes. It looks like we're breaking words in the keyboard template and code template which causes some keys to look broken:

eclipse_cheat_sheet_at_duckduckgo
regex_cheat_sheet_at_duckduckgo

Looks like we should update the `word-break` CSS so that it only applies to the reference template.

Potentially we may need to get creative and only apply it to keys that have particularly long "words", but I think that should be a last resort.

alohaas:
Hi @moollaza, apologies for the slow response. I have just returned from a week-long trip with no access to my developer tools :sob: 

I will be looking into this issue tomorrow and will have a more thorough fix!

Issue #1466, opened 2015-08-29
RandomNumber: bounds are ignored if lower bound is 0


null-minus-null:
When the lower bound is zero, the upper bound is ignored and instead of the expected integer, a float between 0 and 1 is generated.

Example links:
* [random 0 100](http://ddg.gg/?q=random+0+100)
* [random 0 10000](http://ddg.gg/?q=random+0+10000)

------
IA Page:  http://duck.co/ia/view/random_number

loganom:
I'll look at this.

mintsoft:
closed by @loganom in #1490

PR #1490, opened 2015-08-30
iss1466 RandomNumber Correction for bounds of 0

loganom:
If the start or end value was 0, then no matter what, the default rand() value was returned (which is a fractional number). I added a check for the difference of the two numbers. If it's greater than 1, then we enter the random number if block.

Added a few tests to show this.

Here's to hoping I've done PR creation correctly.. it's been a while.

#1466

### DIFF
--- a/lib/DDG/Goodie/RandomNumber.pm
+++ b/lib/DDG/Goodie/RandomNumber.pm
@@ -37,9 +37,11 @@ handle query_lc => sub {
 
     ($end, $start) = ($start, $end) if ($start > $end);
 
+    my $valDiff = $end - $start;
+
     my $rand = rand;
 
-    if ($start && $end) {
+    if ($start && $end || $valDiff > 1) {
         $rand *= ($end - $start + 1);
         $rand = int($rand) + $start;
     }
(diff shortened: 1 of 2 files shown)


loganom:
@null-minus-null  @moollaza

loganom:
Should I also add myself to the attribution list?

mintsoft:
@loganom no reason not to add yourself :)

lib/DDG/Goodie/RandomNumber.pm
```diff
@@ -37,9 +37,11 @@ handle query_lc => sub {
 
     ($end, $start) = ($start, $end) if ($start > $end);
 
+    my $valDiff = $end - $start;
+
     my $rand = rand;
 
-    if ($start && $end) {
+    if ($start && $end || $valDiff > 1) {
         $rand *= ($end - $start + 1);
```
mintsoft: I'd probably reuse `$valDiff` here, so it'd be `$rand *= $valDiff +1 ;`

mintsoft:
@loganom Looks good to me, if you want to make that tiny tweak and add yourself to the attribution I'll marge when done :+1:

loganom:
@mintsoft thanks for the comments. Good to go.

mintsoft:
Thanks for the fix!

PR #1259, opened 2015-07-08
SubnetCalc: Updating to use the record template

mintsoft:
This basically looks the same; however it uses the template:

image

I found writing the tests for the structured answers really laborious; is there a short `.*` for the hash so that the structure only needs to be tested once?

PR #1492, opened 2015-08-31
SubnetCalc: Switched to structured_answer

mintsoft:
I've also changed the answer type to "subnet_calc" to be more consistent. I don't know if this is actually used anywhere except the tests/will cause any problems internally!

image


mintsoft:
I'm an idiot; I've already done this once! https://github.com/duckduckgo/zeroclickinfo-goodies/pull/1259

PR #1536, opened 2015-09-09
Cheat Sheets Parser Simplification and JSON Update

### LINKED ISSUE #1920: Add support for newline in cheatsheet
`\n` Must be parsed as a newline in cheatsheet key values

moollaza:
When I initially wrote this "parser" for Cheat Sheets I had very few JSON files to work with and one of these was the rege Cheat Sheet. It had several lines, that contained no spaces, but had unescaped `[]` which were meant to be taken literallly. E.g. `[:alpha:]`

I tried to make the parser flexible so that it would ignore a string without spaces, that contains un-escaped brackets/braces. At the time it seemed reasonable because the Regex JSON contained so many literal brackets that I thought would look bad if they were all escaped. 

After several weeks and reviewing nearly 100 Cheat Sheet PRs from new developers, I've come to realize that this behaviour confuses everyone, and has caused developers to change their data, likely without fully understanding why the syntax allows for this exception.

It's now clear to me that all un-escaped brackets/braces should be treated as though they are meant to wrap the text they contain inside a `<code>` block and only escaped brackets/braces should be taken literally.

This update simplifies our parser and escapes any previously un-escaped brackets/braces which are meant to be displayed literally.

**Marking as WIP for now**. I want to see if I can do a full, visual regression test across all cheat sheets to ensure this isn't going to break anything...

Fixes #1920 

/cc @mintsoft @jagtalon


share/goodie/cheat_sheets/cheat_sheets.js
```diff
@@ -27,33 +27,31 @@ DDH.cheat_sheets.build = function(ops) {
...
-                .replace(/\\\}/g, "<rcbr>");
-
-        // no spaces
-        // OR
-        // spaces and no un-escaped brackets
-        // e.g "?()", ":sp filename"
-        //  --> wrap whole sting in <code></code>
-        if ( !re_whitespace.test(string) || !re_brackets.test(string) ){
+        string = string
+            .replace(/\</g, '&lt;')
```
mintsoft: @moollaza Do you want to tackle the lack of a generalised html_enc whilst we're here?

mintsoft:
@moollaza Also I was thinking; one option to 'simplify' the escaping for everything would be to use html encoded entities instead. i.e. (`&#91;` and `&#93;`); I'm not sure if that'd be more or less intuitive for most people

share/goodie/cheat_sheets/json/regex.json
```diff
@@ -227,31 +263,6 @@
         }, {
             "val": "Octal Digit",
             "key": "\\O"
-        }],
```
mintsoft: This lot intentionally removed?

mintsoft: Sorry; didn't notice it'd been moved

moollaza: Whoops didn't actually mean to commit the moving part. I was contemplating re-organizing this one because the sections aren't actually created in the order they're displayed which just seems a little odd

share/goodie/cheat_sheets/cheat_sheets.js
moollaza: Now that I'm looking at the code more, I actually do run anything that goes inside the `<code>` block through `handlebars.escapeExpression` which is meant to do exactly that. Removing `.replace(/\>/g, '&gt;')` seems to have no effect now, which seems right. But I'll do some testing to see if I can break anything

moollaza: Actually I think I'm wrong, we do need to replace the angle brackets at this point...

moollaza:
> Also I was thinking; one option to 'simplify' the escaping for everything would be to use html encoded entities instead. i.e. (&#91; and &#93;); I'm not sure if that'd be more or less intuitive for most people

@mintsoft We're trying to keep the syntax as simple as possible right now and so I'd like to avoid HTML entities if we can reliably handle escaping things for users.

I think my most recent updates should manage to handle everything correctly, but I'm going to run a test to ensure there's no unintended regression.

moollaza:
@jagtalon would like a review. Don't merge though, would like to see if we can use the ddg-screen-diff tool to test for regression across cheat sheets...

PR #1544, opened 2015-09-09
Week: ensure we have a defined location and timezone

moollaza:
We don't always have a `$loc` or `$loc->time_zone` so we need to ensure we have on before creating a DateTime object.

/cc @mintsoft


mintsoft:
:+1:

PR #1561, opened 2015-09-13
New Kerala Cheat Sheet

jophab:
image


moollaza:
@jophab Thanks again for taking the time to contribute! We really appreciate it, but unfortunately I can't accept this as I don't believe this kind of information (about a City) is appropriate for cheat sheet currently.

I think this kind of information is best served by Wikipedia as it will be reliably maintained outside of DuckDuckGo. Currently our Wikipedia results don't show this granular information, but we are working towards exposing it, as Wikipedia does contain this data.

---

For the future, please make sure you reach out to use before making a working on something you plan to submit so we can pre-approve it and add it to the Community Platform so others will know what you're working on.

We really hope you'll stick around and continue to contribute, but please make sure you check in with us first.

PR #1585, opened 2015-09-17
Cheat sheet tester

jdorweiler:
Most of the code is from @dnmfarrell who gave us permission to commit this to our repo. 
https://github.com/dnmfarrell/DDG-cheatsheet-check

I made some changes to warn on non critical errors but it's mostly the same.

Running `prove` will test all cheatsheets in `share/goodies/cheat_sheets/json`.  You also have the option to test a single cheatsheet with 
`prove t/cheatsheet_check.t :: <name of file without .json>`
ex: `prove t/cheatsheet_check.t :: vim`

@moollaza @jagtalon


t/cheatsheet_check.t
```diff
@@ -0,0 +1,104 @@
...
+         # spacing in keys ([a]/[b])'
+         if($entry->{val}){
+             if($entry->{val} =~ /\(\s+\[.*\]\s+\/\s+\[.+\]\s+\)/g){
+                diag("keys ([a]/[b]) shouldn't have white spaces: $entry->{val} from  $name");
+            }
+            diag( "No trailing white space in the value: $entry->{val} from: $name") if $entry->{val} =~ /\s"$/;
+        }
+         if($entry->{key}){
+             if($entry->{key} =~ /\(\s+\[.*\]\s+\/\s+\[.+\]\s+\)/g){
+                diag("keys ([a]/[b]) shouldn't have white spaces: $entry->{key} from  $name");
```
moollaza: I think we should allow spaces because they're easier to read. We could warn when there aren't spaces, but the output *will* have spaces

moollaza:
@jdorweiler looks like the test caught a couple legitimate bugs! I've pushed some updates to master and restarted the test.

moollaza:
@jdorweiler this is pretty good, I think my biggest concern is the organization of the ouput. If it's possible to better group the issues (and visually organize them) on a per IA basis that may help.

Ideally though we should only see 1 IA failing on any given PR so perhaps it's not as much an issue once we get all existing cheat sheets "passing"?

jdorweiler:
@moollaza sounds good.  I see if I can make the output more readable.

t/cheatsheet_check.t
jdorweiler: So we do what spaces then?

mintsoft: @moollaza and when exactly will there not be spaces but the output will have spaces? Confusing!

jdorweiler:
@MariagraziaAlastra is working on the output here https://github.com/duckduckgo/zeroclickinfo-goodies/pull/1612

Issue #1221, opened 2015-06-29
DaysBetween: Improve triggering


moollaza:
I tried a few queries this morning and none of them triggered the IA. It turns out "sept" doesn't not work but "sep" does when specifying the short form for "September".

Also, the only trigger that seems to work right now is "days between". So "number of days between" and "number of days from A to B" don't work, but definitely should!

Let's improve this one :+1: 

------
IA Page: https://duck.co/ia/view/days_between

Bujiraso:
I made a pull request for the extra triggers. Handling the short version of September like you asked is something I'm interested in doing but not sure how to accomplish. It looks like it's being parsed by another library or function, would I have to edit that or should the handling be added in this IA?

mintsoft:
@Bujiraso The date handling is through our shared Role: https://github.com/duckduckgo/zeroclickinfo-goodies/blob/master/lib/DDG/GoodieRole/Dates.pm if it's updated here, then all goodies using it get the same benefit!

Bujiraso:
I'll have a crack at that, too, then!

moollaza:
I just noticed that we also don't support "how many days between" as a valid trigger, we should definitely support that as well!

Bujiraso:
I'll add that trigger next time I have a few minutes and send a pull request for it.

@mintsoft, have we come to a decision as to the changes on the Dates role? I know you said you were taking a look at it in #1294, so let me know if you have any ideas but since it's less of a quick fix than originally thought, I'll not be forging ahead on it on my own -- at least not until the expected functionality is neatly pinned down.

mintsoft:
@Bujiraso not completely as of yet, the conclusion I came to was that relative statements such as "next $x", "last $x", "tomorrow", "yesterday" should always be relative to today, whereas ambiguous statements like "december" should be relative to the first unambiguous date in the set. Something along those lines anyway.

Bujiraso:
@mintsoft, no worries. I also found it needs more pondering than code and just wanted to mention that I'm not actively hacking at it.

moollaza:
This looks to be fixed!

PR #1594, opened 2015-09-18
DaysBetween - Added another trigger

Bujiraso:
As requested in #1221, added the "how many days between" trigger.


mintsoft:
Cool; thanks for the work @Bujiraso

PR #1597, opened 2015-09-22
Password: Fixes #1565

### LINKED ISSUE #1565: Password: Hard to trigger
It would great to expand the triggering for this IA. I couldn't figure out how to trigger it until I found the IA page, which a random user probably wouldn't do.

Some queries I tried:

- generate password
- generate strong password
- generate random password
- password generator
- random password generator
- random strong password
- random password 16 characters
- create random password

-----
IA Page: https://duck.co/ia/view/password

cinlloc:
A simple fix, maybe not sufficiently filtering triggers. Let me know what you think about it... 

-----
IA Page: https://duck.co/ia/view/password

### DIFF
--- a/lib/DDG/Goodie/Password.pm
+++ b/lib/DDG/Goodie/Password.pm
@@ -19,7 +19,7 @@ topics 'cryptography';
 zci answer_type => 'pw';
 zci is_cached   => 0;
 
-triggers start => 'password', 'random password', 'pw', 'random pw', 'pwgen';
+triggers startend => 'password', 'random password', 'generate password', 'generate random password', 'password generator', 'pw', 'random pw', 'pwgen';
 
 use constant MAX_PWD_LENGTH => 64;
 use constant MIN_PWD_LENGTH => 8;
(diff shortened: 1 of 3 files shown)


lib/DDG/Goodie/Password.pm
```diff
@@ -19,7 +19,7 @@ topics 'cryptography';
 zci answer_type => 'pw';
 zci is_cached   => 0;
 
-triggers start => 'password', 'random password', 'pw', 'random pw', 'pwgen';
+triggers startend => 'password', 'random password', 'generate password', 'generate random password', 'password generator', 'pw', 'random pw', 'pwgen';
```
moollaza: @cinlloc we can keep this to `start` unless we expect these words to appear at the end of a mult-word query (which might not be ideal)

lib/DDG/Goodie/Password.pm
```diff
@@ -49,13 +49,11 @@ my $strengths = join('|', keys %pw_strengths);
 
 handle remainder => sub {
 
-    my $query = shift;
-
-    return if ($query && $query !~ /^(?<fw>\d+|$strengths|)\s*(?<sw>\d+|$strengths|)$/i);
```
moollaza: this is still needed to ensure the query matches our expectation. We can't assume that any query that matches our triggers above is a valid query for this IA. For example "password test" now triggers this IA and returns a random password, which isn't good

t/Password.t
```diff
@@ -91,6 +83,87 @@ ddg_goodie_test(
...
+    ),
+    'random password strong 15' => test_zci(
+        qr/.{15} \(random password\)/,
+        structured_answer => {
+            input     => ['15 characters', 'high strength'],
+            operation => 'Random password',
+            result    => qr/^.{15}$/
+        }
+    ),
     'password 65' => undef,
```
moollaza: we should add some more examples that shouldn't trigger such as `password`, `password strength`, etc

cinlloc: Ok for adding non-passing tests, but I think `password` should still trigger IA, shoudn't it?

moollaza:
@cinlloc thanks for updating, apologies for the delay. I've set this up for deployment to Beta and will review it shortly.

We appreciate your patience!

moollaza:
@cinlloc this LGTM :+1:

Issue #1393, opened 2015-08-13
Calculator: shows results when searching for a phone number


jdorweiler:
When looking up a phone number I get a calculator result.  

------
IA Page:  http://duck.co/ia/view/calculator

mintsoft:
@jdorweiler Do you have an example of the format? UK phone numbers don't seem to do it!

jdorweiler:
@mintsoft sorry!  Here's a US phone example https://duckduckgo.com/?q=555-555-5555&t=canonical&ia=answer

wongalvis:
I don't think it's easy to differentiate searching for a phone number and entering a formula in the query.
There could be a suggestion link above the search results "Looking for a calculation?" (or something like that) and expands to display the Goodie when it's clicked.

moollaza:
@wongalvis I think it's OK to bail on queries that match a North American phone number regex like:

```perl
/^\(?\d{3}\)?-\d{3}-\d{4}$/
```

If there's no spaces and 10 digits, its almost certainly a phone number. The user can adjust the query or add "calculate" or `=` to the query to ensure the calculator picks it up.

Not that it matters much, but for comparison, Google and Bing handle it this way. Interestingly Yahoo shows their calculator for queries that match that regex.

jagtalon:
Or you can https://metacpan.org/pod/Number::Phone

PR #1605, opened 2015-09-25
Calculator - Ignore phone numbers

123dj:
Fixes [#1393 ](https://github.com/duckduckgo/zeroclickinfo-goodies/issues/1393).

### DIFF
--- a/lib/DDG/Goodie/Calculator.pm
+++ b/lib/DDG/Goodie/Calculator.pm
@@ -82,7 +82,7 @@ handle query_nowhitespace => sub {
     return if ($query =~ /\b0x/);      # Probably attempt to express a hexadecimal number, query_nowhitespace makes this overreach a bit.
     return if ($query =~ $network);    # Probably want to talk about addresses, not calculations.
     return if ($query =~ qr/(?:(?<pcnt>\d+)%(?<op>(\+|\-|\*|\/))(?<num>\d+)) | (?:(?<num>\d+)(?<op>(\+|\-|\*|\/))(?<pcnt>\d+)%)/);    # Probably want to calculate a percent ( will be used PercentOf )
-    return if ($query =~ /\b([0-9]{1}[-\.\s])?([\(\[]?[0-9]{3}[\)\]]?[-\.\s])?([0-9]{3})[-\.\s]([0-9]{4})(?:\s?(?:x|ext)\s?([0-9])+)?\b/); # Probably are searching for a phone number, not making a calculation
+    return if ($query =~ /^(?:(?:\+?1\s*(?:[.-]\s*)?)?(?:\(\s*([2-9]1[02-9]|[2-9][02-8]1|[2-9][02-8][02-9])\s*\)|([2-9]1[02-9]|[2-9][02-8]1|[2-9][02-8][02-9]))\s*(?:[.-]\s*)?)([2-9]1[02-9]|[2-9][02-9]1|[2-9][02-9]{2})\s*(?:[.-]\s*)?([0-9]{4})(?:\s*(?:#|x\.?|ext\.?|extension)\s*(\d+))?$/); # Probably are searching for a phone number, not making a calculation
 
     $query =~ s/^(?:whatis|calculate|solve|math)//;
 
--- a/lib/DDG/Goodie/Calculator.pm
+++ b/lib/DDG/Goodie/Calculator.pm
@@ -79,9 +79,10 @@ my $network   = qr#^$ip4_regex\s*/\s*(?:$up_to_32|$ip4_regex)\s*$#;         # Lo
 handle query_nowhitespace => sub {
     my $query = $_;
 
-    return if ($query =~ /\b0x/);      # Probable attempt to express a hexadecimal number, query_nowhitespace makes this overreach a bit.
+    return if ($query =~ /\b0x/);      # Probably attempt to express a hexadecimal number, query_nowhitespace makes this overreach a bit.
     return if ($query =~ $network);    # Probably want to talk about addresses, not calculations.
     return if ($query =~ qr/(?:(?<pcnt>\d+)%(?<op>(\+|\-|\*|\/))(?<num>\d+)) | (?:(?<num>\d+)(?<op>(\+|\-|\*|\/))(?<pcnt>\d+)%)/);    # Probably want to calculate a percent ( will be used PercentOf )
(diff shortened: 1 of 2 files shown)


lib/DDG/Goodie/Calculator.pm
```diff
@@ -79,9 +79,10 @@ my $network   = qr#^$ip4_regex\s*/\s*(?:$up_to_32|$ip4_regex)\s*$#;         # Lo
 handle query_nowhitespace => sub {
     my $query = $_;
 
-    return if ($query =~ /\b0x/);      # Probable attempt to express a hexadecimal number, query_nowhitespace makes this overreach a bit.
+    return if ($query =~ /\b0x/);      # Probably attempt to express a hexadecimal number, query_nowhitespace makes this overreach a bit.
     return if ($query =~ $network);    # Probably want to talk about addresses, not calculations.
     return if ($query =~ qr/(?:(?<pcnt>\d+)%(?<op>(\+|\-|\*|\/))(?<num>\d+)) | (?:(?<num>\d+)(?<op>(\+|\-|\*|\/))(?<pcnt>\d+)%)/);    # Probably want to calculate a percent ( will be used PercentOf )
+    return if ($query =~ /\b([0-9]{1}[-\.\s])?([\(\[]?[0-9]{3}[\)\]]?[-\.\s])?([0-9]{3})[-\.\s]([0-9]{4})(?:\s?(?:x|ext)\s?([0-9])+)?\b/); # Probably are searching for a phone number, not making a calculation
```
moollaza: out of curiosity, any reason you used `[0-9]` instead of `\d`?

```perl
/\b(\d[-\.\s])?([\(\[]?\d{3}[\)\]]?[-\.\s])?(\d{3})[-\.\s](\d{4})(?:\s?(?:x|ext)\s?\d+)?\b/); # Probably searching for a phone number, not making a calculation 
```

t/Calculator.t
```diff
@@ -774,6 +774,7 @@ ddg_goodie_test(
     '$5'                              => undef,
     'calculate 5'                     => undef,
     'solve $50'                       => undef,
+    '382-538-2546'                    => undef,    # Calling DuckDuckGo
```
moollaza: Mind adding a few more tests to ensure the other formats your regex captures also produce `undef`

lib/DDG/Goodie/Calculator.pm
123dj: @moollaza Generally, `[0-9]` is safer and more specific than `\d` per [this](http://stackoverflow.com/a/890734) question.  (Real answer: I copied and pasted from [Regexr](http://www.regexr.com/3avas) :smile:.

123dj:
@moollaza What kinds of tests were you thinking about?

mintsoft:
@123dj it'd be other telephone number formats that you'd expect to not get caught, i.e. I notice that the regex supports extensions as well as actual numbers.

123dj:
@mintsoft Made any progress?  If not I can take a stab at this.

mintsoft:
@123dj If you would, that'd be great

123dj:
@mintsoft @moollaza How does that look?

t/Calculator.t
```diff
@@ -774,6 +774,11 @@ ddg_goodie_test(
...
+    '382-5382'                        => undef,
+    '+1-(382)-538-2546'               => undef,
```
mintsoft: @moollaza looking at a few(through my British eyes) of these I'd expect the calculation returned as I don't recognise it as a phone number

123dj:
@mintsoft @moollaza I updated the regex from this [SO question](http://stackoverflow.com/a/123666) and slightly modified it so numbers like `845-1234` are calculated instead of being returned as a phone number.  How does it look?

mintsoft:
@123dj The regex looks fine, it's just the test coverage that's left. Ideally we'd want some pretty broad test coverage here, given how varied telephone numbers can be. For example: +44-1780111111 or +441780-111-111; +44 1780 111-111; 01780-111-111; 01780-111-111x400 etc; can you add tests for as many telephone number styles that you can think of?

123dj:
@mintsoft I added your examples, can't think of any more.  I've never seen many of those formats before :D.

mintsoft:
@123dj those are UK formats, @moollaza do you have some examples of common use US formats?

123dj:
[Here's](https://en.wikipedia.org/wiki/National_conventions_for_writing_telephone_numbers) a page describing different formats.  It's long 0_0.

mintsoft:
Hmm, there are a lot of different examples there. It might be worth adding some tests for those with brackets `(01780) 999 999` and the common US formats?

123dj:
@mintsoft look good?

mintsoft:
@123dj I'm happy if @moollaza is happy!

123dj:
@moollaza look good?

lib/DDG/Goodie/Calculator.pm
```diff
@@ -79,9 +79,10 @@ my $network   = qr#^$ip4_regex\s*/\s*(?:$up_to_32|$ip4_regex)\s*$#;         # Lo
 handle query_nowhitespace => sub {
     my $query = $_;
 
-    return if ($query =~ /\b0x/);      # Probable attempt to express a hexadecimal number, query_nowhitespace makes this overreach a bit.
+    return if ($query =~ /\b0x/);      # Probably attempt to express a hexadecimal number, query_nowhitespace makes this overreach a bit.
     return if ($query =~ $network);    # Probably want to talk about addresses, not calculations.
     return if ($query =~ qr/(?:(?<pcnt>\d+)%(?<op>(\+|\-|\*|\/))(?<num>\d+)) | (?:(?<num>\d+)(?<op>(\+|\-|\*|\/))(?<pcnt>\d+)%)/);    # Probably want to calculate a percent ( will be used PercentOf )
+    return if ($query =~ /^(?:(?:\+?1\s*(?:[.-]\s*)?)?(?:\(\s*([2-9]1[02-9]|[2-9][02-8]1|[2-9][02-8][02-9])\s*\)|([2-9]1[02-9]|[2-9][02-8]1|[2-9][02-8][02-9]))\s*(?:[.-]\s*)?)([2-9]1[02-9]|[2-9][02-9]1|[2-9][02-9]{2})\s*(?:[.-]\s*)?([0-9]{4})(?:\s*(?:#|x\.?|ext\.?|extension)\s*(\d+))?$/); # Probably are searching for a phone number, not making a calculation
```
moollaza: We should really define these regexes outside the handle so we don't need to re-build them each time the calculator runs

moollaza:
@123dj @mintsoft this LGTM :+1:

PR #1623, opened 2015-10-01
fix white space and section order for cheatsheets

jdorweiler:
This adds two new things:
1, Add in white space for keys `([a]/[b])` so that all of these keys show up as `( [a] / [b] )` on the frontend. 
2. Add missing sections to the end of `section_order`.  This handles case mismatches and spelling.

@moollaza @jagtalon

### DIFF
--- a/share/goodie/cheat_sheets/cheat_sheets.js
+++ b/share/goodie/cheat_sheets/cheat_sheets.js
@@ -9,6 +9,13 @@ DDH.cheat_sheets.build = function(ops) {
           path: template_type ? 'DDH.cheat_sheets.' + template_type : 'DDH.cheat_sheets.keyboard'
         };
 
+        // add missing sections in section_order
+        for(var section in sections){
+            if(section_order.indexOf(section) == -1){
+                section_order.push(section);
+            }
+        }
+
         $.each(section_order, function(i, section) {
            if (sections[section]){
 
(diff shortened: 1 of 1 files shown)


share/goodie/cheat_sheets/cheat_sheets.js
```diff
@@ -34,6 +41,9 @@ DDH.cheat_sheets.build = function(ops) {
 
         var out;
         var codeClass = typeof className === "string" ? className : "bg-clr--white";
+
+        // normalize white space for ( [a] / [b] ) keys
+        string = string.replace(/\(\[(.+)\]\/\[(.+)\]\)/, "( [$1] / [$2] )");
```
moollaza: we actually need to be able to account for multiples of `/ [\.+]` e.g. `( [$1] / [$2] / [$3] )`

jdorweiler:
@moollaza  

Updated to handle any multiples of  `[a] / [b] ... ` keys.

Here's how it works.
1. check for `([a]/[b] ..... )`
2. grab the inner portion `[a]/[b]....`
3. split on `/`
4. trim white space for each
4. join with ` / `

GuiltyDolphin:
@jdorweiler @moollaza What's the situation with this?

You've got a case of *merge conflicts!*

moollaza:
@GuiltyDolphin would you be interested in picking this one up? I think all we need from it would be the normalizing of the content, though even that may not be necessary anymore because devs are doing a better job now...

I think we can close for now though because the new tests enforce devs to make sure their sections are correct

GuiltyDolphin:
@moollaza Yeah, I was going to say - the tests should pick up most of this now.

PR #1638, opened 2015-10-05
added language template for cheatsheets, organized languages

iambibhas:
New template
---
Added new template for cheatsheets - `language`. Beside the usual `key` and `val`, this template also accepts `trn` which is used to show the transliterated version of the phrases i.e. the native words in latin/english characters, so that the non-native speakers can also use these language cheatsheets to speak most commonly used phrases. This is how it looks - 

selection_029

Example codes are here - https://github.com/iambibhas/zeroclickinfo-goodies/blob/0e7b7905bf1e0fdcaa57e4fadad7ae6f59d2989d/share/goodie/cheat_sheets/json/language/bengali.json

Reorganization
---
Also moved the language cheatsheets to a separate folders called `language` so they stay separated from rest of the cheatsheets and we can focus on adding more and more language cheatsheets.

I'll create an issue to move the existing language cheatsheets to the `language` template once this gets merged.


share/goodie/cheat_sheets/language.handlebars
```diff
@@ -0,0 +1,9 @@
+<ul>
+  {{#each items}}
+    <li class="text--primary">
+      <span class="cheatsheet__key">{{key}}</span>
+      <span class="cheatsheet__value text--secondary">{{val}}</span>
+      <span class="cheatsheet__translit text--secondary">{{trn}}</span>
```
moollaza: I believe we'll want this to be optional, depending on whether `trn` is defined? (If we're going to use this for all "language" cheat sheets)

I'm not a huge fan of `trn` because it's not as obvious as `key` and `val` but I guess it works for now.

share/goodie/cheat_sheets/cheat_sheets.css
```diff
@@ -102,6 +102,41 @@
...
+.zci--cheat_sheets .language li .cheatsheet__key {
+    display: block;
+    font-weight: bold;
+}
+
+.zci--cheat_sheets .language li .cheatsheet__value {
+  display: block;
+}
+
+.zci--cheat_sheets .language li .cheatsheet__translit {
```
moollaza: I think we set a bad example using `<span>` in the reference template. We should just use `<div>` so we don't need CSS setting it to `display: block`

share/goodie/cheat_sheets/language.handlebars
iambibhas: I also dont like `trn` that much, but just wanted to have some uniformity between the the key names. I first named if `translit`, but seeing that the others are `key` and `val`, renamed it. I'll make it optional.

share/goodie/cheat_sheets/cheat_sheets.css
iambibhas: Yeah, thought so too. Will do that.

iambibhas:
@moollaza should I update the reference template to use `div` too?

iambibhas:
@moollaza seems like forcing the `val` to use `font-family: monospace` renders the non-latin languages properly. Here is a before-after comparison of Bengali. Shows the different in fused letters - 

goodie-font

Although that forces other languages to show up monospaced - 

selection_034

What to do?

moollaza:
@iambibhas how does the site look to you, when you switch the preferred language to Bengali? It seems like things are a little clearer when the font is slightly larger, so perhaps we'll need to bump increase the font size a bit, or reduce the font weight. Let's try that first.

As a last resort, I suppose we can apply the monspace font-face when `trn` is defined, meaning we're dealing with a non latin-based alphabet

AhmedBenmoussa:
Great work @iambibhas. Really great. can't wait to try it.

iambibhas:
> how does the site look to you, when you switch the preferred language to Bengali?

Same problem with fused letters.

selection_036

If I make the fonts monospaced, then only they look proper. Maybe we can try the same with some other language? Maybe that Arabic cheatsheet PR we have pending. Just ask the dev to try it out.

> As a last resort, I suppose we can apply the monspace font-face when trn is defined, meaning we're dealing with a non latin-based alphabet

@moollaza I think this is what we should do.

moollaza:
@iambibhas what browser/OS are you using? Are the font's being rendered with our DDG_Proxima_Nova font? Things look much clearer to me on OSX Firefox, Safari

, Chrome

iambibhas:
@moollaza Firefox and Chromium, Ubuntu. It seems like that Proxima Nova font is being used. But still the fonts are not being rendered properly. Do you have any clue?

moollaza:
@iambibhas I'm not sure what the problem could be. Do you have any other computers/devices you could test on to see if it's just your computer that has the problem?

iambibhas:
@moollaza it works on other devices. Maybe something wrong with my comp somehow. Will figure that out. This is good to go otherwise. Should I make `reference` template use `div` too?

iambibhas:
Fixed the issue on my comp. Turns out the font family `FreeFont` was being used as default for Bengali texts and it didn't handle the conjuncts well. Removed it. Now everything is good. :)

moollaza:
>Fixed the issue on my comp. Turns out the font family FreeFont was being used as default for Bengali texts and it didn't handle the conjuncts well. Removed it. Now everything is good. :)

Awesome! I'm glad to hear you figured this out :+1: 

> Should I make reference template use div too? 

I think it's OK for now. We can review the HTML/CSS for all templates and simplify things if necessary.

@iambibhas I need you to add the actual Bengali Cheat Sheet in a separate PR though, after this one is submitted so the connection to the IA Page will work properly.

moollaza:
@iambibhas ping!

iambibhas:
@moollaza done. will send separate PR for the bengali cheatsheet once this is merged.

moollaza:
@iambibhas Thanks! This LGTM :+1:

Issue #1624, opened 2015-10-01
CheatSheets: Remove leading "$  " from Terminal style Cheat Sheets


moollaza:
It looks like a few Cheat Sheets were accepted during Quack and Hack that have a leading `$ ` before each command to help signify that it's a command being entered at a command prompt.

Based on the styling it should be pretty obvious that these commands are being entered in a terminal. I think we should remove it to make sure all the terminal style cheat sheets maintain the same look. Also, it takes up a few extra character that will help to prevent any lengthy strings from wrapping.

If anyone has strong opinions about this, we'd love to hear them. If we agree unanimously that they all *should* have a the `$ ` then we should probably build it into the template, not the data.

preemeijer:
Hello @moollaza, I want to wait for my svn pull request before I do anything else.

Further more I've scripted this change, just wanted to do that... Sorry about that I just wanted to have some fun and not 'just' remove the dollar signs ;)
Anyone can use this script for less manual change, but be aware: Please check before pull request!
```bash
#!/bin/bash   
fileorg="$1"                                                             
destfile=$fileorg.sed                                                                  
difffile=$fileorg.diff

sed '/key/s/$ //' $fileorg > $destfile
diff $fileorg $destfile > $difffile
diff $fileorg $destfile
```
Copy this script in for example you're codio machine and do the following: ./[scriptname] [original json file]

Give this script the original json file and two files will be made a .sed and a .diff.
The .sed file is the file you could rename to the original json file and the .diff file you can check what have been changed.

Greetings

moollaza:
@preemeijer thanks a lot, that script looks great, I see you used it for SVN. How about running it on all the cheat sheets so we fix all of them?

preemeijer:
Okay, will have a go at it if I can :)

PR #1651, opened 2015-10-09
svn cheatsheet remove of dollar signs

preemeijer:
Regarding #1624 I've made this change to the SVN cheat sheet.
This is my very first time working with branches so hopefully everything works like it should be. (and my scratch cheat sheet isn't lost. Please advise me on the scratch cheat sheet.
Greetings


moollaza:
@preemeijer this looks great! Your commit log is a little messy because you're adding then removing the same file but the overall result is fine :+1:

PR #1652, opened 2015-10-10
Dollarsign removal for tar.json nmap.json gnupg.json coffeescript.json

### LINKED ISSUE #1624: CheatSheets: Remove leading "$  " from Terminal style Cheat Sheets

preemeijer:
Hello @moollaza,

I did a search on all the files on the string "key": "$ and found these files (there are more files I was getting back but as far as I could understand legit use of the $ signs)
This pull request fixes #1624 

Greetings, Paul

### DIFF
--- a/share/goodie/cheat_sheets/json/coffeescript.json
+++ b/share/goodie/cheat_sheets/json/coffeescript.json
@@ -20,22 +20,22 @@
     "sections": {
         "Usage": [{
             "val": "Run CoffeeScript file",
-            "key": "$ coffee src\/file.coffee"
+            "key": "coffee src\/file.coffee"
         }, {
             "val": "Compile files from src to output",
-            "key": "$ coffee -c -o output\/ src\/"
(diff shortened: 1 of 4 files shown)


preemeijer:
Is something not okay with this pull request? Please advice ^_^

moollaza:
@preemeijer this LGTM :+1:

Thanks a lot for contributing (we really appreciate it :smiley:).

PR #1612, opened 2015-09-28
cheatsheet_check.t - show nicely formatted output

MariagraziaAlastra:
@jdorweiler I still need to test this properly, and maybe the bit of code at the end should be a function, but I wanted to start showing you the code.

Screenshot:
new-blank

PR #1668, opened 2015-10-19
CheatSheet Validator

MariagraziaAlastra:
Making a PR to master instead! Continuation of #1612 
Current output: 
new-blank

//cc @jdorweiler @moollaza


t/cheatsheet_check.t
```diff
@@ -0,0 +1,171 @@
...
+    $temp_pass = (ref $sections eq 'HASH')? 1 : 0;
+    push(@tests, {msg => 'sections is a hash of section key/pairs', critical => 1, pass => $temp_pass});
+
+    map{ 
+        $sections->{lc$_} = $sections->{$_}; 
+        delete $sections->{$_};
+    } keys $sections;
+  
+    for my $section_name (@$order) {
+        push(@tests, {msg => 'Expected "$section_name" but not found', critical => 0, pass => 0})  unless $sections->{$section_name};
```
moollaza: we need to use double quotes here, because the single quotes take the whole string literally, hence the `$section_name` isn't recognized as a variable, and is taken literally:
1_____ssh_


You'll need to escape the quote around the variable with `\"`.

moollaza: Or, like we've done below swap the single and double quotes around: `"Expected '$section_name' but not found"`

MariagraziaAlastra:
@moollaza 
quoted

moollaza:
@MariagraziaAlastra Thanks! This LGTM :+1:

I have a couple minor fixes I'll push after:

- remove the newline between each test
- us green coloring, plus OK status for test that only have a WARN

jdorweiler:
:cake:

PR #1674, opened 2015-10-20
Figlet Hotfix

moollaza:
Hotfix for Figlet to plug XSS hole.

Now html encodes text when using one of the 3 plaintext fonts. I decided to only do this for these three fonts so that the output will still look correct in the "bigtext" fonts, like the default Figlet font.

Examples:

Encoded Plaintext
figlet_term__img_src___favicon_ico__onload__alert_1____at_duckduckgo_and_file_finder

Un-Encoded Non-plaintext
figlet__img_src___favicon_ico__onload__alert_1____at_duckduckgo_and_file_finder

/cc @nilnilnil

### DIFF
--- a/lib/DDG/Goodie/FIGlet.pm
+++ b/lib/DDG/Goodie/FIGlet.pm
@@ -32,6 +32,7 @@ closedir DIR;
 # Renders a figlet.
 sub render_figlet {
 	my ($font, $text) = @_;
+	$text = html_enc($text);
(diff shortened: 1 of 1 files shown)


jagtalon:
@moollaza checking!

moollaza:
This one obviously needs to be converted to use templates, but this is a quicker fix.

jagtalon:
@moollaza looks good! :+1:

PR #1681, opened 2015-10-22
added the template type to vi cheat sheet

### LINKED ISSUE #1680: Template type is missing in vi cheat sheet
```template_type``` : ```terminal``` need to added.

sagarhani:
This will fix #1680


moollaza:
@sagarhani I'm not sure the terminal template is appropriate for this IA. Only a few of the commands here are actually entered on a command line. The rest are key-presses. Also, because they're all single keys, they end up taking up much more space than is needed.

moollaza:
If you want, you could add "template_type": "keyboard" to make it explicit?

Issue #1685, opened 2015-10-23
Bitsum: Should also accept binary input


moollaza:
This IA already accepts base10 and hex numbers as input, and converts them to binary. We might as well also accept binary inputs!

------
IA Page:  http://duck.co/ia/view/bitsum

ozdemirburak:
@moollaza I guess that this one should be closed.

PR #1693, opened 2015-10-26
Bitsum: Accept binary input

kbhat95:
This is in reference to #1685 .
Hope this helps!


moollaza:
@kbhat95 this LGTM :+1:

Thanks a lot for contributing (we really appreciate it :smiley:).

lib/DDG/Goodie/Bitsum.pm
```diff
@@ -28,7 +28,7 @@ triggers start => "bitsum", "hammingweight", "hw",
 handle remainder => sub {
     
     # Return if input is no hex or decimal number
```
moollaza: nit: should probably update this comment as well

PR #1703, opened 2015-10-28
Atom cheat sheet minor fixes

codenirvana:
- Changed Description to official [atom](http://atom.io) description
- Added aliases
- Added optional template_type
- Fixed some typo

PR #1705, opened 2015-10-28
#1703 Removed unnecessary alias

codenirvana:
@moollaza this will fix #1703
You was right ```atom``` alias is unnecessary, it works without that.


moollaza:
@codenirvana this LGTM :+1:

PR #1707, opened 2015-10-28
Is website up or down Goodie 

codenirvana:
Check if given URL is onine or not

Provides the status of URL user looking for.

"isup duckduckgo.com", "is up https://duckduckgo.com",  "online duckduckgo.com", "github.com is down"

Geeks, computing_info, Web Dev

Yes, [Check if a site is online](https://duck.co/ideas/idea/6602/check-if-a-site-is-online)

This will overlap !isup bang

Yes I want to add multiple trigger like this- :start-trigger: query :end-trigger:

screenshot

### DIFF
--- a/lib/DDG/Goodie/ISUP.pm
+++ b/lib/DDG/Goodie/ISUP.pm
@@ -1,0 +1,54 @@
+package DDG::Goodie::ISUP;
+
+use DDG::Goodie;
+use LWP::Simple;
+use strict;
+
+zci answer_type => "isup";
+zci is_cached   => 1;
+
+name "ISUP";
+description "Check if website is onine or not";
+primary_example_queries "isup duckduckgo.com", "isup https://duckduckgo.com";
+secondary_example_queries "online duckduckgo.com";
+topics 'computing';
+category 'computing_info';
+code_url "https://github.com/duckduckgo/zeroclickinfo-goodies/blob/master/lib/DDG/Goodie/Isup.pm";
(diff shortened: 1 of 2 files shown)


mintsoft:
@codenirvana Hello! Thanks for the PR, unfortunately Goodies cannot perfom HTTP requests. If there's an API available then a spice could be used instead. That said, there is https://github.com/duckduckgo/zeroclickinfo-spice/blob/master/lib/DDG/Spice/IsItUp.pm already?

moollaza:
@codenirvana please make sure you share your idea with us before diving into it so we can ensure you don't spend your time working on something we can't accept!

codenirvana:
@moollaza I made this because its listed in Instant Answer Ideas and its not flagged duplicate and under development.

codenirvana:
@mintsoft seems like that spice is not working...
I searched a lot but its not giving any instant answer.

moollaza:
@codenirvana @mintsoft looks like the API switched to HTTPS so the IA isn't working right now. I pushed an update an it will be fixed with the next release.

> I made this because its listed in Instant Answer Ideas 

to clarify (and it's not your fault, because we haven't made it very clear) ideas marked as `new` haven't been approved by us yet. Once the idea's status is changed to "Needs a developer" it means we've approved it and it is now claimable. You'll see the "claim this idea" button beside the IA and clicking it will claim it, let everyone know you've claimed it, and also generate an IA Page for the idea.

> and its not flagged duplicate and under development.

Sorry about that. We need to do a better job of curating the ideas so this doesn't happen in the future.

Issue #1695, opened 2015-10-26
CheatSheets: CheatSheets.pm should force lowercasing of aliases to ensure they always work


moollaza:
We've updated the documentation to tell developers that they should only use lowercase letters, but we should also ensure that the `getAliases` function forces the aliases to be lower-cased so that everything still works and developers don't get frustrated.

We should probably print a warning if an uppercase alias is detected. We already do this in DuckPAN for any Spice/Goodies that have uppercase trigger words defined.

Here's the line that needs to be fixed: https://github.com/duckduckgo/zeroclickinfo-goodies/blob/master/lib/DDG/Goodie/CheatSheets.pm#L49

nemeth:
I made changes in  #1710.
Is [this the warning](https://github.com/duckduckgo/p5-app-duckpan/blob/master/lib/App/DuckPAN/DDG.pm#L97) what you mean, @moollaza?

moollaza:
@nemeth yup, that's the warning I was referring to :+1:

moollaza:
I forgot that we were working on https://github.com/duckduckgo/zeroclickinfo-goodies/pull/1668. This is probably the best place to put in a warning for uppercase aliases (also, uppercases in the keys like `metadata`, `sections`, etc.

nemeth:
oh, thanks to point that PR! I'll take a look at it

GuiltyDolphin:
@moollaza #2281 Has a test for uppercase characters in aliases; feel free to comment any other test suggestions there.

GuiltyDolphin:
@moollaza Fixed in #2281 as the test was made critical - so aliases can no longer contain uppercase characters.

PR #1710, opened 2015-10-29
#1695 - Lowercase cheetsheat aliases

nemeth:
This small change lets trigger cheatsheet aliases that contain uppercase letters in .json files.


moollaza:
@nemeth this LGTM :+1:

PR #1714, opened 2015-10-29
explicitly mentioning the template_type for i3 cheat sheet

sagarhani:


moollaza:
@sagarhani thanks. For the future, you can group these similar changes into a singular PR. That's totally fine. No need to submit several 1-line changes as separate PR's ;)

moollaza:
This LGTM :+1:

PR #1755, opened 2015-11-11
Remove metadata from template

zachthompson:
* Remove metadata from module template
* Clean up the syntax


moollaza:
Thanks! This LGTM :+1:

PR #1759, opened 2015-11-13
CheatSheets: Don't add the .sep--after class if there is no sourceUrl.

jagtalon:
We're doing this because we don't want to add a separator if we don't have any "More at ..." link.

Before:
screen shot 2015-11-13 at 9 43 50 am

After:
screen shot 2015-11-13 at 9 43 36 am

giphy

### DIFF
--- a/share/goodie/cheat_sheets/detail.handlebars
+++ b/share/goodie/cheat_sheets/detail.handlebars
@@ -11,7 +11,7 @@
 </div>
 
 <div class="cheatsheet__footer c-list__links">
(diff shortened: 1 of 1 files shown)


jagtalon:
The `moreAt` helper does the same check for bailing: https://github.com/duckduckgo/duckduckgo-template-helpers/blob/17744309035a316489be9a90ec8790714216f5f8/template_helpers.js#L576

moollaza:
@jagtalon Thanks! This LGTM :+1:

PR #1764, opened 2015-11-15
New Music Scale Goodie

sublinear:

Generates musical scales

The set of queries that this goodie responds to currently get results from Wikipedia or piano and guitar websites that are annoying to navigate if all you want is a simple list of notes

Musicians and educators

Wikipedia Answers currently responds to the queries that trigger this goodie.

-----
IA Page: https://duck.co/ia/view/music_scale

### DIFF
--- a/lib/DDG/Goodie/MusicScale.pm
+++ b/lib/DDG/Goodie/MusicScale.pm
@@ -1,0 +1,78 @@
+package DDG::Goodie::MusicScale;
+# ABSTRACT: Return the diatonic scale of the key and mode given in the query
+
+use DDG::Goodie;
+use strict;
+
+zci answer_type => "music_scale";
+zci is_cached   => 1;
+
+name "MusicScale";
+description "Generate the notes of the queried musical scale";
+primary_example_queries "c major scale", "f# minor scale";
+category "reference";
+topics "music";
+code_url "https://github.com/duckduckgo/zeroclickinfo-goodies/blob/master/lib/DDG/Goodie/MusicScale.pm";
+attribution github => ["sublinear", "sublinear"];
+
+# Triggers
+triggers end => "major scale", "ionian scale", "dorian scale", "phrygian scale", "lydian scale", "mixolydian scale", "minor scale", "aeolian scale", "locrian scale";
+
+# Handle statement
+handle sub {
+    return unless $_;
+
+    # must be a note letter, optional sharp or flat, and the mode name followed by the word "scale"
+    # e.g. "Gb Dorian Scale"
+
+    return unless ($_ =~ /^([A-Ga-g])([b#])?\s+([mM]ajor|[dD]orian|[pP]hrygian|[lL]ydian|[mM]ixolydian|[mM]inor|[lL]ocrian)\s+([sS]cale)$/ );
+
+    my( $letter, $accidental, $mode, $noteIndex, $modeIndex, @scale, @majorScale, @notes, @modes );
+
+    # regex captures
+    if (defined $1) { $letter     = uc($1);      } else { $letter     = ""; }
+    if (defined $2) { $accidental = $2;          } else { $accidental = ""; }
+    if (defined $3) { $mode       = ucfirst($3); } else { $mode       = ""; }
+    
+    # lookup arrays
+    @majorScale = (0, 2, 4, 5, 7, 9, 11);
+    @notes = ("C", "Db/C#", "D", "Eb/D#", "E", "F", "Gb/F#", "G", "Ab/G#", "A", "Bb/A#", "B");
+    @modes = ("Major", "Dorian", "Phrygian", "Lydian", "Mixolydian", "Minor", "Locrian");
+
+    # get note index
+    if    ( $letter eq "C" ) { $noteIndex = 0;  }
+    elsif ( $letter eq "D" ) { $noteIndex = 2;  }
+    elsif ( $letter eq "E" ) { $noteIndex = 4;  }
+    elsif ( $letter eq "F" ) { $noteIndex = 5;  }
+    elsif ( $letter eq "G" ) { $noteIndex = 7;  }
+    elsif ( $letter eq "A" ) { $noteIndex = 9;  }
+    else                     { $noteIndex = 11; }
+
+    # apply accidental to note index
+    if    ( $accidental eq "b" ) { $noteIndex -= 1; }
+    elsif ( $accidental eq "#" ) { $noteIndex += 1; }
+    $noteIndex %= 12;
+
+    # get mode index
+    for (my $i=0; $i < $#modes; $i++) {
+        if ($mode eq $modes[$i]) {
+            $modeIndex = $i;
+            last;
+        }
+    }
+
+    # generate the scale
+    for (my $i=0; $i < 7; $i++) {
+        push @scale, $notes[ ($noteIndex + (($majorScale[($i + $modeIndex) % 7] - $majorScale[$modeIndex]) % 12)) % 12 ];
+    }
+
+    # result
+    return join(", ", @scale),
+        structured_answer => {
+            input => [html_enc($letter.$accidental." ".$mode." Scale")],
+            operation => "Musical Scale",
+            result => html_enc(join(", ", @scale)),
+        };
+};
+
+1;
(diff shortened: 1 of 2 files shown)


daxtheduck:
@sublinear Hey!
        
Thanks!

moollaza:
@sublinear so sorry for the delay here!  This one slipped through the crack. It's now live on Beta though for testing: https://beta.duckduckgo.com/?q=c+minor+scale+&ia=answer

lib/DDG/Goodie/MusicScale.pm
```diff
@@ -0,0 +1,78 @@
...
+category "reference";
+topics "music";
+code_url "https://github.com/duckduckgo/zeroclickinfo-goodies/blob/master/lib/DDG/Goodie/MusicScale.pm";
+attribution github => ["sublinear", "sublinear"];
+
+# Triggers
+triggers end => "major scale", "dorian scale", "phrygian scale", "lydian scale", "mixolydian scale", "minor scale", "locrian scale";
+
+# Handle statement
+handle sub {
```
moollaza: should probably `handle query_lc` then we can simplify the regex below

lib/DDG/Goodie/MusicScale.pm
```diff
@@ -0,0 +1,78 @@
...
+    return unless $_;
+
+    # must be a note letter, optional sharp or flat, and the mode name followed by the word "scale"
+    # e.g. "Gb Dorian Scale"
+
+    return unless ($_ =~ /^([A-Ga-g])([b#])?\s+([mM]ajor|[dD]orian|[pP]hrygian|[lL]ydian|[mM]ixolydian|[mM]inor|[lL]ocrian)\s+([sS]cale)$/ );
```
moollaza: I think we could relax this a little by allowing for some prefix and suffix phrases:

- "notes of the X scale"
- "what are the notes of the X scale"
- "X scale notes"

moollaza: we can also remove the uppercase letter once we switch to `query_lc` above. FWIW you can also add the `i` flag to the end of the regex to make it case insensitive: e.g. `m/ignore case/i`

lib/DDG/Goodie/MusicScale.pm
```diff
@@ -0,0 +1,78 @@
...
+topics "music";
+code_url "https://github.com/duckduckgo/zeroclickinfo-goodies/blob/master/lib/DDG/Goodie/MusicScale.pm";
+attribution github => ["sublinear", "sublinear"];
+
+# Triggers
+triggers end => "major scale", "dorian scale", "phrygian scale", "lydian scale", "mixolydian scale", "minor scale", "locrian scale";
+
+# Handle statement
+handle sub {
+    return unless $_;
```
moollaza: seeing as we're handling the `query` and we know it contains one of our triggers, this will always be true, so we can remove it

moollaza: if we switch to `handle remainder_lc` though then we'd want to check `$_`. 

You can always access the original query inside the `handle` via `$req->query` or `$req->query_lc` or `$req->query_clean`, etc

lib/DDG/Goodie/MusicScale.pm
moollaza: note: it would be cool to support `♭` in the regex! If you add `use utf8` to the top, it should work...

lib/DDG/Goodie/MusicScale.pm
```diff
@@ -0,0 +1,78 @@
...
+
+    my( $letter, $accidental, $mode, $noteIndex, $modeIndex, @scale, @majorScale, @notes, @modes );
+
+    # regex captures
+    if (defined $1) { $letter     = uc($1);      } else { $letter     = ""; }
+    if (defined $2) { $accidental = $2;          } else { $accidental = ""; }
+    if (defined $3) { $mode       = ucfirst($3); } else { $mode       = ""; }
+    
+    # lookup arrays
+    @majorScale = (0, 2, 4, 5, 7, 9, 11);
```
moollaza: Seeing as these are constants, let's create these outside `handle` then we won't re-create them each time our handler executes.

lib/DDG/Goodie/MusicScale.pm
moollaza: you could also replace `♭` with `b` to keep the regex as-is?

daxtheduck:
Automated screenshot from [beta.duckduckgo.com](https://beta.duckduckgo.com/?q=C major scale)
                
                [Mobile screenshot](https://ia-screenshots.s3.amazonaws.com/music_scale_mobile.png)

daxtheduck:
Automated screenshot from [beta.duckduckgo.com](https://beta.duckduckgo.com/?q=C major scale)
                
                [Mobile screenshot](https://ia-screenshots.s3.amazonaws.com/music_scale_mobile.png)

lib/DDG/Goodie/MusicScale.pm
mintsoft: Also, could we expand it out to handle; diminished, flat-13, major-7, minor-7 etc?

sublinear:
All very good feedback. Thanks. I will be implementing those changes soon. This one got away from me too!

moollaza:
@b1ake I believe you mentioned that there were some errors when I demoed this? Can you try some queries and verify the scales are all correct?

This one looks a little odd: https://beta.duckduckgo.com/?q=C%23+major+scale&ia=answer

sublinear:
> @b1ake I believe you mentioned that there were some errors when I demoed this? Can you try some queries and verify the scales are all correct?
> 
> This one looks a little odd: https://beta.duckduckgo.com/?q=C%23+major+scale&ia=answer

F is enharmonically equivalent to E#, so it is technically the correct set of pitches (in western music). The script could take the accidental of the key into account when displaying the scale and then we could also avoid always outputting the equivalents separated by slashes e.g. "Db/C#" and just show them all as sharps in this case. Maybe we should make this a display option with a checkbox? Does the API allow HTML for some kind of form input?

mintsoft:
@moollaza I see nothing wrong with C# major scale? I'd usually write it out as : `C# Eb F F# G# Bb C` however that's more of a personal style thing

@sublinear you can add JS to the template and interact with the output: see https://github.com/duckduckgo/zeroclickinfo-goodies/tree/master/share/goodie/game2048 for an example.

b1ake:
@moollaza Each scale has a correct spelling, where the notes are written so that there's exactly one instance of each letter (A B C D E F G).

For C# Major, the correct spelling is C♯, D♯, E♯, F♯, G♯, A♯, and B♯. One reason this is helpful is when writing music on a musical staff - there's one note for every line and every space on the staff, since each line and space is assigned to one of the musical letters.

For example, in C# Major, we spell the 3rd note of the scale as 'E#' instead of 'F', because the letter F is already used by the 'F#'.

For composers and musicians with more music theory experience, it looks a little weird / incorrect to spell the scale another way.

It appears that Wikipedia has the correct spelling for every scale, so that's where I would look to fix the IA. For example, here's the article on C# major, which includes the correct spelling right at the top:
https://en.wikipedia.org/wiki/C-sharp_major

moollaza:
@sublinear @mintsoft what do you think about what Blake says? I like the idea of going with what Wiki shows seems to be more "standard".

@sublinear we could definitely have a V2 with some interaction or alternate displays. For now let's get this merged and live!

tsurai:
How about showing the scale "formula" when not specifing a key? So that "Major scale" would show something like "w, w, h, w, w, w, h" for whole steps and half steps. That way you can easily find out whats behind a given scale.

And it would be cool if you could add the major and minor pentatonic scales as they are widly used in blues for example.

moollaza:
@tsurai thanks for the input!

@sublinear @mintsoft @b1ake -- anyone have any thoughts on how we should move forward here?

mintsoft:
I agree with @b1ake the canoncial form would be ideal if possible

jdorweiler:
This pull request is being closed due to inactivity but, if anyone would like to continue it, please @ mention 'duckduckhack-contributors' for help and http://docs.duckduckhack.com/submitting/submitting-overview.html for reference.

PR #1771, opened 2015-11-16
Goodie Template Updates

moollaza:
- Update the Goodie Perl template to suggest structured answers w/ full templating.
    - If we plan to continue supporting the simplified templating, we could use that as the default?
- Added a JS skeleton for a Goodie, adapted from Russell's "Game of Life"
    - This won't be used until the DuckPAN template code is updated to use the file

/cc @zachthompson

### DIFF
--- a/template/lib/DDG/Goodie/Example.pm
+++ b/template/lib/DDG/Goodie/Example.pm
@@ -8,34 +8,54 @@ use DDG::Goodie;
 use strict;
 
 zci answer_type => '<: $lia_name :>';
+
+# Caching - https://duck.co/duckduckhack/spice_advanced_backend#caching-api-responses
 zci is_cached   => 1;
 
-name '<: $lia_name :>';
-
-# Triggers
+# Triggers - https://duck.co/duckduckhack/goodie_triggers
 triggers any => 'triggerWord', 'trigger phrase';
 
 # Handle statement
 handle remainder => sub {
 
+    my $remainder = $_;
+
     # Optional - Guard against no remainder
-    # i.e. the query is only 'triggerWord' or 'trigger phrase'
+    # I.E. the query is only 'triggerWord' or 'trigger phrase'
     #
     # return unless $remainder;
 
     # Optional - Regular expression guard
-    # use this approach to ensure the remainder matches a pattern
-    # i.e. it only contains letters, or numbers, or contains certain words
+    # Use this approach to ensure the remainder matches a pattern
+    # I.E. it only contains letters, or numbers, or contains certain words
     #
-    # return unless qr/^\w+/;
+    # return unless qr/^\w+|\d{5}$/;
 
     return "plain text response",
         structured_answer => {
-                id => '<: $lia_id :>', # Should be an existing Instant Answer topic
-                name => 'Answerbar Tab Name', # Should be an existing Instant Answer topic
-                data => { q => $_ },
-                templates => 1
-            };
+
+            # ID - Must be unique and match Instant Answer page
+            # E.g. https://duck.co/ia/view/calculator has `id => 'calculator'``
+            id => '<: $lia_id :>',
+
+            # Name - Used for Answer Bar Tab
+            # Value should be chosen from existing Instant Answer topics
+            # see https://duck.co/duckduckhack/display_reference#codenamecode-emstringem-required
+            name => 'Answer',
+
+            data => {
+              title => "My Instant Answer Title",
+              subtitle => "My Subtitle",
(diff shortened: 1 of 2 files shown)


template/lib/DDG/Goodie/Example.pm
```diff
@@ -9,14 +10,33 @@ use strict;
 zci answer_type => '<: $lia_name :>';
 zci is_cached   => 1;
 
+name '<: $lia_name :>';
```
zachthompson: Not sure I understand why we are re-adding this?

template/share/goodie/example/example.js
```diff
@@ -0,0 +1,61 @@
...
+(function(env) {
+
+    console.log("DDH.'<: $lia_id :>'.build");
+
+    // define private variables and functions here
+
+    DDH.'<: $lia_id :>'.build = function(ops) {
+
+        return {
+            id: "life",
```
zachthompson: Can this be set by ops.id or does it need to be hardcoded?

template/lib/DDG/Goodie/Example.pm
```diff
@@ -9,14 +10,33 @@ use strict;
...
+
+    # Optional - Regular expression guard
+    # use this approach to ensure the remainder matches a pattern
+    # i.e. it only contains letters, or numbers, or contains certain words
+    #
+    # return unless qr/^\w+/;
+
+    return "plain text response",
+        structured_answer => {
+                id => '<: $lia_id :>', # Should be an existing Instant Answer topic
```
zachthompson: Comment is not quite right.

template/lib/DDG/Goodie/Example.pm
```diff
@@ -9,14 +10,33 @@ use strict;
...
+                name => 'Answerbar Tab Name', # Should be an existing Instant Answer topic
+                data => { q => $_ },
+                templates => 1
```
zachthompson: Shouldn't we be doing...?
``` Perl
templates => {
}
```

template/lib/DDG/Goodie/Example.pm
```diff
@@ -9,14 +10,33 @@ use strict;
...
+    # Optional - Regular expression guard
+    # use this approach to ensure the remainder matches a pattern
+    # i.e. it only contains letters, or numbers, or contains certain words
+    #
+    # return unless qr/^\w+/;
+
+    return "plain text response",
+        structured_answer => {
+                id => '<: $lia_id :>', # Should be an existing Instant Answer topic
+                name => 'Answerbar Tab Name', # Should be an existing Instant Answer topic
```
zachthompson: Comment not right here either

template/lib/DDG/Goodie/Example.pm
moollaza: Russell had it in his code, I figured it was there for a reason, but I'll remove it.

template/lib/DDG/Goodie/Example.pm
moollaza: Crap, just realized I forgot to push my last commit -- I had fixed that!

moollaza:
Looks like all my commits didn't make it to the PR. Sorry, will update :disappointed:

moollaza:
@zachthompson alright, this looks much better now :)

More in-line with our current Spice template. I'll submit some updates to Spice too

template/lib/DDG/Goodie/Example.pm
```diff
@@ -7,16 +8,55 @@ use DDG::Goodie;
...
+
+            templates => {
+                group => "text",
+                # options => {
+                #
+                # }
+            }
+        };
+    };
 };
```
zachthompson: Extra brace here.

moollaza: Whoops. Fixed :+1:

zachthompson:
Thank you.

srvsh:
@moollaza the variable `lia_id` is used in [`Example.pm`](https://github.com/duckduckgo/zeroclickinfo-goodies/blob/master/template/lib/DDG/Goodie/Example.pm#L39) and [`example.js`](https://github.com/duckduckgo/zeroclickinfo-goodies/blob/master/template/share/goodie/example/example.js), but there's no corresponding variable in [`New.pm`](https://github.com/duckduckgo/p5-app-duckpan/blob/master/lib/App/DuckPAN/Cmd/New.pm#L63-L69). Would that be an issue?

As of now, on the generated files have an empty string where the `lia_id` is present on my computer.

zachthompson:
That's a bug @srvsh.

PR #1772, opened 2015-11-17
Disable LWP SSL host verification

zachthompson:
So we can use HTTPS with S3 and not fail travis tests, disable host verification for travis only.


moollaza:
LGTM :+1:

Issue #1491, opened 2015-08-31
Chess960: Remove clickable link from result


moollaza:
Currently the result is being wrapped in an anchor tag, with an `href="#"`. This should be removed.

Really, this Instant Answer should be converted to a JS powered Goodie, that uses Handlebars templates so we're not producing any HTML from the Perl.

---

IA Page:  http://duck.co/ia/view/chess960

moollaza:
Fixed

PR #1791, opened 2015-11-21
Issue #1491 - Chess960: Remove clickable link from result

Mailkov:
I'm a chess player so I could not resist to fix this ... 
I removed href="#" ....
I'm interested converting it into a JS powered Goodie!

-----
IA Page: https://duck.co/ia/view/chess960


mintsoft:
Thanks for the fix, can you actually change it from `<a />` to `<span` instead? An a without an href isn't valid HTML

Mailkov:
@mintsoft I think is ok now

moollaza:
@Mailkov I saw in your other PR that you're interested in converting this to JS? If so you're welcome to do that :+1: 

As long as we can get rid of all these Goodies returning HTML :smiley: -- They should all be using templates now with Handlebars, JS and CSS if needed.

The 2048Game, ScreenResolution, BPMToMs and Minecraft Goodies are all examples of Goodies using templates. 2048 is the most JS + CSS heavy

Mailkov:
@moollaza 
the only problem is that the attribution--link is not positioned correctly... I don't understand why.
Can you help me? Thanks.

moollaza:
> the only problem is that the attribution--link is not positioned correctly... I don't understand why.
Can you help me? Thanks.

That's actually just a DuckPAN bug -- it will be fine in production :+1:

lib/DDG/GoodieRole/Chess.pm
```diff
@@ -76,8 +76,8 @@ sub draw_chessboard_html {
         for ($j = 0; $j < 8; $j++){
             # Columns
             $html_chessboard .= '<td id="'.$arr[$j].(8-$i).'">';
-            $html_chessboard .= '<a href="#" class="'.$class_dict{$position[$counter]};
-            $html_chessboard .= '">'.$unicode_dict{$position[$counter]}.'</a>';
+            $html_chessboard .= '<span class="'.$class_dict{$position[$counter]};
```
moollaza: Does anything else still use this file? If not, would be great to remove it, or convert anything else using this to use JS + Templates

share/goodie/chess960/content.handlebars
```diff
@@ -0,0 +1,15 @@
+<div class="zci--fenviewer">
```
moollaza: we actually already wrap the content in a div with this class (and some other special stuff). Let's remove this one and use the existing one on the page.

share/goodie/chess960/chess960.js
```diff
@@ -0,0 +1,94 @@
+DDH.chess960 = DDH.chess960 || {};
+
+DDH.chess960.build = function(ops) {
```
moollaza: We have made a small change to the formatting here and now wrap these in a closure. That way you can define your helper functions outside the callback, and you won't need to worry about altering the global namespace:

```js
DDH.chess960 = DDH.chess960 || {};

(function(DDH) {
    "use strict";

    // define private variables and functions here

    DDH.chess960.build = function(ops) {

        return {

            // Function that executes after template content is displayed
            onShow: function() {

            }
        };
    };
})(DDH);
```

lib/DDG/Goodie/Chess960.pm
```diff
@@ -136,13 +129,29 @@ handle query => sub {
...
-    
-    my $html = draw_chessboard_html(@fen);
+
     $query =~ s/^ chess960|chess960 $|chess960 //i;
-    return draw_chessboard_ascii(@fen), html => $html, heading => "Position $position_num (Chess960)";
+    
+    return 'Chess 960',
+    structured_answer => {
+        id => 'chess960',
+        name => 960,
```
moollaza: Let's keep this as `Answer` or `Games` -- the value should always be `Answer` or one of our Instant Answer topics (like Games, Entertainment, Software, etc) they can be seen in the docs and also here: https://duck.co/ia

lib/DDG/GoodieRole/Chess.pm
Mailkov: FenViewer uses this ... We can convert it after we've completed this, if you agree

moollaza: Ah I see FenViewer does :+1:

Mailkov:
@mollaza ping
This is ready for review

jagtalon:
@Mailkov checking!

daxtheduck:
Automated screenshot from [beta.duckduckgo.com](https://beta.duckduckgo.com/?q=chess960 random)
                
                [Mobile screenshot](https://ia-screenshots.s3.amazonaws.com/chess960_mobile.png)

moollaza:

share/goodie/chess960/chess960.js
```diff
@@ -0,0 +1,94 @@
...
+        var scheme = [];
+        var letters = ["H", "G", "F", "E", "D", "C", "B", "A"];
+        var position = ops.data.position;
+    
+        for (var i = 0; i < position.length; i++) {
+            scheme[7 - i] = position.substring(i, i + 1);
+        }
+
+        // This function creates and prints on page chess board
+        function start() {
```
moollaza: Ah, my bad: we should move these helper functions outside the def'n of the callback:

```js
(function(DDH) {
    "use strict":
    // private functions here

    DDH.chess960.build = function(ops) {
        ...
```

share/goodie/chess960/chess960.js
```diff
@@ -0,0 +1,94 @@
...
+            for (var i = 0; i < 8; i++) {
+                codepiece = getCodePiece(i);
+                $("#" + letters[i] + "8 span").html("&#98" + (codepiece + 6) + ";"); 
+                $("#" + letters[i] + "1 span").html("&#98" + codepiece + ";");
+            }
+        }
+    
+        function getCodePiece(i) {
+            var code;
+            switch (scheme[i]) {
```
moollaza: If we reliably know that this is always going to be one of our cases, we should make an object with this mapping and do a lookup:

```js
var ob = { "R" : 14, "N": 16, ... };
my code = obj[ scheme[i] ];
```

share/goodie/fen_viewer/fen_viewer.css
```diff
@@ -1,4 +1,4 @@
-.zci--answer .zci--fenviewer a {
+.zci--answer .zci--fenviewer span {
 	color:#333;
```
moollaza: If you'd like, we can use our builtin color classes and background classes. Then in the dark theme, everything will be swapped to look dark, but still work with the colors :)

PR #1803, opened 2015-11-23
Wrap yaml in quotes to force string type

moollaza:
- The values I've wrapped in quotes weren't being interpreted as strings causing them to show as "HASH" on the SERP. This fixes that.

/cc @Snapy -- Are you still interested in converting this to a template? It could look very similar to the OpenNIC Spice: 

https://beta.duckduckgo.com/?q=opennic&ia=dns


moollaza:
Result:
public_dns_at_duckduckgo_and_ _duckduckgo_team_-_asana

Snapy:
First of all thanks for having merged my PR and having corrected the few issues left (sorry for that) and keeping me informed of the evolution of this goodie.
Second, the OpenNIC IA was what I had in mind about location-based DNS servers (except that there is no API in this case). This IA has a small although unimportant issue : you can no longer choose the number of DNS servers.
Third, about converting to the new template I am inexperienced with CSS therefore it would take time.
So contact me if you decide to create a PR and I will gladly help you out.

moollaza:
>First of all thanks for having merged my PR and having corrected the few issues left (sorry for that) and keeping me informed of the evolution of this goodie.

You're welcome, it's my pleasure :+1: 

>Second, the OpenNIC IA was what I had in mind about location-based DNS servers (except that there is no API in this case). This IA has a small although unimportant issue : you can no longer choose the number of DNS servers.

I'm not sure I understand what you mean by "*you can no longer choose the number of DNS servers.*"

>Third, about converting to the new template I am inexperienced with CSS therefore it would take time.
So contact me if you decide to create a PR and I will gladly help you out.

Actually you shouldn't need much, if any CSS. Basically you'd be updating the Perl, and writing a *very simple*, HTML template. It would be nearly identical to the HTML currently being used to produce the rows of the current table. It could be adapted from the template used by OpenNIC: https://github.com/duckduckgo/zeroclickinfo-spice/blob/master/share/spice/open_nic/content.handlebars

I actually made a video tutorial last week that walks through the creation of a Goodie that uses the List template: https://vimeo.com/146960913

If this sound like something you'd like to try let me know and I'd be glad to help. If not I'll create a task for the community to do so

PR #1812, opened 2015-11-25
Update drush.json

jophab:
added alias


moollaza:
@jophab for future reference its OK to put several minor changes (like this) into a single PR.

If they're large changes, or new Instant Answers they should be separated.

This LGTM :+1:

PR #1817, opened 2015-11-25
Small Template Update

moollaza:
This PR removes some extra single quotes that don't belong, and comments out the templates block in the JS to prevent it from overwriting the Perl by default

/cc @zachthompson


zachthompson:
ok!

PR #1824, opened 2015-11-27
Update cmus.json

jophab:
updated name,desc and section names


mintsoft:
Thanks for the tidyup @jophab

PR #1830, opened 2015-11-29
Update ed.json

jophab:
updated name ,desc ,source name and section names


mintsoft:
Thanks for the tweaks @jophab

PR #1833, opened 2015-11-30
Kabaddi cheat sheet: Added 'kabaddi' as an alias

tagawa:
This is the only kabaddi Instant Answer at the moment so I think it's OK to have a general "kabaddi" trigger. If there's another one in future then we could consider returning to more precise triggers.


moollaza:
Thanks! This LGTM :+1:

PR #1478, opened 2015-08-29
lights0123's Security Addons Goodie

lights0123:
**What does your Instant Answer do?**

Provides links to common browser's add-on pages with the search key "anonymous"

**What problem does your Instant Answer solve (Why is it better than organic links)?**

It solves the problem of having to go to an add-on store (i.e. about:addons) and searching for add-ons

**What is the data source for your Instant Answer? (Provide a link if possible)**

https://addons.opera.com/en/search/?query=anonymous&type=extensions
https://addons.mozilla.org/en-US/firefox/tag/anonymous
https://extensions.apple.com/?q=anonymous
https://chrome.google.com/webstore/search/anonymous

**Why did you choose this data source?**

They are the official add-on store of each browser.

**Are there any other alternative (better) data sources?**

Not that I know of.

**What are some example queries that trigger this Instant Answer?**

how to browse the web anonymously
search privately
search anonymously
how to browse the web privately

how to browse the web privately firefox
firefox how to browse the web privately

google chrome how to browse the web privately
opera how to browse the web privately
safari how to browse the web privately

**Which communities will this Instant Answer be especially useful for? (gamers, book lovers, etc)**

People who wish to have enhanced privacy

**Is this Instant Answer connected to a DuckDuckHack [Instant Answer idea](https://duck.co/ideas)?**

https://duck.co/ideas/idea/6201/firefox-add-ons

**Which existing Instant Answers will this one supercede/overlap with?**

None

**Are you having any problems? Do you need our help with anything?**

Yes, I would like to have the table's URLs to be hyperlinks.

**Where did you hear about DuckDuckHack? (For first time contributors)**

https://duckduckgo.com/tour

**What does the Instant Answer look like? (Provide a screenshot for new or updated Instant Answers)**

securityaddons

## Checklist
Please place an 'X' where appropriate.

```
- [X] Added metadata and attribution information (https://duck.co/duckduckhack/metadata)
- [X] Wrote test file and added to t/ directory (https://duck.co/duckduckhack/goodie_tests)
- [X] Verified that instant answer adheres to design guidelines (https://duck.co/duckduckhack/design_styleguide)
- [X] Verified that instant answer adheres to code styleguide (https://duck.co/duckduckhack/code_styleguide)
- [X] Tested cross-browser compatibility

    Please let us know which browsers/devices you've tested on:
    - Windows 10
        - [X] Google Chrome
        - [X] Firefox
        - [] Opera
        - [] IE 10
        - [X] IE 11
        - [X] MS Edge

[...]

PR #1835, opened 2015-11-30
Issue #1163 - Convert Average.p to use template

Mailkov:
-----
IA Page: https://duck.co/ia/view/average


lib/DDG/Goodie/Average.pm
```diff
@@ -65,7 +65,25 @@ handle remainder => sub {
...
+    structured_answer => {
+        id => 'average',
+        name => 'Math',
+        data => {
+            mean => $mean,
+            median => $med,
+            rms => $rms
+        },
+        templates => {
+            group => 'text',
```
moollaza: Hmm I'm thinking it may be simpler for us to use the `list` (or `text` group) and use the `record` sub-template for the content.

If you construct a `record_data` it will spit out a table very similar to what we have here. We could then ditch the custom template and CSS.

I left a comment the other day explaining the record template with a code snippet: https://github.com/duckduckgo/zeroclickinfo-goodies/pull/1478#issuecomment-160039528

Mailkov: I used this because I thought that the answer should be the same graphically!

It was more work ... :)

moollaza:
@Mailkov I just spoke with the design team, and they would prefer that the IA only return the answer that was asked for, so Average, only provides the Average, Mean would only give the Mean.

This means we can ditch the custom CSS and templates and opt for our standard approach.

We can use the text template, or we can switch to the simplified `structured_answer` which has three properties: https://talsraviv.gitbooks.io/duckduckhackdocs/content/duckduckhack/frontend-reference/setting-goodie-display.html

Would you mind switching this AI to a simple structured Answer?

Mailkov:
@moollaza there is no problem.

lib/DDG/Goodie/Average.pm
```diff
@@ -17,9 +17,23 @@ code_url 'https://github.com/duckduckgo/zeroclickinfo-goodies/blob/master/lib/DD
...
 topics 'math';
 attribution twitter => ['crazedpsyc','crazedpsyc'],
-            cpan    => ['CRZEDPSYC','crazedpsyc'];
+            cpan    => ['CRZEDPSYC','crazedpsyc'],
+            github  => ['http://github.com/Mailkov', 'Melchiorre Alastra'];
 
-handle remainder => sub {
+
+
+handle query => sub {
```
moollaza: if you want, you could `handle remainder` but the do: 

```perl
my $query = $req->query_lc;
```

moollaza: We expose a `$reg` object, which contains lots of info, including the `query`, `query_lc`, `query_raw`, etc

daxtheduck:
Automated screenshot from [beta.duckduckgo.com](https://beta.duckduckgo.com/?q=average 12, 45, 78, 1234)
                
                [Mobile screenshot](https://ia-screenshots.s3.amazonaws.com/average_mobile.png)

lib/DDG/Goodie/Average.pm
```diff
@@ -17,10 +17,24 @@ code_url 'https://github.com/duckduckgo/zeroclickinfo-goodies/blob/master/lib/DD
...
+            github  => ['http://github.com/Mailkov', 'Melchiorre Alastra'];
+
+
 
 handle remainder => sub {
 
+    my $query = $req->query_lc;
+    
+    my $type;
+    if ($query =~ m/root mean square/) {
```
moollaza: nit: we may want to add `/rms/`

moollaza:

PR #1845, opened 2015-12-02
Remove Shortcut IA

zachthompson:
Related to our findings with the alt_to work, remove this IA which is no longer in service.


moollaza:
Thanks! This LGTM :+1:

PR #1853, opened 2015-12-06
Fix #1837 - ensure it has a remainder to sort

### LINKED ISSUE #1837: Sort: Needs to ensure it has a remainder to sort
Looks like this IA triggers on `array.prototype.sort` (the triggering part is our fault, and I intend to fix that), however the IA goes on to return an answer with no actual sorted output.

The IA should prevent that from happening.

E.g. https://duckduckgo.com/?q=Array.prototype.sort&t=ffcm&ia=answer

cc/ @koosha--

------
IA Page:  http://duck.co/ia/view/sort

Mailkov:
I think that this IA should sort the words too ...


mintsoft:
Thanks again @Mailkov LGTM!

PR #1858, opened 2015-12-07
Issue #1163 - Convert Sudoku to use templates

Mailkov:
-----
IA Page: https://duck.co/ia/view/sudoku

### DIFF
--- a/lib/DDG/Goodie/Sudoku.pm
+++ b/lib/DDG/Goodie/Sudoku.pm
@@ -19,24 +19,8 @@ name 'Sudoku';
 category 'random';
 attribution github => ['DrDub', 'Pablo Duboue'],
 	web => ['http://duboue.net/', 'Pablo Duboue'],
-	github => ['https://github.com/mintsoft', 'Rob Emery'];
-
-sub parse_to_html_table(@)
-{
-	my @sudoku_lines = @_;
-	my $html_table = "<table class='sudoku'>\n";
-	for my $line (@sudoku_lines)
-	{
-		my @chars = split(/ /, $line);
-		for my $char (@chars)
-		{
-			$char = "<input maxlength='1'/>" if $char eq "_";
-		}
-		$html_table .= "<tr><td>" . join("</td><td>", @chars) . "</td></tr>\n";
-	}
-	$html_table .= "</table>";
-	return $html_table;
-}
+	github => ['https://github.com/mintsoft', 'Rob Emery'],
+    github => ['https://github.com/Mailkov', 'Melchiorre Alastra'];
 
 handle remainder => sub {
 
@@ -55,13 +39,40 @@ handle remainder => sub {
 	$sudoku->generate(blanks => (9 ** 2) * $blanks);
 	my $str_output = $sudoku->as_string();
 
-	#switch 0 to more sensible placeholders
-	$str_output =~ s/0/_/g;
-
 	my @sudoku_lines = split(/\n/, $str_output);
-	my $html_table = parse_to_html_table(@sudoku_lines);
 
-	return $str_output, html => $html_table;
+	my @sudoku_tables;
+    
+    my @number;
+    for my $line (@sudoku_lines) {
+        $line =~ s/0//g;
+        @number = split(/ /, $line);
+        if (scalar(@number) == 8) {
+            push @number, "";
+        }
+		push @sudoku_tables, [@number];
+    }    
+    
+    #switch 0 to more sensible placeholders
+	$str_output =~ s/0/_/g;
+    
+	return $str_output,
+    structured_answer => {
+        id => 'sudoku',
+        name => 'Answer',
+        data => {
+            title => 'Sudoku',
+            subtitle => 'Level: ' . $difficulty,
+            sudoku_tables => \@sudoku_tables
+        },
+        templates => {
+            group => 'text',
+            item => 0,
+            options => {
+                content => 'DDH.sudoku.content'
+            }
+        }
+    };
 };
 
 1;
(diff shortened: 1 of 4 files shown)


lib/DDG/Goodie/Sudoku.pm
```diff
@@ -55,13 +39,40 @@ handle remainder => sub {
...
+    #switch 0 to more sensible placeholders
+	$str_output =~ s/0/_/g;
+    
+	return $str_output,
+    structured_answer => {
+        id => 'sudoku',
+        name => 'Answer',
+        data => {
+            title => 'Sudoku',
+            subtitle => 'Level: ' . $difficulty,
```
moollaza: I think we should switch "Level:" for "Difficulty:"

lib/DDG/Goodie/Sudoku.pm
```diff
@@ -55,13 +39,40 @@ handle remainder => sub {
...
+		push @sudoku_tables, [@number];
+    }    
+    
+    #switch 0 to more sensible placeholders
+	$str_output =~ s/0/_/g;
+    
+	return $str_output,
+    structured_answer => {
+        id => 'sudoku',
+        name => 'Answer',
```
moollaza: We can probably go with "Games" here

share/goodie/sudoku/sudoku.css
```diff
@@ -1,4 +1,4 @@
-.zci--answer table.sudoku td {
+.c-base__content table.sudoku td {
```
moollaza: I would say we should target it with `.zci--sudoku table.sudoku td` -- we always start the CSS rules with the class for the `zci--*` wrapper div as there's only 1 on the page.

t/Sudoku.t
```diff
@@ -15,29 +15,95 @@ ddg_goodie_test(
 	],
 	"sudoku" => test_zci(
 		qr/^[0-9_].*[0-9_]$/s,
-		html => qr/.*\<table.*\<\/table\>.*/s,
+		structured_answer => {
```
moollaza: Looks like there's a mix of tabs + spaces. Let's switch it all to spaces :+1:

moollaza:
This is now live on beta: https://beta.duckduckgo.com/?q=sudoku+easy&ia=games

@abeyang this one needs a little design help -- I think we should make everything a little bigger?

**Before:**

sudoku_easy_at_duckduckgo

---

**Here's what I came up with:**

sudoku_at_duckduckgo

- font-size `tx--21`
- input have transparent background, bottom border
- width of square is `2em` (from `1.5em`)

Thoughts?

moollaza:
@Mailkov after some closer inspection, I'm worried that this library might not work. As well the Perl approach could cause problems, because the module states it `could take minutes` to generate a game.

We need to ensure it takes less than 1s, so the ideal approach may be a pure-JS game, where we can load the game once the page starts loading, instead of holding up the page rendering.

Some issues:

- after a few refreshes, the game no longer changes
- on harder difficulties, we're showing many missing tiles

---

Another benefit to the JS approach is that we can react to the game when it's solved, and other things, like reveal the solution.

If you'd like to continue working on this you're welcome to. If not, we can make an issue for someone else to investigate a JS approach -- I don't want to burden you with unexpected work :)

/cc @mintsoft I think you made this originally? Any thoughts/ideas?

mintsoft:
@moollaza IIRC it was an abandoned PR; I think it probably makes sense to switch it out to JS if there already exists a generating library in JS (http://einaregilsson.com/sudoku-in-javascript/ looks plausible). Although it'd basically a new IA (I don't think this one has ever actually been live?) it might be worth just deleting this completely and starting afresh?

Mailkov:
@moollaza @mintsoft 
about issues: 
> * after a few refreshes, the game no longer changes
> * on harder difficulties, we're showing many missing tiles

I think you aren't testing the last version ...

About JS approach , I completely agree ...
Some years ago, I have studied how to generate Sudoku for a Java program and its solution, so I can work on this, I could converting my old code from Java to Javascript or I could using the library suggested by @mintsoft  or other .... but not immediately, so if you want fix this issue quickly you are free to make an issue for someone else.

moollaza:
@Mailkov @mintsoft I think the Pure JS version is ideal. 

>it might be worth just deleting this completely and starting afresh?

Sure. The design I proposed above still works, we just need to have the Goodie use JS.

@Mailkov if you'd like, you could adapt the lib @mintsoft mentioned: https://raw.githubusercontent.com/einaregilsson/sudoku.js/master/sudoku.js

It's about 700 LOC, but part of that handles the UI creation I think?

If we can find a small, actively maintained library I can make it available through `DDH.require()` and host in on DDG Servers outside of the repo.

Mailkov:
@moollaza  @mintsoft
I'm not sure that the lib @mintsoft mentioned is the best way ...
Reading http://einaregilsson.com/sudoku-in-javascript/
I noticed this : 
>I also wanted to make sure that every generated puzzle could be solved using only logic. Humans are (surprisingly enough) not as good at backtracking search as computers, so whenever I have removed a square in the algorithm above I also check if the puzzle can be solved using only elimination techniques. If it can't I reject it. That also means that the puzzles generated are kind of simple, there are a number of techiques humans use in Sudoku that are not just simple elimination, but the puzzles I generate don't really require them. So, there is definitely room for improvement!

So the difficult level will be always simple. 
Usually, the difficult isn't determinated by number of empty cells!

and this:
>875 milliseconds per generated puzzle isn't exactly great. I was planning on generating a new puzzle at the start of every game on http://www.sudoku-webgame.com, but adding almost a second to the load time of the page simply wasn't good enough. I could always see the delay and it annoyed me. So, what I decided to do instead was to just pre-generate a good batch of puzzles and ship them with the game, and then randomly choose one when the game starts. It kind of annoys me, because I'd prefer to generate on-the-fly, but I want things to feel fast.

So this library have only a limitated number of Sudoku puzzles!

If you give me some weeks ... I can make a small, actively maintained library!

moollaza:
@Mailkov thanks for pointing those out -- I agree that it looks like that lib won't be ideal

>If you give me some weeks ... I can make a small, actively maintained library!

That'd be amazing! There's no rush :)

PR #1859, opened 2015-12-07
Add rms as root mean square on Average

Mailkov:
@moollaza #1835


moollaza:

PR #1865, opened 2015-12-08
only return Unicode match result when there is a match

### LINKED ISSUE #1430: Unicode: Throws an error on the query "unicode girls"
While testing in DuckPAN I noticed that this IA returns an `undef` result for the query "unicode girls". This should be fixed asap!

I suspect this should be fairly easy to diagnose and resolve :)

------
IA Page:  http://duck.co/ia/view/unicode

marianosimone:
Sorry about the late response...

As discussed in the issue, the problem was that the code was returning whatever `unicode_lookup` returns, without validating if it was a defined result

fixes #1430


mintsoft:
LGTM; thanks @marianosimone

Issue #1694, opened 2015-10-26
Calculator: Negative explonents are calculated incorrectly


pafcu:
[1\*1e1](https://duckduckgo.com/?q=1*%281e-1%29&t=ffcm&ia=answer) gives 10, but [1\*1e-1](https://duckduckgo.com/?q=1*%281e-1%29&t=ffcm&ia=answer) gives 1.71828182845905

It seems like in the latter case the "e" is interpreted as the mathematical constant, not exponential notation.

------
IA Page:  http://duck.co/ia/view/calculator

moollaza:
@pafcu thanks for pointing this out! We'll take a look..

GuiltyDolphin:
This should be fixed now (from PR #1866).

mintsoft:
Thanks @GuiltyDolphin

marnix:
Just a note: this bug was closed Dec 9th, but at this moment still shows up on https://duck.co/ia/dev/issues.  No idea where to report that meta-bug... :-)

GuiltyDolphin:
@marnix, @moollaza, @mintsoft Yeah, I've noticed this too for many of the IA's how often are the pages updated?

moollaza:
I think there were some bugs that the team is fixing to make sure these update much quicker. They should be fixed soon

PR #1866, opened 2015-12-08
Calculator: Fix issue #1694

GuiltyDolphin:
**Which issues (if any) does this change solve? (please link them here)**

Fixes issue #1694 

**Briefly describe your fix. Are there any important details we should know about?**

I've just corrected some regular expressions to make the expressions work (and added
tests to make sure they continue working in the future).

Not really.

### DIFF
--- a/t/Calculator.t
+++ b/t/Calculator.t
@@ -705,6 +705,24 @@ ddg_goodie_test(
             result    => qr/>400,001</
         }
     ),
+    '3e-2* 9 ' => test_zci(
+        '(3  *  10 ^- 2) * 9 = 0.27',
+        heading           => 'Calculator',
+        structured_answer => {
+            input     => ['(3  *  10 ^- 2) * 9'],
+            operation => 'Calculate',
+            result    => qr/>0.27</
+        }
+    ),
+    '7e-4 *8' => test_zci(
+        '(7  *  10 ^- 4) * 8 = 0.0056',
+        heading           => 'Calculator',
+        structured_answer => {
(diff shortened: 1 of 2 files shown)


t/Calculator.t
```diff
@@ -705,6 +705,24 @@ ddg_goodie_test(
...
+    '3e-2* 9 ' => test_zci(
+        '(3  *  10 ^- 2) * 9 = 0.27',
+        heading           => 'Calculator',
+        structured_answer => {
+            input     => ['(3  *  10 ^- 2) * 9'],
+            operation => 'Calculate',
+            result    => qr/>0.27</
+        }
+    ),
+    '7e-4 *8' => test_zci(
```
mintsoft: Can you add some tests with the format reversed? `8 * 6e-3` for example

GuiltyDolphin: Have done, added a couple more tests.

GuiltyDolphin: @mintsoft I've noticed that with the fix, queries like `1 + 1e - 1` will still be treated as `1 + (1 * 10 ^ -1)`, rather than `1 + 1 * e - 1`, is this acceptible as the `1` and `e` are joined? (queries like `1 + 1 * e - 1` still return `e`, as expected.

mintsoft: @GuiltyDolphin That's what I would expect personally; I think it's fine as is. If it causes confusion/an issue then we can readdress.

mintsoft:
Many Thanks @GuiltyDolphin

PR #1875, opened 2015-12-09
Fix #1873 - Shouldn't trigger when query is only 'Rubics Cube'

### LINKED ISSUE #1873: Rubiks Cube Patterns: Shouldn't trigger on when query is only "Rubics Cube"
https://duckduckgo.com/?q=rubics+cube&t=ffab&ia=answer -- this shouldn't not display the IA.

Let's fix and add a test to ensure this :+1: 

/cc @Mailkov 

------
IA Page:  http://duck.co/ia/view/rubiks_cube_patterns

Mailkov:
-----
IA Page: https://duck.co/ia/view/rubiks_cube_patterns

### DIFF
--- a/lib/DDG/Goodie/RubiksCubePatterns.pm
+++ b/lib/DDG/Goodie/RubiksCubePatterns.pm
@@ -21,7 +21,7 @@ attribution web => ['robert.io', 'Robert Picard'],
 triggers start =>	"rcube", "rubik", "rubiks", "rubix",
 					"rubicks", "rubik's", "rubic's", "rubick's",
 					"rubik cube", "rubiks cube", "rubic cube", "rubics cube",
-					"rubik's cube patterns", "rubiks cube patterns";
+					"rubik's cube";
 
 zci answer_type => "rubiks_cube";
 zci is_cached   => 1;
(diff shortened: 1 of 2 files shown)


lib/DDG/Goodie/RubiksCubePatterns.pm
```diff
@@ -56,6 +56,7 @@ sub render_text($) {
 
 handle remainder_lc => sub {
 
+    return if ($req->query_lc eq 'rubics cube');
```
moollaza: @Mailkov my apologies, I should have been a little more thorough -- all other forms of the trigger words are also able to trigger this IA and provide unexpected results.

We need to ensure we have a useful remainder to work with, when needed. Queries like "rubic's cube", "rubics cube", "rubik's cube", "rcube", etc shouldn't trigger this.

Only when a pattern is provided, or for queries like "rubicks cube patterns", "rcube patters", which contain `"patterns"` should it return a result

daxtheduck:
Automated screenshot from [beta.duckduckgo.com](https://beta.duckduckgo.com/?q=rcube stripes)
                
                [Mobile screenshot](https://ia-screenshots.s3.amazonaws.com/rubiks_cube_patterns_mobile.png)

moollaza:
@Mailkov this is looking much better!

BTW: I just found this site, http://ruwix.com/the-rubiks-cube/rubiks-cube-patterns-algorithms/

I think for "rubics cube patterns" a tile view with those pictures would be neat. I'll make an issue for the community to tackle. Please don't feel obliged unless you'd like to tackle it :+1:

PR #1303, opened 2015-07-18
ChordDiagrams (continues #1023)

charles-l:
PR to continue #1023 which was abandoned.

What I've done:
- Moved chord code to a handlebars template
- Cleaned up old HTML generation code

-----
IA Page: https://duck.co/ia/view/chord_diagrams

PR #1879, opened 2015-12-10
Issue 1163 - Convert GuitarChords to use template

Mailkov:


mintsoft:
@Mailkov before you go much further on this, there is a PR in progress to change the way this IA works entirely. https://github.com/duckduckgo/zeroclickinfo-goodies/pull/1303 I don't know about ETA's on it so it might be worth having this conversion inplace before the other PR is finished. @moollaza thoughts?

PR #1888, opened 2015-12-12
remove html_enc() from Goodies, template will escape content

### LINKED ISSUE #1887: Atbash: Showing html encoding in answer
For example, with the query `"hello"`;

https://duckduckgo.com/?t=lm&q=atbash+%22hello%22&ia=answer

The formatted query is returning as `Atbash: &quot;hello&quot;`, not pretty!

moollaza:
Fixes #1887

### DIFF
--- a/lib/DDG/Goodie/ABC.pm
+++ b/lib/DDG/Goodie/ABC.pm
@@ -53,7 +53,7 @@ handle remainder => sub {
 
     return $selection . " (" . $selection_type . ")",
       structured_answer => {
-        input     => [html_enc($choice_string)],
+        input     => [$choice_string],
         operation => $operation,
         result    => html_enc($selection),
       };
--- a/lib/DDG/Goodie/Anagram.pm
+++ b/lib/DDG/Goodie/Anagram.pm
@@ -108,7 +108,7 @@ handle remainder => sub {
 
     return $response,
       structured_answer => {
-        input     => [html_enc($word)],
+        input     => [$word],
         operation => $operation,
         result    => html_enc($response)
       };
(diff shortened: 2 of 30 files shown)


mintsoft:
@moollaza has the way that the templates handle this changed recently? Last time I converted one of them I believe the `html_enc` was required?

moollaza:
@mintsoft nothing has changed, I don't recall why we needed to HTML encode this in the first place because as mentioned the template is set to safely handle the `input`.

I did notice the page break though on the query `"lowercase <script></script>"` I believe that's the result of a DuckPAN error though. I will confirm and update this.

moollaza:
Closing.

@mintsoft @GuiltyDolphin @Mailkov I've spoken with @zachthompson and for now we need to continue to `html_enc()` the subtitles (or any other user input passed through the templates).

We do have a double encoding issue which the team is going to fix.

---

**Explanation:**

- Structured answer is converted to JSON and passed along to SERP.
- The templating JS grabs this JSON and passes it along as the input to the Goodie's template
- The template renders the HTML you see on the page

If we don't encode unsafe content, we risk breaking the JSON on the SERP.
(e.g. https://beta.duckduckgo.com/?q=ceasar+3+%3C%2Fscript%3E&ia=iatesting -- notice the flash of JSON?)

Once the content is encoded it get's directly passed to the template. All the templates are setup to escape their input (i.e. they use `{{ }}` instead of `{{{ }}}`).

This creates a "double encoding" as the encoded JSON we pass along is then escaped by the templates.

Resulting in: https://duckduckgo.com/?q=lowercase+%3C%2FSCRIPT%3E&t=ffab&ia=answer

PR #1892, opened 2015-12-13
Lowercase: Nil return when output is equal to input

### LINKED ISSUE #1871: Lowercase: Should ignore input that's only numerical
https://duckduckgo.com/?q=lowercase+1235&t=ffab&ia=answer

We should ignore input that only has numbers. We should probably not show a result if the output === input actually?

------
IA Page:  http://duck.co/ia/view/lowercase

altern8tif:
Fixes #1871.

### DIFF
--- a/lib/DDG/Goodie/Lowercase.pm
+++ b/lib/DDG/Goodie/Lowercase.pm
@@ -25,12 +25,16 @@ handle remainder => sub {
 
     my $lower = lc $input;
 
-    return $lower,
-      structured_answer => {
-        input     => [html_enc($input)],
-        operation => 'Lowercase',
-        result    => html_enc($lower)
-      };
+    if ($lower eq $input) {
+    	return;
+    } else {
+	    return $lower,
+	      structured_answer => {
+	        input     => [html_enc($input)],
+	        operation => 'Lowercase',
+	        result    => html_enc($lower)
+	      };
+    }
 };
 
 1;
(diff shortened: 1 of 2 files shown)


t/Lowercase.t
```diff
@@ -11,14 +11,8 @@ zci is_cached   => 1;
...
-    'lowercase foo' => test_zci(
-        'foo',
-        structured_answer => {
-            input     => ['foo'],
-            operation => 'Lowercase',
-            result    => 'foo'
-        },
-    ),
+    'lowercase foo' => undef,
+    'lowercase 123' => undef,
```
GuiltyDolphin: @altern8tif Maybe add a case with mixed lowercase and numeric input? And a case with mixed uppercase and numeric input.

altern8tif: :+1:

lib/DDG/Goodie/Lowercase.pm
```diff
@@ -25,12 +25,16 @@ handle remainder => sub {
 
     my $lower = lc $input;
 
-    return $lower,
-      structured_answer => {
-        input     => [html_enc($input)],
-        operation => 'Lowercase',
-        result    => html_enc($lower)
-      };
+    if ($lower eq $input) {
```
mintsoft: I'd probably write this as 
```perl
return if ($lower eq $input);
return $lower,
    structured_answer => {
        input     => [html_enc($input)],
        operation => 'Lowercase',
        result    => html_enc($lower)
    };
```

moollaza:
@altern8tif Thanks! This LGTM :+1:

PR #1894, opened 2015-12-13
Issue #1163 - Convert PrivateNetwork to use templates

Mailkov:
-----
IA Page: https://duck.co/ia/view/private_network


moollaza:
@Mailkov we're having some problems with Beta right now. I'll get it deployed shortly.

I need to speak with the design team about the best way to handle this. I think we either need a proper table with columns/headers, or we need to move to a tile view potentially. This doesn't look great, and the live version is somehow worse. I think our CSS classes changed and it looks a little odd now...

Mailkov:
@moollaza 
And 'since I worked on PublicDNS I think it would be appropriate to create a template table ...If you agree I have some ideas, when you are free we could talk about it

moollaza:
>And 'since I worked on PublicDNS I think it would be appropriate to create a template table ...If you agree I have some ideas, when you are free we could talk about it

Sure, happy to chat. I imagine we would want a special template, similar to https://github.com/duckduckgo/duckduckgo-answerbar-templates/blob/master/shared/record.handlebars

Except something that can handle multiple rows/columns

moollaza:
@Mailkov are you still interested in making a new `table` helper for this IA (and the others)? I think it would be really useful!

Mailkov:
@moollaza 
Yes, I'm interested ...

jdorweiler:
@Mailkov we can re-open if you want to pick this back up.

PR #1895, opened 2015-12-13
Constants: Update IA description

altern8tif:
Remove default description in metadata.  Updated with description provided in IA page.

------
IA Page:  http://duck.co/ia/view/constants


moollaza:
@altern8tif Thanks! This LGTM :+1: 

(fwiw, very shortly the Perl metadata will be removed and the IA page will be the only place for this stuff :))

PR #1904, opened 2015-12-15
Fix #1836 Regex trigger needs anchors to limit triggering

### LINKED ISSUE #1836: LaserShip: Regex trigger needs anchors to limit triggering
Hi,

A DDG search for a LaserShip Tracking number in the form "1LS12345678999999999", results in a truncated tracking ID link link like so http://lasership.com/track/LS12345678 (no leading "1" truncated to 8 trailing digits).  This shortened URL results in a LaserShip validation error, "So sorry, the tracking number could not be located ...", whereas the full form, http://lasership.com/track/1LS12345678999999999 displays expected results.

------
IA Page:  http://duck.co/ia/view/laser_ship

jonk1993:
Fixes #1836. Updated regex to include anchors so it will not match on strings such as "1LS12345678999999" but will still match on "LS12345678"

### DIFF
--- a/lib/DDG/Goodie/LaserShip.pm
+++ b/lib/DDG/Goodie/LaserShip.pm
@@ -15,7 +15,7 @@ category 'ids';
 topics 'special_interest';
 attribution github => [ 'https://github.com/duckduckgo', 'duckduckgo'];
 
-triggers query_nowhitespace_nodash => qr/(l[a-z]\d{8})/i;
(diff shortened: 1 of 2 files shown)


mintsoft:
@jonk1993 Can you add a test for the case that this solves?
```perl
"1LS12345678999999999" => undef
```

jonk1993:
Yep, included.

mintsoft:
Excellent, many thanks @jonk1993 !

Issue #1318, opened 2015-07-21
Calculator: FR - bar notation


blackhat999:
Show bar notation for repeating decimals

---

IA Page:  http://duck.co/ia/view/calculator

moollaza:
We've change the UI of the calculator. This is not something we plan to implement for now.

PR #1927, opened 2015-12-19
Issue #1880 - Re-write calculator

### LINKED ISSUE #906: Calculator: Show decimal notation as much as possible before showing exponent notation?
Maybe this should still be displayed as 0.00005 instead of 5x10^-5

screen shot 2015-01-07 at 1 04 43 pm

------
IA Page: https://duck.co/ia/view/calculator

### LINKED ISSUE #1187: Calculator: Add support for advanced calculations
Add the ability to solve e.g. `10!`, `20nCr8`, etc. as well as equations.

Suggestion source: https://duck.co/forum/thread/13748/calculator-improvements

---

IA Page: https://duck.co/ia/view/calculator

### LINKED ISSUE #1389: Calculator: Switch Math::BigInt to use GMP when available for increased performance
If we switch our usages of `Math::BigInt` and `bigint` to `Math::BigInt try => "GMP"` It will allow us to use GMP when available (meaning production) but fallback to a slower pure Perl implementation otherwise (for devs that don't have GMP installed).

This should be updated wherever we're using BigInt.

------
IA Page: https://duck.co/ia/view/calculator

### LINKED ISSUE #1418: Factorial: Move integration into Calculator.pm; Calculate factorial with Math::BigInt bfac()
We recently merged a new Factorial Goodie. The current implementation works, but really this functionality should be a part of the existing Calculator Goodie.

As well. the current implementation uses a naive method of calculating the Factorial using recursion via `List::Util::reduce`. This should definitely be switched to use `Math::BigInt`'s `bfac()` method.

This will enable us to calculate considerably larger values, however we need to wrap this logic in an `alarm` (or similar) so that we don't spend more than 1s calculating extremely large numbers.

Alternatively we could enforce a reasonable upper limit on the highest input we're willing to accept. Maybe 10,000? or something smaller, like 1000?

Any Perl experts out there interested in tackling this one?

---

IA Page: https://duck.co/ia/view/calculator

### LINKED ISSUE #1765: Calculator: Basic functions not recognized
See the example:

`exp(0) - log(e) + sqrt(9) / tan(45°)` 

The result should be 1 - 1 + 3 / 1 = 3

Instead [no answer is returned](https://duckduckgo.com/?q=exp%280%29+-+log%28e%29+%2B+sqrt%289%29+%2F+tan%2845°%29)

The goody lacks of support to the following notations:
- [`exp(x)`](https://duckduckgo.com/?q=exp%283%29) ([`e^x`](https://duckduckgo.com/?q=e^3) is already supported) 
- [`TrigonometricalFunction(x°)`](https://duckduckgo.com/?q=sin%2845°%29) ([support to radians](%28https://duckduckgo.com/?q=sin%28pi%29%29) is already in the code)
- [`atan(x)`](https://duckduckgo.com/?q=atan%281%29) / [`arctan(x)`](https://duckduckgo.com/?q=arctan%281%29) (same with asin/arcsin and acos/arccos)
- [`HyperbolicTrigonometricalFunction(x)`](https://duckduckgo.com/?q=sinh%280%29) (`sinh`/`sh`, `cosh`/`ch`, `tanh`/`th`, `arcsinh`/`asinh`, `arccosh`/`acosh` and `arctanh`/`atanh`)

---

IA Page: http://duck.co/ia/view/calculator

### LINKED ISSUE #1880: Calculator: re-write
_(I blabber, scroll down a bit if you want to see some issues and suggestions)_

_Disclamer: I'm not a long-time DuckDuckHack contributer, I have a lot to learn and I'm not as well versed in Perl as many of you will be (I mostly write Haskell), so please, feel free to scream (or politely shout) at me if I've made some horrible error in the way things are supposed to work._

**_Begin**_

I recently wrote a small fix for the Calculator Goodie, but had a bit of pain finding where exactly the issue was due to the verbosity of the goodie. @moollaza informed me that the Calculator is one of the oldest IAs, so through updates it is getting a bit messy.

Whils't doing my fix, I noticed some odities in the way the goodie was written, and since then I have looked further in to it and come up with many suggestions for improvements.

**I propose that the Calculator IA be re-written**

`Calculator.pm` is approximately 190 sloc, with roughly 120 tests - not huge. I believe that a re-write would allow this code-base to be reduced significantly, whilst making the goodie more robust and potentially add more features.

I am aware that there are other powerful calculator implementations, such as [this](https://duck.co/ia/view/symbolab) spice, and [this](https://github.com/duckduckgo/zeroclickinfo-spice/pull/1727) promising GUI implementation; but Perl is a very powerful language, and a **lot** of people have invested in it - resulting in [some powerful math libraries](https://metacpan.org/search?q=Math).

The current Calculator implementation is very focused on using regular expressions to attempt to normalize user queries (`s/ (\d+?)E(-?\d+)([^\d]|\b) /\($1 * 10**$2\)$3/ixg` `2e3` anyone?) many of which are unnecessary.

I propose that a 'new' Calculator Goodie be made, with the intent of giving the user as much mathematical power as possible, whilst keeping the codebase clean and well documented, and delegating hard-work to libraries such as `Math::BigFloat` and `Math::BigInt`.
## Some current issues

_I would include screenshots... but think of the scrolling..._
1. _The code._
   It ain't pretty... And I'm pretty sure a lot of it doesn't need to be there. A good re-write would naturally fix this through careful planning and constant refactoring.
2. _Queries._ Some queries **just don't work**. Take for example, `e` (`* 10 to a power`). The query `1 + 3e8` works fine, but `3e8` on its own? Nope. This could be taken as erring on the side of safety, who knows what `3e8` could really be referring to (things like `2e3` and `2e4` already come up with IA results)? But prefixing the query with `calculate` doesn't fix this, even though the user is being _very_ explicit. This should change.
3. _Premature delegation._ The [spice](https://duck.co/ia/view/symbolab) mentioned earlier has its uses, but they lie with more complex mathematics such as calculus. The current Goodie is relying on it too much. As an example, the query `calculate 1 + 2e`. This isn't a complicated computation, and should be handled directly by the Goodie, rather than having to send out an API call (as it currently does). Besides, the SymboLab call is (by default, don't know if you can change this) only giving back 5 decimal digits! Perl supports 40 (ish) **_cached**_ digits of `e` out-of-the-box with `Math::BigFloat->bexp(1)`.

I have many other issues and things that could be improved noted down, but I'll stop here for now, hopefully I've convinced you.

If you agree - I'll be happy to put time into this, I think this is a very fundamental Goodie (too much maths?).
## TL;DR

_(you lazy... - that took me ages to write!)_

**I think the calculator needs re-writing.**

------ 
IA Page:  https://duck.co/ia/view/calculator

### LINKED ISSUE #1917: Calculator: Operation for `fact()` should be "factorial"
When computing only the factorial of a number, regardless of the query, we should show the operation as "factorial" instead of "fact"

i.e. it should look like this:

fact_3__at_duckduckgo

---

IA Page:  http://duck.co/ia/view/calculator

### LINKED ISSUE #2850: Calculator modulo operation
https://duckduckgo.com/?q=(10%5E100)+%25+10&ia=answer

(10^100) % 10 should be 0. 
I am not sure why the answer is wrong. Is the symbol % represent the modulo operation?

### LINKED ISSUE #2879: Calculator: exponential not always triggered
- `1 * 1e1` triggers correctly into calculator and evaluates to 10.
- `1 * 1.0e1` does not trigger calculator.  For some reason, tabs for Steve Miller's Joker are the first search hit :-)

Test cases should probably include
- all combinations of sign and no sign on exponent and mantissa 
- all combinations of float and int exponent and mantissa 
## Thanks.

IA Page:  http://duck.co/ia/view/calculator 
[Maintainer](http://docs.duckduckhack.com/maintaining/guidelines.html): @GuiltyDolphin

### LINKED ISSUE #3150: Calculator: Unable to evaluate some expressions
Some expressions are not being triggered

`cos(pi) + sin(pi)` works
`cos(pi/2) + sin(pi)` doesn't work
`cos(pi)/2 + sin(pi)` doesn't work

The same thing happens with other functions

---

IA Page:  http://duck.co/ia/view/calculator 
[Maintainer](http://docs.duckduckhack.com/maintaining/guidelines.html): @GuiltyDolphin

### LINKED ISSUE #3450: Calculator:  -A + B not works
Expr. [`-2 + 2 =`](https://duckduckgo.com/?q=-2+%2B+2+=) isn't works because starts with `-`.

---

IA Page:  http://duck.co/ia/view/calculator 
[Maintainer](http://docs.duckduckhack.com/maintaining/guidelines.html): @GuiltyDolphin

### LINKED ISSUE #3935: Calculator: Not triggering when using pi symbol
**π * 180**
[https://duckduckgo.com/?q=%CF%80%20%2A%20180](https://duckduckgo.com/?q=%CF%80%20%2A%20180)
_When searching for a calculating term like "π * 180" you won't get a result of calculation. The π symbol is easy to type in macOS, by just pressing and holdin "alt + P", which I often use._   

------
IA Page:  http://duck.co/ia/view/calculator 
[Maintainer](http://docs.duckduckhack.com/maintaining/guidelines.html): @GuiltyDolphin

GuiltyDolphin:
This regards issue #1880.

Re-write of the [Calculator IA](https://duck.co/ia/view/calculator).
## Main design changes

The old calculator relied on Regular expressions in order to ensure correct triggering as well as parse the user input. The calculator would then use `eval` (on a timer) to generate the result. Additionally, the old Goodie made heavy use of substitutions to format the user's input and answer.

The new Goodie uses much fewer regular expressions  - instead I've used the parsing library `Marpa::R2`, along with a grammar, to generate a parse tree from the user's input.

Because the parse tree is guaranteed to be well-formed, the triggers can be much looser, as the Goodie will not accept inputs that do not entirely conform to the grammar.

I then use a set of functions for generating the formatted input and the result from the parse tree.
### Answer format

_The Old_

screenshot from 2015-12-20 14 26 30

_The New_

screenshot from 2015-12-20 14 26 46

As you can see I've opted for a more fraction-based approach to the answer format. Floating-point errors are annoying and will likely just annoy some people.

_"But wait!"_ I hear you cry. _"I just love those floating-point errors!"_
No worries, just re-query with the simplified fraction!

screenshot from 2015-12-20 14 42 48

**See the To-do section below, under 'Functionality' for an improvement upon this.**
### Adding new functionality

Adding new features to the old calculator was a pain! Even trying to fix a teensy little parsing bug was a nightmare because of the mass of regular expressions.

In the new calculator, adding simple functionality like new functions is a breeze.

Firstly, add the relevant line to the grammar in `share/goodie/calculator/grammar.txt`;
screenshot from 2015-12-20 09 11 23
(this allows the user to use `exp(x)` to action the `exponential` function)

Then add the definition to `lib/DDG/Goodie/Calculator.pm`;
screenshot from 2015-12-20 09 11 56

Don't forget to add tests!

I'm thinking of automagically generating the function grammar from the definitions in `Calculator.pm`, so maybe only one line will be needed for simple functions soon!
### Source-related
- [ ] Provide auto-magical generation of the grammar so developers don't have to worry about it.
  - [x] Unary functions
  - [x] Binary operations
  - [x] Constants
  - [ ] Prefix Factor Modifiers
  - [x] Postfix Factor Modifiers
### Functionality
- [x] Add support for binary functions.
  
  I've spoken with @mintsoft about this one - there are some issues with the argument delimiter as commas will be a big issue due to the way numbers are handled (nationalization). @mintsoft suggested using semi-colons (so instead of `F(1, 2)` you would write `F(1; 2)`) - let me know what you think!
- [x] Add hyperbolic functions (`sinh`, `tanh` etc..)
- [ ] Display recurring decimal notation where possible.

I want your feedback on this one - I should be able to implement it fairly easily for short cycles (e.g., `1/3 = 0.333333...`) but I'll have to see about longer ones.

In terms of notation, using ASCII characters is going to be best. The [Wikipedia](https://en.wikipedia.org/wiki/Repeating_decimal#Notation) article mentions the parenthesis notation - wherein the recurring part is surrounded in parentheses (thus `1/3` becomes `0.(3)`). I've also seen notation using square brackets (`0.[3]`). Let me know what you think!
- [ ] Provide support for recurring decimals as input.

Similar to the previous point, but instead allow the user to enter queries like `0.(3)` and have the calculator recognize it as a third.
- [x] Display both decimal notation as well as fractions in output.
- [x] Support calculation modifiers.

Examples of this might be `arcsin(1/2) in degrees` or `1 + 1/7 as a fraction`. I need to think on this a bit more, but if anyone can come up with some good 'modifiers', please let me know!
- [ ] Support word descriptions of numbers (e.g., `three hundred` or `four thousand`).
- [ ] Support output in terms of irrational constants - e.g., `3π ≈ 9....`.
### Internal Functionality
- [ ] Implement a more robust precision-calculation system. Currently, actions such as `sin` and `cos` only support 15 figures, this is likely fine, but make sure that the rest of the calculations know this!
### Bugs
- [x] Exponents do not support more than `2 ^ 49` or less than `2 ^ -49`, roughly `10 ^ 16` and `10 ^ -16`.
- [x] Throws a fit with inputs such as `2e-11`.
- [x] Throws a tantrum with inputs such as `0.3333333333 + 0.7777777777`.
- [x] Incorrect precedence with `^` operator; `2 + (2 + 2) ^ 2 - 0 + 0` returns `36` when it should return `18`. This issue seems to occur when using `-`.
## Issues this affects
- [x] Resolves #1880 as the issue regards the re-write.
- [x] Resolves #906 (open to discussion about notation).
- [x] Resolves #1187 which also discusses a re-write.
- [ ] #1318 (recurring decimal notation).
- [x] ~~Resolves #1389 as there is no longer a reliance on `Math::BigInt`~~ We now use GMP.
- [x] Resolves #1418 as factorial is now dealt with by `Math::BigRat`/`Math::BigInt`. The calculator supports up to `1000!`.
- [x] Resolves #1765 more functions are being added, please comment with any suggestions for functions I've missed.
- [x] Resolves #1917 as the calculator now has this functionality.
- [x] Fixes #2850 as the answer returned is now correct.
- [x] Fixes #2879
- [X] Fixes #3150 as the expressions now work.
- [X] Fixes #3450 as these queries now work.
- [ ] #3871 (Pending `3^6067`)
- [X] Fixes #3935 as the π symbol is now supported.
## Features

<a id="features"></a>
### Functions

This is a comprehensive list of functions supported by the Calculator:
#### Unary Functions

Unary functions take the form `F(X)`.
##### Trigonometric Functions

All of the trigonometric functions support both radians and degrees.

`Trig(X)` will treat `X` as radians, `Trig(X degrees)` and `Trig(X°)` will treat `X` as being in degrees.
- `sin`, `sine` - sinus.
- `cos`, `cosine` - cosinus.
- `tan`, `tangent` - tangent.
- `sec`, `secant` - secant.
- `csc`, `cosec`, `cosecant` - cosecant.
- `cot`, `cotan`, `cotangent` - cotangent.
##### Inverse Trigonometric Functions
- `asin`, `arcsin` - inverse sine.
- `acos`, `arccos` - inverse cosine.
- `atan`, `arctan` - inverse tangent.
##### Hyberbolic Functions
- `sinh` - hyperbolic sine.
- `cosh` - hyperbolic cosine.
- `tanh` - hyperbolic tangent.
##### Inverse Hyperbolic Functions
- `asinh`, `arsinh` - inverse hyperbolic sine.
- `acosh`, `arcosh` - inverse hyperbolic cosine.
- `atanh`, `artanh` - inverse hyperbolic tangent.
##### Logarithms
- `ln`, `log` - natural logarithm.
- `logB`, `log_B` (where `B` is a term) - logarithm base `B`.
##### Miscellaneous Functions
- `floor` - floor.
- `ceil`, `ceiling` - ceiling.
- `sqrt` - square root.
- `fact`, `factorial` - factorial.
- `exp` - exponential.
#### Binary Functions

Binary functions take the form `F(X; Y)`.
- `mod` - the modulo operation.
#### Unary Operations
- `X!` - factorial.
#### Binary Operations
- `X ^ Y`, `X ** Y`, `X to the power of Y`, `X to the power Y` - `X` raised to the power `Y`.
- `X * Y`, `X x Y`, `X ∙ Y`, `X ⋅ Y`, `X × Y`, `X times Y`, `X multiplied by Y` - multiplication.
- `X / Y`, `X ÷ Y`, `X divided by Y` - division.
- `X + Y`, `X plus Y`, `X add Y` - addition.
- `X - Y`, `X minus Y`, `X take Y`, `X subtract Y` - subtraction.
### Modifiers

Modifiers act upon individual factors in a calculation. They either perform a simple operation (e.g., `squared`), or cause some attribute of the factor to change (e.g., `in degrees`). Modifiers are tightly binding, so if multiple terms need to be 'modified' then they should be wrapped in parentheses (i.e, `4 + 3 squared == 4 + (3 squared)`, but `(4 + 3) squared == 7 squared`).
#### Postfix Modifiers
- `X squared` - `X` to the power of `2`.
### Constants
- `pi`, `π` - π (internally represented to 40 decimal places).
- `e` - Euler's number (base of the natural logarithm).
- `dozen` - `12`.
- `score` - `20`.
#### Input Format

Relies on the `NumberStyler` GoodieRole for supporting additional numeric conventions. I will describe number formats using the British Decimal System.
- Integers are represented by a stream of digits without a decimal place, e.g, `42`.
- Deimals are represented by an integer part, followed by a decimal place, followed by a fractional part; where either the integer part or fractional part are optional, but not both. Examples: `1.3`, `1.`, and `.3` are all valid, but `.` is not.
- Numbers can be optionally prefixed with a negative sign (`-`), but only if there is an integer part.
- Numbers can optionally be represented in `e` scientific notation. I.e, `1e3` is equivalent to `1000`.
- Numbers can optionally be prefixed with a currency symbol. Currently only dollars (`$`) are supported. Example: `$4.75 + $3.25` equals `$8.00`.
- A number placed directly before a constant will cause the constant to be multiplied by a factor of that number. E.g., `3π` would equal `3 * π`, and `4 score + 7` would equal `87`.
#### Output Format

Relies on the `NumberStyler` GoodieRole for correct formatting of numbers in the result.
- If the result is an integer, then just that integer will be displayed. `1 + 7` becomes `8`.

Results will display fractions unless:
- The result is a factor of an irrational calculation (e.g, a multiple of `e` or `sqrt(2)`).
- The input is already a simplified fraction.
- Something has gone wrong.

Results will display decimals unless:
- The input is already a single decimal.
- Something has gone wrong.

Rules for displaying fractions:
- Fractions will always be shown in their simplified form.
- Fractions are displayed as `X/Y`.

Rules for displaying decimals:
- If the result is an approximation, then the rounded figure is prefixed with `≈`.
- If the result is exact, and no fraction is displayed, then just the decimal result is shown.
- If the result is exact, and a fraction is displayed, then the decimal is prefixed with `=`.
- Decimals are always placed after fractions.
- Decimals **MUST** be correct to 13 total figures, if this is not the case then the calculator has failed. **If any counter-examples are found, please let me know as a priority!**

Thus all the possible result formats are (where `I` is an integer, `D` is a decimal):
`I`, `D`, `≈ D`, `≈ I`, `I/I`, `I/I = D`, `I/I ≈ D`, `I/I ≈ I`.


daxtheduck:
Metadata post coming soon

jophab:
@GuiltyDolphin I know nothing about this  code. Just a suggestion.
 From the users point of view,querying again to get answer may be an overhead.So  can we avoid the necessity of querying the fractional answer to get floating point answer. 
Can we provide both answers as the result?
What you have done so far is great . :+1: 
I would like to learn too..

mintsoft:
@GuiltyDolphin I think I agree with @jophab in that we'll probably need to have both floating point output as well as formatted output.

t/Calculator.t
```diff
@@ -4,860 +4,199 @@ use strict;
...
+    '123.123.123.123/255.255.255.256' => build_test('0,4823529411746', '123.123.123.123 / 255.255.255.256'),
+    'pi/1e9'                          => build_test('3.14159265359 * 10^-9', 'pi / 1e9'),
+    'pi*1e9'                          => build_test('3,141,592,653.59', 'pi * 1e9'),
+    # Nationalization
+    '1 234 + 5 432'      => build_test('6,666', '1,234 + 5,432'),
+    '1_234 + 5_432'      => build_test('6,666', '1,234 + 5,432'),
+    '4.243,34+22.538,28' => build_test('1.339.081/50', '4.243,34 + 22.538,28'),
+    # Factorial
+    'fact(3)'                         => build_test('6', 'factorial(3)'),
+    'factorial(3)'                    => build_test('6', 'factorial(3)'),
```
mintsoft: Is `3!` notation supported?

GuiltyDolphin:
@mintsoft No, not currently - I'm a bit worried about clashing with bang notation.

@mintsoft, @jophab Should be easy enough - but I want the output to be 'correct'.

When the correct result cannot be fully displayed, I could display (from the query `1 + 7/9`) `16/9 (approximately 1.777777777778)` for example - which is correct.

For results that are already correct (e.g, `5/8 = 0.625`) I could display `5/8 (0.625)` or `5/8 = 0.625` or `5/8 or 0.625` etc...

If I implement the functionality for recurring decimals, then more of the results could be displayed in the second form.

What do you think?

t/Calculator.t
GuiltyDolphin: @mintsoft No, not currently - I'm a bit worried about clashing with bang notation.

mintsoft: @GuiltyDolphin I'm not aware of any other use of `!` in mathematics other than factorial; what clashes are you referring to?

GuiltyDolphin: The DDG bang patterns, like `!github` - if there is no risk of them triggering then there aren't any other reasons why I wouldn't include the `!` notation.

mintsoft: Yeah they're handled by the platform before any IA's are executed, so it's totally safe to use them

GuiltyDolphin: @mintsoft It supports `!` notation now.

mintsoft: Awesome! Great job

jophab:
@GuiltyDolphin 
If you have checked google's calculator, it doesnt even show the fractional result.
So our calculator has an extra feature :+1:

jophab:
@GuiltyDolphin Actually what I thought  About the result is


You are doing a great job on this.
So I dont want to put extra work  on you.
What do you think about it?

mintsoft:
@jophab I think @GuiltyDolphin would say that to be 'correct' the result would have to be: `16/7 ≈ 1.77777777777778`

jophab:
@mintsoft :+1: 
Yeah, approximatly equal to symbol is better

GuiltyDolphin:
@mintsoft, @jophab Sure, but it's not ASCII. Am I able to check if the user can support non-ASCII and display the symbol if possible, but resort to (approximately) when they don't have it?

GuiltyDolphin:
As a basic prototype - it might look something like this then:

screenshot from 2015-12-21 10 46 36

Where the tilde is replaced with whichever sign.

mintsoft:
@GuiltyDolphin We're pretty heavy on the UTF-8 in most places now, so it should be acceptable to use unicode

GuiltyDolphin:
Okay, here's what it looks like at the moment:

screenshot from 2015-12-21 11 06 17

screenshot from 2015-12-21 11 06 32

screenshot from 2015-12-21 11 06 46

jophab:
@GuiltyDolphin :+1:

GuiltyDolphin:
@mintsoft, @jophab Rightio - I've implemented the fractions and decimals displaying at the same time, feel free to pull and check it out!

GuiltyDolphin:
@mintsoft I might have to switch to using the `Math::Cephes::Fraction` package - bit inconvenient (non-overloaded) but `Number::Fraction` keeps having issues for me (would remove a dependency).

mintsoft:
@GuiltyDolphin I've used `Math::BigRat` before and found it pretty solid

lib/DDG/Goodie/Calculator.pm
```diff
@@ -1,190 +1,503 @@
...
+use Math::Round;
+use Number::Fraction;
+
+zci answer_type => "calculation";
 zci is_cached   => 1;
 
-primary_example_queries '$3.43+$34.45';
-secondary_example_queries '64*343';
-description 'Basic calculations';
+primary_example_queries '1 + 7', '5 squared', '8 / 4';
```
moollaza: We can actually remove the Perl metadata. We should port the examples to the IA Page: https://duck.co/ia/view/calculator

GuiltyDolphin: @moollaza From line 23 `primary example queries...` to line 33 `attribution...`?

GuiltyDolphin:
@mintsoft I've just re-written a bunch of the source to allow for a wider range of inputs. Also hopefully this will allow for more features in the future (things like `700 in hexadecimal`, `what is 1 / 75317 to 30 decimal places` etc...). I'm wrapping the result in a custom 'object' that allows tracking of certain attributes so that I can support `Math::BigRat`.

Will push soon - let me know if you've come across any issues with the existing code!

mintsoft:
@GuiltyDolphin it all sounds good, I'll have a more thorough look tomorrow hopefully as the Christmas shutdown kicks in at work. 

Unicode Pi should be fine btw!

GuiltyDolphin:
Regarding the recent commits: 

I've enabled the use of Unicode Pi - but I've noticed that spacing exists around constants: e.g, in the formatted output, you get results such as `3 dozen`, `3 e` etc...
Would you prefer it if it were formatted with spaces for words (so keep `3 dozen`), but for letter constants there would be no spacing (`3e`)?

Also, I noticed that on large values of factorial (`50!` plus), the result wasn't wrapping on the browser formatted output so I couldn't see many of the digits. @mintsoft is aware of my current plight with incorrect formatting, so if anyone else experiences this, please let me know!

jophab:
@GuiltyDolphin I prefer "3 dozen" and "3e" . :+1:

lib/DDG/Goodie/Calculator.pm
moollaza: Yup. All of that is now handled by the IA Page:
https://duck.co/ia/view/calculator -- I can add you to the contributors list

GuiltyDolphin: @moollaza That'd be great - I'll remove the appropriate lines.

mintsoft:
@moollaza I'll pass this back to you once it's ready to review?

GuiltyDolphin:
@mintsoft, @jophab Do you have any preferences when it comes to trig functions (in terms of degrees or radians).

I have three choices:
 - Prefer radians over degrees and only take input as degrees if they specify that it is.
 - Prefer degrees over radians and only take input as randians if they specify that it is.
 - Assume radians for small values (up to 4 pi or so?), but above that assume the input is degrees. Then allow specifying degrees or radians if they need to override.

Other possibilities would just be a variation on the last option.

mintsoft:
@GuiltyDolphin I'd like it to assume rads unless explicitly told otherwise:
`sin(0.5)` is in rads `sin(0.5°)` is in degrees

jophab:
@GuiltyDolphin @mintsoft In case which not specifies whether degree or rad,
I would prefer both. Take the  given value as both rad and degree and show the results for both.
We dont know what the user expect. What do you think?

GuiltyDolphin:
@jophab I'd rather not do that - it would complicate the results somewhat. I agree with the principle - but speaking from experience, radians are more likely to be useful as they are more mathematically useful (they have certain properties that do not hold for degrees). If you look at what [Wolfram Alpha does](https://www.wolframalpha.com/input/?i=sin%283.14%29), it says that it has interpreted the input as `radians` or `degrees`. Would you say that something along those lines would be more suitable?

jophab:
@GuiltyDolphin 
What I think is that radians plays with pi.
Since a user has no way to enter the exact value of pi the result will not be "correct".
We will have to use the approximately equal to symbol here also as in case of fractional answers.
What do you think?

jophab:
@GuiltyDolphin About Displays I would prefer sin(45°) and r for radians

mintsoft:
@jophab @GuiltyDolphin I'd definitely not like "both" displayed as it's too confusing.
I think it makes the most sense to display the parsed output as suggested. I'd probably say `rads` as the unit for radians and `°` for degrees.

I'd always push for radians as they are always more useful; beyond highschool mathematics all interactions with trig functions will be in radians. All physical sciences (Physics, Chemistry) rely on the properties of radians for essentially every calculation.

Assuming that users who have measured an angle with a protractor (which is the only situation in which degrees are really used) then I would argue that it's acceptable as long as it's clear in the output how the numbers have been parsed and it's obvious how to modify the input to get the desired affect.

There is more difficulty around `arcsin`; we need a way of providing the user a method of supplying which unit they want the output in.

The only way I can think of doing it is to provide 2 functions with different names

GuiltyDolphin:
@mintsoft See the new `modifiers` suggestion under the functionality to-dos. I suggest that the user may be able to supply modifiers such as `arcsin(1/2) in degrees` etc...

mintsoft:
@GuiltyDolphin I like that idea; as long as the grammar remains unambigous. I was also toying with some sort of function which could wrap the `arcsin`; like:  `to_degrees(arcsin(pi/2))` I think it's probably worth having that, but not as the only option. 

with a query like `arcsin^2(0.4)` is the user supposed to use: `(arcsin(0,4) in degress)^2 ? Granted arcsin isn't used that much in this situation.

GuiltyDolphin:
@mintsoft As it stands, `arcsin^2(0.4)` is not a valid query - it'd have to be `arcsin(0.4) ^ 2` (at least, once I've implemented inverse trig functions!). And yes, the grammar would remain unambiguous as the modifier would be at the end of the expression and act upon the whole calcualtion (probably). It doesn't make much sense to wan't to use degrees in calculations.

mintsoft:
@GuiltyDolphin Yeah that sounds fine. Thinking about it, I don't think I've ever used `arcsin^2(x)`; however I do think that `sin^2(x)` should be supported

GuiltyDolphin:
@mintsoft Maybe - from a brief thought it'll probably make the grammar a lot more complicated, as this would mean being able to inject powers into functions. I'll have a look though, and if I can come up with a better way I'll have a poke.

mintsoft:
@GuiltyDolphin Yeah again with reflection there's a slight problem here anyway; in general `f^2(x)` means `f(f(x))`, yet trig functions are an exception (¬_¬) where it means `f(x)^2)`

GuiltyDolphin:
@mintsoft Also for inverse functions - that would be a bit of a pain. I'll look at function powers later on, for now I think they are fairly low priority.

mintsoft:
I agree it can be dealt with if it's ever actually needed

GuiltyDolphin:
@mintsoft, @moollaza, @tagawa I was just reading [this](https://duck.co/forum/thread/13748/calculator-improvements) post which describes some of the potential features the calculator could have - how much of this do you guys want implemented?

It seems that a couple of big features would be solving for variables and support for complex numbers.

mintsoft:
@GuiltyDolphin I'd say that the calculator should simply calculate, not solve equations or work with variables. We have symbolab for complex things etc.

GuiltyDolphin:
@mintsoft Yeah... I don't really want to have to try and get it to solve them ^^. What are your thoughts about complex numbers?

mintsoft:
@GuiltyDolphin I don't really see a problem it having support for them, however given the size of this rewrite etc I'm keen to get it merged and reviewed etc so that any snagging issues can be resolved sooner rather than later when it will be more difficult

cc @moollaza

GuiltyDolphin:
@mintsoft Yeah, it is a bit large ^^.

I've mostly achieved what I was aiming at - it is much more robust than the old calculator and adding new features can be as little as one line for simple functions.

Adding Complex number support would probably mean factoring out values from results, but I can do that in a future revision if we want it.

mintsoft:
@GuiltyDolphin Yeah I agree, I think new features we can look at adding in post-release 1.0 as it were

GuiltyDolphin:
@mintsoft, @moollaza I'm starting to add a list of features to the PR so people know what this adds - it's at the bottom under the heading `Features`.

mintsoft:
@GuiltyDolphin Thats excellent thanks, makes it much clearer to see what we need to test etc

GuiltyDolphin:
@moollaza Can we get a re-deploy? I've added a limiter to the displayed output so hopefully factorials won't be scary anymore ^^

mintsoft:
Myself and @GuiltyDolphin have been discussing this for most of the afternoon and we think we're in a state where this is ready for final review. There are some minor white-space changes to the Styler role which is why there are changes to `Conversions` and so forth.

I'm relatively keen to get this merged in before it becomes too unwieldy. There are some additional changes that it would be nice to have implemented however I believe it would be best to tackle them in a separate PR.

cc @jagtalon @moollaza @duckduckgo/community-leaders

GuiltyDolphin:
@mintsoft Yeah - I'm happy for any additional changes/features (recurring decimal notation maybe?) to be in separate PR's for the future.

I've got pretty high ~ 100% coverage of the code through tests, so there shouldn't be any major bugs.

mintsoft:
@GuiltyDolphin I've spoken to a German colleague who *thinks* that they use semicolon to delimit between arguments to functions

GuiltyDolphin:
@mintsoft Alrighty - I'll keep it as the delimiter for now, maybe it'll change in a future PR.

mintsoft:
@GuiltyDolphin just noticed there are conflicts on this PR now

GuiltyDolphin:
@mintsoft Mhmm, I can rebase, but I was going to wait until final review - prepare for a bunch of commits!

mintsoft:
@GuiltyDolphin well it looks good to me, just waiting sign-off from @moollaza !

GuiltyDolphin:
@mintsoft I did warn ;)

That should deal with the conflict.

mintsoft:
hmm @GuiltyDolphin The tests are failing now:https://travis-ci.org/duckduckgo/zeroclickinfo-goodies/builds/103435122

GuiltyDolphin:
@mintsoft I can't replicate the failing `Calculator` tests either locally or on Codio... Any ideas?

GuiltyDolphin:
@mintsoft, @moollaza The failing tests are not caused by `Calculator` to my knowledge.

I've tried various things, and finally I've just done a Travis CI build on the last time the Calculator passed the tests - this time it failed.

I would imagine this is either a dependency issue or a change in `DDG` in the last 19 days.

I'll keep looking - but I've not been able to find out much more.

GuiltyDolphin:
@moollaza, @mintsoft Fixed it!

That was rather stressful! - Dependency issue :angry:

mintsoft:
@GuiltyDolphin Glad to hear, sorry I didn't get a chance to help you out with it

lib/DDG/Goodie/Calculator.pm
```diff
@@ -1,180 +1,1340 @@
...
+    } elsif ($self->is_radians) {
+        return ' ㎭' if $self->declares('radian');
+    };
+    return '';
+}
+
+sub declare {
+    my ($self, $to_declare) = @_;
+    my $copy = $self->copy();
+    push $copy->{declared}, $to_declare;
```
mintsoft: Ooh this is a little bleeding edge, I'm surprised nothing's complained about it

mintsoft: Edit: It does I've never noticed it:
`push on reference is experimental at lib/DDG/Goodie/Calculator.pm line 367.`

lib/DDG/Goodie/Calculator.pm
```diff
@@ -1,180 +1,1340 @@
...
+
+sub contains_bad_result {
+    my $self = shift;
+    return $self->value() =~ /(inf|nan)/i;
+}
+
+__PACKAGE__->meta->make_immutable;
+
+
+package DDG::Goodie::Calculator::Parser::Grammar;
```
mintsoft: I feel we should break the packages up a little; you could move them into a directory called `Calculator` within Goodie then they'd be `DDG::Goodie::Calculator` by lib path

mintsoft:
@moollaza I've given this a good runaround; the only thing I can see is the push on a reference which isn't future-safe and the sheer size of it. maybe it'd be worth breaking into submodules however @GuiltyDolphin and I have discussed this previously and came to the conclusion that as we're planning to refactor it into a Role which can be used in other IA's etc (particularly for Conversions for queries like 4m + 12inches in yards) I'm not sure it's necessary at this stage. Other than that it LGTM

lib/DDG/Goodie/Calculator.pm
GuiltyDolphin: @mintsoft I've tried this after speaking with @moollaza but it breaks the build due to the way IAs are detected for Goodies. I'd be happy to split it out if that can be fixed.

lib/DDG/Goodie/Calculator.pm
GuiltyDolphin: @mintsoft Huh.. I'm not seeing that. I'll come up with a different solution.

GuiltyDolphin:
@mintsoft I've fixed that experimental feature - should've probably seen it in the docs!

@moollaza Yeah, I've been working on a `Grammar` role (mostly playing around atm) which I hope to get ready for Goodies eventually.

lib/DDG/Goodie/Calculator.pm
mintsoft: Yeah that's fine, I think we can kick all this around post-merge for this anyway, so no problem

mintsoft:
@GuiltyDolphin Excellent thanks, LGTM then!

moollaza:
@mintsoft @GuiltyDolphin thanks for all the updates and great work!

I'll take a dive into this shortly. It's been running now for a few weeks on Beta and there haven't been any fires so that's a god sign ;)

jdorweiler:
I'm still not sure why the beta installs are failing.  I'd just stick to testing this locally until I can figure it out.

mintsoft:
@moollaza @jdorweiler I just noticed there's a slight bug with Dax/github, the link for "deployed to beta" hasn't uriencoded the + in the example query so it adds a space instead of a + literal

lib/DDG/Goodie/Calculator.pm
```diff
@@ -1,180 +1,1341 @@
...
+    $query =~ s/^\s*(?:what\s*is|calculate|solve|math)\s*//;
+    my ($generated_input, $result) = to_display $query or return;
+    return unless defined $result && defined $generated_input;
+    return $result,
+        structured_answer => {
+            id   => 'calculator',
+            name => 'Answer',
+            data => {
+                title    => "$result",
+                subtitle => "Calculate: $generated_input",
```
mintsoft: Is this already `html_enc`'d by this point btw? With the outstanding bug we'll need it htmlencoded

GuiltyDolphin: @mintsoft I don't think the calculator can take or output that kind of data - it goes through a lot of processing.

mintsoft: Looking at the grammar; I agree, it shouldn't be required.

mintsoft:
@moollaza We seem to be getting a good back log of issues against the calculator that are fixed by this rewrite. I've given it a roundaround and I think it's ready to go?

GuiltyDolphin:
@moollaza To add to that I'm actively working on the Interactive Calculator on #2848 so we can sort any bugs/additional features over there - I'm refraining from modifying this version because I have to add the changes manually over there now (histories differ, I've rewritten some parts).

mintsoft:
@GuiltyDolphin I've just cloned fresh to give this another look over and I'm getting:
```
rob@set $ duckpan server
Checking for DuckPAN requirements...
Checking perl version...
Checking for latest App::DuckPAN...
installing DDG at /home/rob/perl5/lib/perl5/App/DuckPAN.pm line 568.
Checking for latest DDG Perl package...
Checking for ssh...
Checking for git...
Loading Instant Answers...
Can't locate object method "triggers_block_type" via package "DDG::Goodie::Calculator::Parser" at /home/rob/perl5/lib/perl5/App/DuckPAN/DDG.pm line 85.
```

lib/DDG/Goodie/Calculator.pm
```diff
@@ -1,180 +1,217 @@
...
+    my @numbers = grep { $_ =~ /^$number_re$/ } (split /[^\d,.]+/, $text);
+    return number_style_for(@numbers);
+}
 
-        $
-        >xi;
+sub get_currency {
+    my $text = shift;
+    # Add new currency symbols here.
+    $text =~ /(?<currency>[\$])$decimal/;
```
mintsoft: `\p{Sc}` should be all currency symbols in one class

lib/DDG/Goodie/Calculator.pm
```diff
@@ -1,180 +1,217 @@
...
+    return 1 if ($query =~ /\b0\s*x/);
+    # Probably want to talk about addresses, not calculations.
+    return 1 if ($query =~ $network);
+    return 0;
+}
 
-        [\( \) x X × ∙ ⋅ * % + ÷ / \^ 0-9 \. , _ \$ -]* =?
+sub get_style {
+    my $text = shift;
+    my @numbers = grep { $_ =~ /^$number_re$/ } (split /[^\d,.]+/, $text);
```
mintsoft: Hmm you could just $text =~ /($number_re)/g;? Unless there's a specific reason you want to split it out?

lib/DDG/Goodie/Calculator.pm
GuiltyDolphin: @mintsoft I'm thinking of removing currencies - it'd be best if we had a dedicated currency calculator that could handle conversion rates etc. possibly by plugging into the calculator back-end as the Interactive Calculator does.

mintsoft: Yeah I agree, it feels like a level of complexity that we really don't need

lib/DDG/Goodie/Calculator.pm
GuiltyDolphin: Uhh... No, that should be fine. I'll update #2848 with your various comments (unless there's anything major that needs to be fixed).

GuiltyDolphin:
@mintsoft Oh come on.... I thought the multi-package version was working now!

mintsoft:
@GuiltyDolphin I swear to god it was last time I looked! The tests pass etc ¬_¬

GuiltyDolphin:
@moollaza Can we move forward with this? It fixes a bunch of open issues and there's WIP being held up by it; @mintsoft has given it a decent review, and AFAIK it  doesn't *break* anything.

It's been 6 months and we've received several issues with the *existing* calculator in this time which are *already* fixed by this PR.

cc @duckduckgo/community-leaders does anyone have any dire comments, are you guys all happy with this?

mintsoft:
I'm happy as long as the performance/dependency aspect is OK; I can't actually test that in any sensible way currently. @GuiltyDolphin The tests are failing currently; looks like a missing dependency. Can we resolve that?

GuiltyDolphin:
@mintsoft Yeah... I think most of the bottlenecks are from Marpa (though some other aspects could probably be improved); I might be able to replace it with a custom regex-based parser seeing as the definitions should just plug into whatever back-end we give it (hopefully).

Regarding the dependency, not sure why that's failing now and wasn't before - it probably just needs installing on prod or something.

mintsoft:
@GuiltyDolphin IIRC the dependencies for Travis are installed through: https://github.com/duckduckgo/zeroclickinfo-goodies/blob/master/dist.ini#L18

GuiltyDolphin:
@mintsoft Hmm, might not be detecting it due to the custom `AutoPrereqs`.

zachthompson:
This failed on [listdeps](https://travis-ci.org/duckduckgo/zeroclickinfo-goodies/builds/132562706#L336).  We currently don't support IAs with module hierarchies.  Will need *at least* [this change](https://github.com/duckduckgo/p5-dzp-iachangelog/pull/11) to be considered for release.

GuiltyDolphin:
@zachthompson Ah, yeah - I had issues with it before; I recently put it back into the heirarchy and it seemed to be working (at least temporarily).

Thanks - support for hierarchies would be sweet!

zachthompson:
@moollaza tests are good now.

moollaza:
It's live on beta! https://beta.duckduckgo.com/?q=3+*+%2434.45&ia=answer

zachthompson:
[1.03 + 2.05 / 3.1](https://beta.duckduckgo.com/?q=1.03+%2B+2.05+%2F+3.1&ia=answer)

zachthompson:
[(5^5)^5](https://duckduckgo.com/?q=(5%5E5)+%5E+5&t=ffab&ia=answer) works on the site but not beta.

I was benching this with the existing calculator using two random operators - `+ - / * x` -  and three random numbers < 5. It's about 20x slower: [old](https://zach.duckduckgo.com/nytprof/old-calc/Calculator-pm-8-line.html#71) vs. [new](https://zach.duckduckgo.com/nytprof/new-calc/Calculator-pm-8-line.html#194).  That doesn't necessarily concern me, though we should be aware of it.

However, when I add exponents to the possible operators the new calc starts randomly pegging the cpu on things like [3.66707834218797 ^ (4.38132278605936 - 4.84952590339233)](https://beta.duckduckgo.com/?q=3.66707834218797+%5E+(4.38132278605936+-+4.84952590339233)&t=ffab&ia=iatesting) whereas the [existing calc](https://duckduckgo.com/?q=3.66707834218797+%5E+(4.38132278605936+-+4.84952590339233)&t=ffab&ia=answer) does not.  The existing also burns through all [1000 random queries](https://zach.duckduckgo.com/nytprof/old-calc-exp/Calculator-pm-8-line.html#71) consistently.

GuiltyDolphin:
@zachthompson I think there's a fairly comprehensive spec for the output in the PR main description.

Thanks (example `(5^5)^5`); I'll check the grammar. ***Edit:*** Hmm, nope, it's working fine locally - are you able to reproduce this locally (`duckpan query`)? 

Yeah - I'm aware of the slowness; hopefully it is something we can fix (or improve) in the future - it's mostly Marpa, I believe.

Hmm... I'm using `Math::BigRat` (shouldn't be an issue) and `Math::BigFloat` (I've had issues with this) which might be causing CPU issues. 

By the way @zachthompson - I've also got the Interactive PR: https://github.com/duckduckgo/zeroclickinfo-goodies/pull/2848 which changes some of this (for example it uses a different method for dealing with approximate results). Once this is merged I plan to do a PR to update the result output to better match that PR so we can extend it without screwing that up.

@zachthompson It's weird that everything is working fine locally but some simple queries seem to be screwing up on Beta. @moollaza @zachthompson Could you check that `Math::BigRat`, `Math::Cephes`, `Math::BigFloat`, `Math::BigInt` are all up to date on Beta? Could you also check that `GMP` is available for `Math::BigRat`? (I imagine it'd be quite slow if not).

moollaza:
@GuiltyDolphin @zachthompson Updated Beta. Mostly looks like minor changes, but it seems `Math::BigInt::GMP` may not have been installed at all?

```
Successfully installed Math-Cephes-0.5305 (upgraded from 0.5304)
Successfully installed Math-BigInt-1.999722 (upgraded from 1.999715)
Successfully installed Math-BigRat-0.260804 (upgraded from 0.260801)
Successfully installed Math-BigInt-GMP-1.51
```

`BigFloat` was up to date.

GuiltyDolphin:
@zachthompson That seems to have fixed most of the query problems, but I'm getting an error for https://beta.duckduckgo.com/?q=3.66707834218797+^+%284.38132278605936+-+4.84952590339233%29&t=ffab&ia=iatesting - can you replicate?

GuiltyDolphin:
@zachthompson It was an issue with GMP unfortunately, I've fixed it (though some answers may not give exact fractions where otherwise they would :( ) - let me know if you find any 'dodgy' queries.

GuiltyDolphin:
@moollaza @mintsoft @zachthompson Heh, right, I re-wrote a bunch of the tracking behavior so it should be much more robust and accurate now.

Can we get a deploy? Hopefully this'll fix most (if not all) the current bugs on beta.

GuiltyDolphin:
@zachthompson Would you mind trying out some of those queries again? I've just tried out with the updates and everything seems to be working fine :)

Benestar:
Is there any kind of progress on this PR? The current version of DuckDuckGo calculator has very limited functionality and it would be great to get an improved version.

mintsoft:
@GuiltyDolphin FYI: this has conflicts again :(

GuiltyDolphin:
@mintsoft It's just going to be a case of overwriting everything AFAIK, so I'll wait. Hopefully no new features or functionality changes will occur in other PRs that require updates to this!

Benestar:
So it's even more important that this gets reviewed + merged asap, otherwise all PRs changing the calculator need to be stalled.

mintsoft:
@GuiltyDolphin do you want to do the NumberRole changes separately from this? It'd probably resolve the conflicts (with conversions for example) if we get that into master first?

GuiltyDolphin:
@mintsoft Good call! I'll pull out the changes later.

moollaza:
We've been discussing the future of the calculator internally and are currently progressing towards an interactive front-end implementation using Math.js to handle the evaluation of expressions.

We have decided that moving forward we don't want to maintain a complex Perl backend when we have a frontend that's able to provide the same results.

I am going to close this PR, and in the future, we expect to deprecate most or all of the Perl logic. We may decide to maintain a very simple Perl backend for /html and the API, but for now, those aren't a priority.

FWIW, almost all of the queries the calculator currently triggers on are simple expressions with only one or two basic operators (i.e. `+, -, *, /`).

Thanks very much to @GuiltyDolphin and @mintsoft for working on this. I'm sorry we can't use this, but I see no reason why you can't carry this forward to CPAN and offer a comprehensive calculator library. Nothing of this nature seems to exist there currently!

PR #1956, opened 2015-12-25
Passphrase: Fixes #1144

### LINKED ISSUE #1144: Passphrase: Isn't triggering properly
The passphrase IA isn't triggering properly. It's [supposed to trigger](https://github.com/duckduckgo/zeroclickinfo-goodies/blob/37ee5f2d3a03555e339f3215c063d868d2f09dd8/lib/DDG/Goodie/Passphrase.pm#L19) when a user searches any of the phrases/words listed on line 19, but instead, "random passphrase" is the only search that triggers it. 

IA: https://duck.co/ia/view/passphrase

cngarrison:
All 'startend' triggers are compared to query when being handled. If query is an exact match with trigger (no remainder), then assume 4 words. 

IA Page: https://duck.co/ia/view/passphrase


daxtheduck:
@cngarrison Hey!
        
While the team and the community review your submission, we'd like to make sure you look good! If you'd like a free DuckDuckGo tshirt, please send over the following info to merch@duckduckgo.com, and be sure to put "Dev Merch" in the subject line:

Name:
Street Address:
City:
State / Province:
Country:
Phone (Non-US Orders only): 
T-shirt size:
GitHub ID:
Link to this Pull Request:

Thanks!

daxtheduck:
Metadata post coming soon

jagtalon:
@cngarrison excellent! 

giphy

moollaza:
@jagtalon @cngarrison looks like we aren't seeding the `rand` in `handle` so this IA isn't generating random results!

@cngarrison I know you didn't introduce this bug, but would you mind fixing it? 

You need to add ` srand();` to the top of the handle so we seed the function each time it runs.

cngarrison:
@jagtalon @moollaza I don't believe `srand` is the correct solution. Note this from the pod:

> Do not call srand() (i.e., without an argument) more than once per process. The internal state of the random number generator should contain more entropy than can be provided by any seed, so calling srand() again actually loses randomness.

My understanding is Goodies are run in persistent processes. So we don't want to just call srand, at random. ;-)

Another recent project of mine (raffle system for our local lawn bowls club) needed a reliably random generator, and I chose [Data::Entropy](https://metacpan.org/pod/Data::Entropy) and [Math::Random::MT](https://metacpan.org/pod/Math::Random::MT) to solve it for that project. Are those modules an acceptable dependancy to add? The real issue is solved with just Data::Entropy, but it's solved more elegantly with both modules.

I've pushed my suggested changes using those modules. Hmm, I've pushed to a different branch (didn't realise this PR was using master when I started) and I don't know how to 'link' the other branch to this PR. Should I just create a new PR? The commit is:

https://github.com/cngarrison/zeroclickinfo-goodies/commit/e7f74882dc36ce195861a3b620e0aab50e0d799b

moollaza:
@cngarrison you're sorta correct -- Goodies are run in a persistent process, but we have several processes running to handle IAs and the problem we ran into a while back was that they all had the same seed and so as the request was handled by different processes, the same results were being seen.

As a simple fix, we added in `srand()` to the `handle` to ensure random results.

As you've pointed out this isn't the best solution, and `rand` isn't the best random number generator. Moving to something more robust would be great. Adding the dependency is not a problem, we just need to add it to Beta before we can test, and our production stack before we merge the changes.

Would you be interested in improving the "randome" Goodies? We could potentially add a helper function so that all the IA's don't need to add in the `use` statements:

https://github.com/duckduckgo/duckduckgo/blob/0690bb6d666a47df29e65ad93ad1c550ee681957/lib/DDG/Meta/Helper.pm

cngarrison:
@moollaza Agreed, a helper function looks like a good idea. I don't believe the helper function should override the builtin `rand` (& `srand`), but it would mean it becomes a drop-in replacement and all existing Goodies would immediately benefit. 

I found 13 Goodies which use `[s]rand` directly, and 5 Goodies which use 3rd module to generate random data (e.g. `Fortune`). There could be others I missed. Which is why I said I don't think `rand` should be overridden. It also means there could be other modules depending on `rand` and will have the same problem. We can fix the Goodies easily enough, but it doesn't fix __all__ places where random data is generated. 

For now, I'm creating a `rand_int` helper. I'll also start reviewing changes for all the Goodies to use the new helper function.

moollaza:
>For now, I'm creating a rand_int helper. I'll also start reviewing changes for all the Goodies to use the new helper function.

Perfect! Thanks a lot @cngarrison

PR #1960, opened 2015-12-26
URLEncode: Fixes #1932

### LINKED ISSUE #1932: URLDecode: Does not correctly decode German umlauts
`äöü` [encoded](https://duckduckgo.com/?q=url+encode+%C3%A4%C3%B6%C3%BC&ia=answer) are `%C3%A4%C3%B6%C3%BC`

Try it: https://duckduckgo.com/?q=url+decode+%25C3%25A4%25C3%25B6%25C3%25BC&ia=answer
It returns this: `Ã¤Ã¶Ã¼`

------
IA Page:  http://duck.co/ia/view/urldecode

cngarrison:
IA for URLEncode was not handling UTF8 correctly, so round-trip values for URLEncode/URLDecode didn't match. Changed from `en|decodeURIComponent` to `uri_(un)escape` to handle unicode characters. `URI::Escape::XS::uri_escape` expects a byte string, so downgrade the string before encoding.

Tests are comparing HTML entities rather than raw utf8 characters since IA isn't using templated answers. 

IA Page: https://duck.co/ia/view/urlencode

### DIFF
--- a/lib/DDG/Goodie/URLDecode.pm
+++ b/lib/DDG/Goodie/URLDecode.pm
@@ -2,7 +2,7 @@ package DDG::Goodie::URLDecode;
 # ABSTRACT: Displays the decoded URL for a percent encoded uri
 
 use DDG::Goodie;
-use URI::Escape::XS qw(decodeURIComponent);
+use URI::Escape::XS qw(uri_unescape);
 use warnings;
 use strict;
 
@@ -35,7 +35,7 @@ handle query_raw => sub {
     s/(^\s+)|(\s+$)//;
 
     my $in      = $_;
-    my $decoded = decodeURIComponent($in);
+    my $decoded = uri_unescape($in);
 
     if ($decoded =~ /^\s+$/) {
         $decoded =~ s/\r/CReturn/;
--- a/lib/DDG/Goodie/URLDecode.pm
+++ b/lib/DDG/Goodie/URLDecode.pm
@@ -45,13 +45,20 @@ handle query_raw => sub {
     }
 
     my $text = "URL Decoded: $decoded";
+    my $subtitle = "URL decode: $in";
 
-    return $text,
-      structured_answer => {
-        input     => [html_enc($in)],
-        operation => 'URL decode',
-        result    => html_enc($decoded)
-      };
+    return $text, structured_answer => {
+        id => 'url_decode',
+        name => 'Answer',
+        data => {
+            title => $decoded,
+            subtitle => $subtitle
+        },
+        templates => {
+            group => 'text',
+            moreAt => 0
+        }
+    };
 };
 
 1;
--- a/lib/DDG/Goodie/URLEncode.pm
+++ b/lib/DDG/Goodie/URLEncode.pm
@@ -2,7 +2,7 @@ package DDG::Goodie::URLEncode;
 # ABSTRACT: Displays the percent-encoded url.
 
 use DDG::Goodie;
-use URI::Escape::XS qw(encodeURIComponent);
+use URI::Escape qw(uri_escape);
 use warnings;
 use strict;
 
@@ -29,7 +29,7 @@ handle remainder => sub {
 
     return unless $in;
 
-    my $encoded_url = encodeURIComponent($in);
+    my $encoded_url = uri_escape($in);
 
     my $text = "Percent-encoded URL: $encoded_url";
 
--- a/lib/DDG/Goodie/URLEncode.pm
+++ b/lib/DDG/Goodie/URLEncode.pm
@@ -36,13 +36,20 @@ handle remainder => sub {
     my $encoded_url = uri_escape($in);
 
     my $text = "Percent-encoded URL: $encoded_url";
+    my $subtitle = "URL percent-encode: $in";
 
-    return $text,
-      structured_answer => {
-        input     => [html_enc($in)],
-        operation => 'URL percent-encode',
-        result    => html_enc($encoded_url),
-      };
+    return $text, structured_answer => {
+        id => 'url_encode',
+        name => 'Answer',
+        data => {
+            title => $encoded_url,
+            subtitle => $subtitle
+        },
+        templates => {
+            group => 'text',
+            moreAt => 0
+        }
+    };
 };
 
 1;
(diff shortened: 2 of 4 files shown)


daxtheduck:
Metadata post coming soon

GuiltyDolphin:
@cngarrison Thanks for doing the PR!

@moollaza If the `URLEncode` and `URLDecode` Goodies need to be so compatible with each other, might it be wise to merge them into one goodie that handles encoding and decoding?

cngarrison:
Makes sense to me to combine them; shall I do that? Although I don't think they need to be compatible (or combined). It was really just a problem with URLEncode, and tests added for URLDecode. The two Goodies are similar so I treated them as a pair. 

OK, URLDecode can be triggered on just encoded characters (%xx) in query, so keep the goodies separate, otherwise triggers get more complicated.

mintsoft:
@cngarrison I'd suggest that it's OK to roll both into the same PR as long as the number of changes is relatively small

mintsoft:
@cngarrison This all looks fine to me, are you happy with the tests etc now?

lib/DDG/Goodie/URLEncode.pm
```diff
@@ -29,7 +29,11 @@ handle remainder => sub {
 
     return unless $in;
 
-    my $encoded_url = encodeURIComponent($in);
+	## URI::Escape::XS::uri_escape expects a byte string, so downgrade our string
```
mintsoft: Nit: Can you fix the tabbing here? We've settled on 4 spaces/tab

GuiltyDolphin:
@mintsoft Yeah - I just noticed that one uses Regex triggers and the other uses word triggers. Would be a bit harder to combine without it getting messy!

t/URLDecode.t
```diff
@@ -101,6 +102,14 @@ ddg_goodie_test(
             result    => 'https://duckduckgo.com/'
         }
     ),
+    '%E4%F6%FC' => test_zci(
+        "URL Decoded: äöü",
+        structured_answer => {
+            input     => ['%E4%F6%FC'],
+            operation => 'URL decode',
+            result    => '&auml;&ouml;&uuml;'
```
mintsoft: @cngarrison As we discussed earlier, the only problem now is that we need to change the results so that it expects utf-8 output rather than html entities.
This should be:
```perl
'%E4%F6%FC' => test_zci(
        "URL Decoded: äöü",
        structured_answer => {
            input     => ['%E4%F6%FC'],
            operation => 'URL decode',
            result    => 'äöü'
    }
),
```
I'll put a comment further up indicating where the tweak is needed.

t/URLEncode.t
```diff
@@ -85,6 +86,15 @@ ddg_goodie_test(
         }
     ),
 
+    'äöü escapeurl' => test_zci(
+        "Percent-encoded URL: %E4%F6%FC",
+        structured_answer => {
+            input     => ['&auml;&ouml;&uuml;'],
```
mintsoft: @cngarrison Regarding my previouscomment, it was on the wrong file. This one is the one that needs to be not `html_enc` (https://github.com/duckduckgo/zeroclickinfo-goodies/pull/1960/files#diff-97a2bf2e8692e549cba1b224a3332f7dR42)

If you'd rather leave this for another PR, that's also fine we can tackle it separately.

lib/DDG/Goodie/URLEncode.pm
cngarrison: Tabbing has been fixed, thanks for pointing it out.

t/URLEncode.t
cngarrison: @mintsoft Either I don't understand the proposed fix or it's not working as expected. If I remove `html_enc` from the `structured_answer` and change test file to compare against utf8 characters instead of html entities, then test fails. I'm happy to keep working on it, but agree maybe best as another PR.

GuiltyDolphin: @cngarrison Have you switched the Goodie over to use a template?

cngarrison: @mintsoft I haven't switched it, but I will. Thanks for the example to follow.

mintsoft: @cngarrison It works for me as long as I change the other goodies too. As @GuiltyDolphin mentions, switching to a template would involve this being changed anyway and would too resolve the problem.

If you're curious what I did is here: https://github.com/duckduckgo/zeroclickinfo-goodies/commit/3c3db2de2c26e3c5a2f5c58003d430b7461fa36a

cngarrison: Yep, I was partially confused with your proposed change since I thought `html_enc` was required for `structured_answer`. I'm working on templated version today.

cngarrison:
@mintsoft Goodies switched to using templates. All tests are passing and html entities have been removed from tests.

mintsoft:
@cngarrison Absolutely perfect! Many thanks for the fix and battling through the confusion around these IA's :+1:

mintsoft:
Also thanks for helping out @GuiltyDolphin !

lib/DDG/Goodie/URLDecode.pm
```diff
@@ -45,13 +45,20 @@ handle query_raw => sub {
...
-        input     => [html_enc($in)],
-        operation => 'URL decode',
-        result    => html_enc($decoded)
-      };
+    return $text, structured_answer => {
+        id => 'url_decode',
+        name => 'Answer',
+        data => {
+            title => $decoded,
+            subtitle => $subtitle
```
moollaza: @cngarrison in light of https://github.com/duckduckgo/zeroclickinfo-goodies/pull/1888#issuecomment-169523789

Can you please update these IAs to `html_enc()` the `subtitle`?

cngarrison: @moollaza I'm confused, I thought templates should get passed *unencoded* data, from #1888:

> Once the content is encoded it get's directly passed to the template. All the templates are setup to escape their input (i.e. they use {{ }} instead of {{{ }}}).

I'm happy to make the change of course, but I would like to understand why the change is needed (so I can learn).

moollaza: @cngarrison that's what I was hoping we could do as well, but for now, due to the fact that we take the `structured_answer` as-is, and convert it to JSON that we put in the page, the JSON is able to break, so to prevent that, we need to encode the content inside the Goodie (for now!).

As a result, we have a double encoding problem that we'll be fixing. I'd live to see us be able to remove `html_enc` :)

cngarrison: Sorry for delay getting back to this; had some $work to attend to. 

Changes made and pushed. I've read #1888 again and understand the issue better. I'll ponder some possible solutions.

PR #1961, opened 2015-12-26
Issue #1163 - Convert ConvertLatLon to use template

Mailkov:
-----
IA Page: https://duck.co/ia/view/convert_lat_lon


daxtheduck:
Metadata post coming soon

mintsoft:
@Mailkov Looks good, the only thing that bothers me is the subtitle:
image

`Convert in <unit>` doesn't quite read right to me, `Convert into <unit>` would work, or `Convert to <unit>` sounds better imo.

moollaza:
Deploying to beta...

lib/DDG/Goodie/ConvertLatLon.pm
```diff
@@ -213,9 +214,14 @@ handle query_nowhitespace => sub {
...
-    my $html = wrap_html(\@queries, \@results, $toFormat);
-
-    return $answer, html => $html;
+    my $result = join(", ",@results);
+
+    return $answer,
+    structured_answer => {
+        input     => [@queries],
+        operation => "Convert to $toFormat",
+        result    => html_enc($result),
```
mintsoft: I don't think this needs `html_enc` as long as the result is returned in unicode

moollaza:
@Mailkov looks like it's triggering on queries with broken numbers, by ignoring the broken parts -- I think we should just ignore the whole query?

e.g. https://beta.duckduckgo.com/?q=71+degrees+10+minutes+3+foo+east+in+decimal&ia=answer

Mailkov:
@moollaza I think is now ok

moollaza:
Thanks! This LGTM :+1:

lib/DDG/Goodie/ConvertLatLon.pm
moollaza: :+1:  I don't think there's risk of any XSS here, we can ditch the `html_enc` if we'd like

Issue #1231, opened 2015-07-04
Dice: 6 times 3d6


yegg:
"Why I cannot roll few dices few times? I cannot write 6 times 3d6 to get skill points roll."

https://twitter.com/mikolajhalber/status/616965696426258432?refsrc=email&s=11

---

IA Page:  http://duck.co/ia/view/dice

loganom:
@zekiel will look into this at the weekend.

loganom:
"this weekend".. I totally spaced on this issue with everything else going on. 

To be clear, the suggestion here is that the user would like to be able to type "6 times 3d6" and the expected result would be 6 multiplied by the sum of the 3 dice rolls?

GuiltyDolphin:
@loganom If it is for a skills-point roll I imagine it would be 6 sums of 3d6 (or clearly displaying 6 separate rolls of 3 dice).

cngarrison:
I made some progress on this, but haven't had time to get back to complete it. Incomplete fix has been pushed to https://github.com/cngarrison/zeroclickinfo-goodies/tree/dice-multi-roll. Not creating a PR for unfinished changes. Feel free to continue with my changes, or start over again. :-)

pjhampton:
Hey all,

[DuckDuckHack is now in Maintenance Mode](https://duckduckhack.com/) and from now on, we are only accepting issues and PRs for essential bugs and bug fixes.

Unfortunately, with the above in mind, this isn't something we can action and will be closed.

We appreciate you taking the time to contribute and apologize for not being able to triage this issue.

PR #1963, opened 2015-12-26
Minecraft: Fixes #1949

### LINKED ISSUE #1949: Minecraft: Need to reign in triggering; any query with "map" & "minecraft" triggers
We need to ensure that the remainder of the query is only the name of the recipe we're looking up. We can use a regex to ensure the query matches our expectation.

Valid queries
- minecraft cake
- mincraft cake recipe
- minecraft how to make cake

bad
- minecraft cake design
- minecraft map guide

e.g https://duckduckgo.com/?q=minecraft+foo+map&t=ffab&ia=answer

/cc @engvik 

------
IA Page:  http://duck.co/ia/view/minecraft

cngarrison:
Logic for triggers and recipe lookup has changed. Loop through all recipe names looking for exact match in query string; then remove all 'extra' words allowed in query. Continue if no remaining words in query.

Added a couple of 'negative' trigger tests; all existing tests still pass.

Fixes #1949 

IA Page: http://duck.co/ia/view/minecraft

### DIFF
--- a/lib/DDG/Goodie/Minecraft.pm
+++ b/lib/DDG/Goodie/Minecraft.pm
@@ -28,23 +28,35 @@ my $decoded = decode_json($json);
 my %recipes = map{ lc $_->{'name'} => $_ } (@{ $decoded->{'items'} });
 
 # Good words: All the words that recipe names consist of.
-# Okay words: Words that are in the good words list, but also could be a part of the query.
+# Okay words: Words that could be a part of the query, but not part of a recipe.
 # Bad words: Words related to Minecraft, but not related to recipes.
 my %good_words = map { $_ => 1 } map { split /\s+/ } (keys %recipes);
-my %okay_words = map { $_ => 1 } (qw(a crafting));
 my %bad_words  = map { $_ => 1 } (qw(download server tutorial mod mods skins skin texture pack packs project projects));
+my @okay_words = ('a','an','in','crafting','recipe','how to make','how to craft','how do I make','how do I craft');
 
 handle remainder => sub {
-    my @query = split /\s+/, lc $_; # Split on whitespaces.
+	my $remainder = $_;
+	# remove the ok words (or phrases)
+	foreach my $ok_word (@okay_words) {
+		$remainder =~ s/\b$ok_word\b//gi;
+	}
+	$remainder =~ s/(^\s*|\s*$)//; # trim leading/trailing whitespace
+    my @query = split /\s+/, lc $remainder; # Split on whitespaces.
     my @lookup;
+    my @unhandled;
 
     # Loop through the query.
     foreach (@query) {
         return if(exists($bad_words{$_})); # Not looking for a recipe.
-        push (@lookup, $_) if(exists($good_words{$_})); # Word exists in a recipe, add it.
+        if (exists($good_words{$_})) {
+        	push (@lookup, $_); # Word exists in a recipe, add it.
+        } else {
+        	push (@unhandled, $_); # otherwise word is not part of recipe, add it to unhandled
+        }
     }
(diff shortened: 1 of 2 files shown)


daxtheduck:
Metadata post coming soon

moollaza:
@cngarrison thanks for this! Just FYI, if you put `fixes #1949` into the actual PR description text, it'll really help us, because then GitHub will automatically close the issue when your PR is merged

cngarrison:
@moollaza Thanks for the note about auto-closing. I'd suggest the docco needs updating; it says to put 'fixed ...' into title, and doesn't mention putting into description.

http://docs.duckduckhack.com/submitting/pull-request.html

Which repo contains the docs; I'll have a look at making change to that page.

moollaza:
>Thanks for the note about auto-closing. I'd suggest the docco needs updating; it says to put 'fixed ...' into title, and doesn't mention putting into description.

D'oh! I forgot that was there. Thanks for letting me know. I'll updated the docs!

lib/DDG/Goodie/Minecraft.pm
```diff
@@ -26,26 +26,29 @@ triggers startend => "minecraft";
...
-    # Loop through the query.
-    foreach (@query) {
-        return if(exists($bad_words{$_})); # Not looking for a recipe.
-        push (@lookup, $_) if(exists($good_words{$_})); # Word exists in a recipe, add it.
-    }
-
-    my $recipe = $recipes{join(' ', @lookup)} || $recipes{join(' ', grep { !$okay_words{$_} } @lookup)};
+	my $recipe;
+	# find recipe name
+	foreach my $recipe_name (@recipe_names) {
```
moollaza: let's make sure this list is sorted from longest phrase to shortest. Also, this is a good place to use `grep` instead of foreach

cngarrison: @moollaza I've added a sort based on length of the phrase. 

I don't see how `grep` is appropriate though. The `grep` block would need to change two variables which are not just getting a list or count from the grep. I can certainly make it work with a grep block but I think it convolutes the meaning.

moollaza: @cngarrison thanks for the comments, looking at this again I'm not entirely sure what I thought this would look like with Grep, but I was thinking of something like this, but I'm not entirely sure this works ;) -- you can leave as-is if you'd like :+1: 

```perl
my ($match) = grep { $remainder =~ s/\b$_//gi } @recipe_names;
my $recipe = $recipes{$match}
```

cngarrison: I prefer the code is as it is. I believe the intent is more clear, and a side benefit is that `last` allows bailing out of loop as soon as recipe is found.

moollaza:
@cngarrison sorry for the delay! This is looking good. One nit: we use 4-space soft tabs for all our Instant Answer repos so in the future please make sure the indentation matches the repos

moollaza:
Hey @cngarrison thanks again for improving the Minecraft Goodie! You did a great job. We at DuckDuckGo, the DuckDuckHack community, and all the people who will encounter this Instant Answer really appreciate it.

We could really use your help on something else now. There was a nice suggestion about Improving the Dice Goodie triggering to allow multuple sets of throws in https://github.com/duckduckgo/zeroclickinfo-goodies/issues/1231 and I think you'd similarly do a great job with it. Please comment on this issue if you'd be willing to take it up!

I think it is a good fit for you, but I'm of course not entirely sure, so I'd love to chat if you're around this week or next. Please message me on slack (@moollaza) or send me an email (moollaza@duckduckgo.com) if you'd be interested in talking.

Thanks again!

PR #1965, opened 2015-12-27
Switching ABC to text template

mintsoft:
Lo,

This one's still using the structured answer so I've converted it to the full template.

There are some slight issues with the tests here, the existing `-ANY-` mechanism doesn't work too well for this, as ideally all components of the `structured_answer` would be asserted and then only the random component could be asserted against a regex (something like : https://github.com/duckduckgo/zeroclickinfo-goodies/commit/4ef7c85ce66f15bb62cfdf474bc029a029818b30) 

https://duck.co/ia/view/abc

### DIFF
--- a/lib/DDG/Goodie/ABC.pm
+++ b/lib/DDG/Goodie/ABC.pm
@@ -46,8 +46,8 @@ handle remainder => sub {
             id => 'abc',
             name => 'Answer',
             data => {
-                title => "$selection",
-                subtitle => "$operation: $choice_string"
+                title => html_enc("$selection"),
+                subtitle => html_enc("$operation: $choice_string")
             },
             templates => {
                 group => 'text',
--- a/lib/DDG/Goodie/ABC.pm
+++ b/lib/DDG/Goodie/ABC.pm
@@ -43,7 +43,7 @@ handle remainder => sub {
 
     return $selection . " (" . $selection_type . ")",
         structured_answer => {
-            id => 'atbash',
+            id => 'abc',
             name => 'Answer',
             data => {
                 title => "$selection",
--- a/lib/DDG/Goodie/ABC.pm
+++ b/lib/DDG/Goodie/ABC.pm
@@ -42,11 +42,18 @@ handle remainder => sub {
     my $operation     = $selection_type . ' selection from';
 
     return $selection . " (" . $selection_type . ")",
-      structured_answer => {
-        input     => [html_enc($choice_string)],
-        operation => $operation,
-        result    => html_enc($selection),
-      };
+        structured_answer => {
+            id => 'atbash',
+            name => 'Answer',
+            data => {
+                title => "$selection",
+                subtitle => "$operation: $choice_string"
+            },
+            templates => {
+                group => 'text',
+                moreAt => 0
+            }
+        };
 };
 
 # The query must look like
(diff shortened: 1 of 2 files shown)


daxtheduck:
Metadata post coming soon

moollaza:
Deploying to beta...

moollaza:
This LGTM, but I need to hold on merging this as we have a JSON encoding issue that we need to solve first:

https://beta.duckduckgo.com/?q=choose+yes+or+no+or+%3C%2Fscript%3E

lib/DDG/Goodie/ABC.pm
```diff
@@ -52,11 +52,18 @@ handle remainder => sub {
...
-        input     => [html_enc($choice_string)],
-        operation => $operation,
-        result    => html_enc($selection),
-      };
+        structured_answer => {
+            id => 'abc',
+            name => 'Answer',
+            data => {
+                title => "$selection",
+                subtitle => "$operation: $choice_string"
```
moollaza: @mintsoft let's add in `html_enc()` here, and then this should be good to go!

mintsoft: @moollaza on just subtitle or title too?

mintsoft:
@moollaza ready to `$rock & $roll`

GuiltyDolphin:
@mintsoft What's going on with the extra commits?

mintsoft:
@GuiltyDolphin GOod question, probably just needs a rebase

mintsoft:
I've just rebased and now it's worse ¬_¬

t/ABC.t
```diff
@@ -8,6 +8,19 @@ use DDG::Test::Goodie;
 zci answer_type => 'choice';
 zci is_cached   => 0;
 
+sub create_structured_answer
+{
+    return {
+        id => 'abc',
+        name => 'Answer',
+        data => '-ANY-',
```
GuiltyDolphin: Hmm... I feel we're now losing information?

Before we were able to check against a fairly specific regex (e.g., `(pick|axe)`), but now we are essentially just testing that it triggers.

Not sure if there is an easy way around this until we get nested comparisons sorted.

GuiltyDolphin:
@mintsoft By the way, what I tend to try and do with any massive rebases/things where it could go wrong is just checkout another branch, do it there, and either duplicate on the main branch or just rename.

t/ABC.t
mintsoft: Yeah exactly; I tried regexing up the individual elements in `data` but it doesn't work unfortunately

t/ABC.t
```diff
@@ -19,67 +32,35 @@ ddg_goodie_test(
...
         'duckduckgo (Non-random)',
-        structured_answer => {
-            input     => ['duckduckgo, google, bing or something'],
-            operation => 'Non-random selection from',
-            result    => 'duckduckgo',
-        }
+        structured_answer => create_structured_answer()
     ),
     'choose Google OR DuckDuckGo OR Bing OR SOMETHING' => test_zci(
         'DuckDuckGo (Non-random)',
```
GuiltyDolphin: Humm... There are a few (like this one) that you should be able to test anyway - the answer is static.

moollaza:
LGTM :+1:

PR #1967, opened 2015-12-27
New GibberishGenerator Goodie

GuiltyDolphin:

It... Generates 'pronounceable' gibberish.

I feel there is a lerge need for more random nonsense in this world (queries likely to be very specific and links would be a hinderance).

N/A

`3 words of French gibberish`, `a sentence of nonsense`.

Recreationalists?

Not to my knowledge.

I don't think it has any collisions.

Mhmm, I couldn't find any good examples of how to test with random data. The norm seemed to be to do a regex match on the output, but the Goodies that did this weren't using templates. I managed to get it working after playing around for a bit, but this is now only testing the result, not the formatted input as well (which I would like).

screenshot from 2015-12-27 14 11 03

## Checklist

```
- [] Added metadata and attribution information (https://duck.co/duckduckhack/metadata)
- [] Wrote test file and added to t/ directory (https://duck.co/duckduckhack/goodie_tests)

    - Windows 10
        - [] Google Chrome
        - [] Firefox
        - [] Opera
        - [] IE 11
        - [] Microsoft edge
        
    - Windows 8
        - [] Google Chrome
        - [] Firefox
        - [] Opera
        - [] IE 10

    - Windows 7
        - [] Google Chrome
        - [] Firefox
        - [] Opera
        - [] IE 8
        - [] IE 9
        - [] IE 10

    - Windows XP
        - [] IE 7
        - [] IE 8

    - Linux
        - [] Google Chrome
        - [X] Firefox
        
    - Mac OSX
        - [] Google Chrome
        - [] Firefox
        - [] Opera
        - [] Safari

    - iOS (iPhone)
        - [] Safari Mobile
        - [] Google Chrome
        - [] Opera

    - iOS (iPad)
        - [] Safari Mobile
        - [] Google Chrome
        - [] Opera

    - Android
        - [] Firefox
        - [] Native Browser
        - [] Google Chrome
        - [] Opera

```

-----
IA Page: https://duck.co/ia/view/gibberish_generator

### DIFF
--- a/share/goodie/gibberish_generator/content.handlebars
+++ b/share/goodie/gibberish_generator/content.handlebars
@@ -3,7 +3,5 @@
         <p>{{this}}</p>
     {{/each}}
 {{else}}
-    <div class="c-info__content chomp js-ellipsis">
-        {{{gibberish_paragraphs}}}
-    </div>
+    {{{gibberish_paragraphs}}}
 {{/if}}
--- a/share/goodie/gibberish_generator/content.handlebars
+++ b/share/goodie/gibberish_generator/content.handlebars
@@ -1,0 +1,9 @@
+{{#if use_paragraphs}}
+    {{#each gibberish_paragraphs}}
+        <p>{{this}}</p>
+    {{/each}}
+{{else}}
+    <div class="c-info__content chomp js-ellipsis">
+        {{{gibberish_paragraphs}}}
+    </div>
+{{/if}}
--- a/lib/DDG/Goodie/GibberishGenerator.pm
+++ b/lib/DDG/Goodie/GibberishGenerator.pm
@@ -43,12 +43,12 @@ sub generate_werds {
     my @modifier_args = @{$languages{$modifier}};
 
     my $werd = Silly::Werder->new();
+    $werd->set_language(@modifier_args);
     my $werds;
     if ($type eq 'sentence') {
-        $werds .= join ' ', map { sentence() } (1..$amount);
+        $werds = join ' ', map { $werd->sentence() } (1..$amount);
     } elsif ($type eq 'word') {
         $werd->set_werds_num($amount, $amount);
-        $werd->set_language(@modifier_args);
         $werds = $werd->sentence();
     };
     return $werds;
--- a/lib/DDG/Goodie/GibberishGenerator.pm
+++ b/lib/DDG/Goodie/GibberishGenerator.pm
@@ -6,6 +6,7 @@ use DDG::Goodie;
 use utf8;
 
 use Silly::Werder qw(sentence get_werd);
+use Lingua::EN::Numericalize;
 
 triggers any => qw(nonsense gibberish);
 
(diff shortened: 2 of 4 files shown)


daxtheduck:
Metadata post coming soon

GuiltyDolphin:
Regarding the speed of the goodie: It seems to slow down quite quickly (bad running time complexity). The package I'm using provides a 'small' version of each language which should be faster, but if anyone can come up with some rough estimates of when it starts becoming too slow I'd appreciate the knowledge (then I can add some limiters).

moollaza:
Deploying to beta for testing!

> Regarding the speed of the goodie: It seems to slow down quite quickly (bad running time complexity). 

@GuiltyDolphin how much does it slow down? longer than 1s? could we simplify the IA by triggering on "random X gibberish" and just return, for example 5 words?

FWIW we've used https://metacpan.org/release/Devel-NYTProf to profile code before to determine where it slows down :+1:

GuiltyDolphin:
@moollaza For example: `500 lines of gibberish` in `duckpan query` gives a noticable wait time before returning the result. 

I'm almost certain it is the package I'm using - I'll use the profiler to see actual wait times.

I'm thinking either hard-code an upper-bound after which it won't trigger, or we could have it display something along the lines of `Displaying X lines instead`? Which would be more appropriate?

moollaza:
@GuiltyDolphin sorry for the delay. For now I'd say let's set some sane limits on the amount of content this can generate for now so that we don't risk running forever or using too many resources (how is this on memory?)

500 lines is quite a bit. I'd say even 20 is probably enough? They can run it a couple of times if needed. If we get demand to increase the limits we can always revise and go from there. For now, lets get a decent, working V1 live for the world to use :+1: 

BTW, we now have some code in-place internally that will stop the Goodies after 1s of execution to protect ourselves. We should still try our best to make sure Goodies can't run forever or consume too much memory.

GuiltyDolphin:
@moollaza I've limited it to 200 words (approximately 25 lines/sentences) so hopefully that'll be enough.

Like you say, if we get requests for more we can either expand this range, or use my other suggestion of falling-back onto fully randomly generated gibberish.

share/goodie/gibberish_generator/content.handlebars
```diff
@@ -0,0 +1,9 @@
+{{#if use_paragraphs}}
+    {{#each gibberish_paragraphs}}
+        <p>{{this}}</p>
+    {{/each}}
+{{else}}
+    <div class="c-info__content chomp js-ellipsis">
```
moollaza: @GuiltyDolphin the content is already wrapped in this when chompContent is enabled -- we’re getting a double wrappe content here which has an extra big top-margin as a result

moollaza: https://github.com/duckduckgo/duckduckgo-answerbar-templates/blob/master/detail/basic_info_detail.handlebars#L16

moollaza: The team is working on fixing the paragraph chomping bug BTW so that should be fixed soon.

moollaza:
@GuiltyDolphin aside from that this LGTM :+1:

share/goodie/gibberish_generator/content.handlebars
GuiltyDolphin: @moollaza Whoops! Thanks, I'll get that sorted and then we can *wrap* this up ;)

GuiltyDolphin:
@moollaza All done :smile:

GuiltyDolphin:
@duckduckgo/community-leaders Anyone want to give this a once over?

javathunderman:
@GuiltyDolphin LGTM :+1: Any further suggestions @moollaza?

moollaza:
Thanks! This LGTM :+1:

PR #1973, opened 2015-12-28
Periodic Table Updated - Fixes #1914 

### LINKED ISSUE #1914: Periodic Table: Provide Atomic Mass for queries containing the name or "periodic table"
Queries for atomic elements by name, e.g. "Helium", "Lathanum" should show the atomic mass instead of repeating the full name.

Suggested on Duck.co here: https://duck.co/ideas/idea/6804/show-all-information-in-periodic-table-goodie

----

Also it looks like "periodic table" isn't a trigger for this IA and I think it should be added. Queries like https://duckduckgo.com/?q=periodic+table+he&t=ffab should also show the element name and mass.

We may need to update the design to always allow for the atomic mass /cc @abeyang 

/cc @zblair 

------
IA Page:  http://duck.co/ia/view/periodic_table

amitdev:
Fixes #1914

Changes the display format to always show atomic number and mass, and change the presentation based on query. Some examples:

Query: Helium
Query: atomic number of Helium
he

Query: atomic mass of helium
he2

Query: chemical name of Ag
he3

-----
IA Page: https://duck.co/ia/view/periodic_table


daxtheduck:
Metadata post coming soon

moollaza:
@amitdev thanks for the PR! FYI, if you add `Fixes #1914` to your PR description (I went ahead and did it for you this time), GitHub will automatically close the issue when your PR is merged.

lib/DDG/Goodie/PeriodicTable.pm
```diff
@@ -46,19 +47,17 @@ handle query_lc => sub {
     my $match = first { lc $_->[2] eq $query || lc $_->[3] eq $query } @elements or return;
     my ( $atomic_number, $atomic_mass, $element_name, $element_symbol, $element_type ) = @{$match};
 
-    # Default to displaying chemical symbol info.
-    my $title = $element_name;
-    my $subtitle = "Chemical Element";
-    my $raw = "$element_symbol, chemical symbol for " . lc($element_name);
+    # Default display
+    my $title = "$atomic_number";
+    my $subtitle = "$element_name (atomic mass: $atomic_mass u)";
```
moollaza: instead of parenthesis, let's return an array. This will force the subtitle to add a small vertical separator between the strings in the array. The result should look like:

> **2**
> Helium - Atomic Number | 4.0026 u - Atomic Mass

moollaza: We also have the option of using the `altSubtitle` which would be a separate line. I think this works for now.

@abeyang @thm any thoughts?

The goal is to always represent the atomic number and mass, right now we reiterate the name and don't provide the other useful information

amitdev:
Using altSubtitle also looks pretty decent (but I don't like the text : "atomic number" or "atomic mass" along with it). So something like the following - based on the query one of the rows will become the title (like atomic number).

he

thm:
@amitdev @moollaza thanks for the contribution! Any particular reason why you don't like the "atomic number" and "atomic mass" labels? I think they're actually really useful. These number can be pretty obscure for a lot of users. Obvious always wins! :)

alt-screenshot

amitdev:
@thm, No particular reason, just aesthetics I guess :grinning: But I agree that explicit is better than implicit. It looks like the following now (with other combinations as usual)

he

duckduckhacker:
Its great that this is being updated (I was the original poster on duck.co). This will be really useful for me when it leaves beta.

Would it be possible to get information on which block of the periodic table the element is in (e.g. "transition metal" or "rare earth element")? It appears that this information is already known by the IA as the colour of the square seems to change depending on this information. It would be great to get this in the text. This might be next to the element name, perhaps.

Any thoughts?

amitdev:
It can be added, though I'm not sure it should be shown in all cases. @thm, @moollaza, any thoughts?

duckduckhacker:
Would it be possible to have it appear if you over over the coloured box perhaps? And then just appearing if there is room. It would be useful information, particularly for the less well known elements.

moollaza:
>Would it be possible to have it appear if you over over the coloured box perhaps? And then just appearing if there is room. It would be useful information, particularly for the less well known elements.

@duckduckhacker thanks for the idea - typically we avoid any kind of hover interaction as it doesn't migrate well to mobile.

I'm speaking with our design team internally to determine the best way to always show the important information :+1:

duckduckhacker:
@moollaza I had not considered the mobile interface - you make a very good point. It would be interesting to see what you come up with for this improvement. Thanks

moollaza:
@amitdev @duckduckhacker the design team has a new design:

proposed-dark

The major difference is that the header would now be static, and should display the element name.

Below the title/subtile we'll have a list highlighting the atomic mass, number and the type of the element.

If the query specifically asks for the mass/number/type we can highlight that detail in the list by adjusting the font colour (using our handy classes!).

Note: the lower portion will need to be implemented as a `content.handlebars` sub-template, with a little logic to highlight text when necessary.

@amitdev would you be able to implement this?

amitdev:
@moollaza, I can give it a try. One question about the new design though - do we need the text "Chemical element" ? That seems redundant and maybe we can replace it with Type ("Noble Gas" in this case). So something like (added with proper highlighting):
```
Helium
Noble Gas
Atomic number: 2
Atomic mass: 4.0026 u
```

duckduckhacker:
@moollaza @amitdev 

I am very impressed about this new design and feel that it is a great improvement over the previous design. With regards to the phrase "chemical element" I guess it depends on how many people might accidentally find themselves looking at this and not quite understanding wha they are looking at.

thm:
@amitdev @duckduckhacker I think the "Chemical Element" line would make this IA a bit more clear. I'd rather give a complete answer, than leaving details out. However, I don't feel strongly about it. Feel free to make your own judgement on its usefulness. @moollaza

amitdev:
The current design looks good to me as well (the 'Chemical Element' part is a minor point anyway). @moollaza, we need to use handlebars template for the whole thing right? Or is it possible to use it only for subtitle/altsubtitle?

moollaza:
>@moollaza, we need to use handlebars template for the whole thing right? Or is it possible to use it only for subtitle/altsubtitle?

@amitdev nope, the title, subtitle, icon are all taken care of already by the `icon` template. You just need a `content.handlebars` template that builds the portion below with the 3 lines.

>http://docs.duckduckhack.com/frontend-reference/subtemplates.html

amitdev:
@moollaza, Check now.

he
he

btw, there seems to be some problem in the placement of the "i" icon of more info which seems to be there for all goodies I tested.

amitdev:
@moollaza, any comments?

moollaza:
@amitdev apologies for the delay! this is looking good :+1: 

Let me get it deployed to beta for further review and testing...

moollaza:
@amitdev this looks perfect. The only minor bug I noticed is that in the dark theme the lower text all has the same color, losing it's emphasis.

@abeyang @thm should text--primary have a different color in the dark theme? It looks like both `text--primary` and `text--secondary` have the same color/weight etc so there's no visual separation. Should `text--primary` be white perhaps?

moollaza:
I will merge for now, we may have to fix the dark-theme css internally

PR #1974, opened 2015-12-28
SigFigs: Update to use templates + other improvements

GuiltyDolphin:
Regards the [SigFigs](https://duck.co/ia/view/sig_figs) Goodie.

Does a few things:

 - Updates the Goodie to use the `text` template.
 - Allows input of numbers using different systems (via the `NumberStyler` GoodieRole).
 - Fixes a bug wherein the Goodie was not triggering if input consisted of leading zeros.
 - Improves triggers.
 - Improves handling of significant figures.

I must say that in researching for this, I've repeatedly come across the ambiguity of significant figures when it comes to trailing zeros on integers without the use of scientific notation.

Currently the Goodie just counts the number of non-zero digits in the above case.


daxtheduck:
Metadata post coming soon

moollaza:
Deploying to beta...

lib/DDG/Goodie/SigFigs.pm
```diff
@@ -14,38 +15,58 @@ topics 'math';
...
+    my $to_compute = $style->for_computation($number_match);
+    my $sigfigs = get_sig_figs $to_compute;
+    return unless defined $sigfigs;
+
+    return $sigfigs, structured_answer => {
+        id   => 'sig_figs',
+        name => 'Answer',
+        data => {
+            title    => "$sigfigs",
+            subtitle => "Significant figures of $formatted_input",
```
mintsoft: maybe `Number of significant figures in $formatted_input` ?

t/SigFigs.t
```diff
@@ -8,18 +8,46 @@ use DDG::Test::Goodie;
...
+    [qw( DDG::Goodie::SigFigs)],
+    'sf 78'                         => build_test('2', '78'),
+    "sigfigs 10.12404"              => build_test('7', '10.12404'),
+    "significant digits of 3030."   => build_test('4', '3,030.'),
+    "significant figures 0.0100235" => build_test('6', '0.0100235'),
+    "sf 302.056"                    => build_test('6', '302.056'),
+    "significant figs 045.30"       => build_test('4', '045.30'),
+    'sigfigs 01.1234000'            => build_test('8', '01.1234000'),
+    'significant figures 000123000' => build_test('3', '000,123,000'),
+    'sigfigs 0'                     => build_test('0', '0'),
```
mintsoft: I think technically this is undefined, i.e. 0 can't be rounded to significant figures however I'd imagine this is fine

lib/DDG/Goodie/SigFigs.pm
GuiltyDolphin: Sure, definitely sounds better. What about `Number of Significant Figures of X`?

t/SigFigs.t
GuiltyDolphin: Yeah, I agree - but considering there are the issues with not using scientific notation - it might pass.

It is but a simple check if anyone else wants it to undefine itself.

moollaza:

PR #1982, opened 2015-12-28
Issue #1163 - Convert Dice to use templates

Mailkov:
IA Page: https://duck.co/ia/view/dice


daxtheduck:
Metadata placeholder post
            
---
            
**How to find your IA page**
- New Instant Answer
              
- Existing Instant Answer

moollaza:
deploying to beta...

moollaza:
@Mailkov let's standardize this to use the `title` and `subtitle`.

In each case, the `title` should always be the total.

When rolling 6-sided die, the subtitle could be used to represent the rolls.

When using multiple die, we could either use a list view to show each die, or if theres only a few die rolled, and array can be given to the subtitle which would separate the rolls.

e.g. standard:
> **18**
> ⚀ ⚅ ⚄ ⚅

non-standard die:
> **19**
> 5 + 2 + 6 = 13 | 4 + 2 = 6

moollaza:
this way we can remove the CSS and custom templates.

mintsoft:
@Mailkov @moollaza Do you think the dice look a little small?

image

moollaza:
@Mailkov @mintsoft my bad, I didn't realize how small the subtitle would be. I think we can maintain the current look then for the standard query where we show the dice on the left and then ` = <total>` on the right as the `title`.

We could put a subtitle of `Random dice throw`

When there are multiple dice, we can do the total as the title, the breakdown as the alt-subtitle and the "Random Dice Throw" as the subtitle

mintsoft:
@moollaza Yeah that sounds good

Mailkov:
@moollaza 
We had spoken about this on Slack  "29 December".
I think the best solution is to increase font-size of subtitle, for example:
```
 zci--dice h4{
     font-size: 1.3em;
 } 
```

>When there are multiple dice, we can do the total as the title, the breakdown as the alt-subtitle and the >"Random Dice Throw" as the subtitle

I don't think this solve the dimension problem of dice

moollaza:
>I don't think this solves the dimension problem of dice

@Mailkov I think @marianosimone has a good solution, which is to use the `list` template when dealing with more than 1 Dice.

I think then the best solution is to use the dice icons for the title when dealing with a single die, and otherwise, show the title as the total, and list out the separate rolls in the list as done in #2043.

I'm not sure we should use dice images in the list though because I think it looks a little odd.

@abeyang do you have any thoughts on this?

moollaza:
@Mailkov @marianosimone apologies for the delay, the design team now has a design for us to use and move forward with. We'll use the standard dice characters for values < 7 and we'll use numbers, surrounded by a border (to look like the dice) for rolls of 7+.

Here are some examples:
dice-a-3
dice-b-3d100
dice-c-multiple

For the standard title/subtitle layout we should be able to use the `text` group with a `subtitle_content` template for the subtitle. We'll need some CSS to wrap the numbers in a border and to increase the font size for the dice.

For the case where we have 2+ dice we'll use the list view.

@Mailkov @marianosimone would one of you like to implement this?

Mailkov:
@moollaza I'll do it ...

moollaza:
:+1:

Mailkov:
@moollaza 
Done !

moollaza:

share/goodie/dice/dice.css
```diff
@@ -1,8 +1,17 @@
...
+.zci--dice .rolls{
+    width: 88%;
+    padding-left: 0px;
+}
+.zci--dice .nodice{
+    text-align: center;   
+    display: inline-block;
+    margin: 0px 4px 9px 0px;
+    min-width: 10px;
+    height: 10px;
```
moollaza: Let's do `height: 12px; line-height: 16px` and remove the `padding`. That looks a bit better

Mailkov: @moollaza
This is with my css

rolldicecss1

This is with your hint
dicecss2


IMHO  the first is better ... the numbers are better fixed in the center

Issue #1847, opened 2015-12-02
Conversions: floating point error (feet to inches)


mrkev:
@mintsoft

Hello! I stumbled across this issue.
![screen shot 2015-12-02 at 4 13 40 pm](https://cloud.githubusercontent.com/assets/1285131/11544451/e618d334-990f-11e5-8c2e-d131dbc44e3d.png)

Thought I'd report it here.

------
IA Page:  http://duck.co/ia/view/conversions

mintsoft:
Thanks for the report @mrkev ! We'll look it into

mrkev:
:+1: 😊

Mailkov:
@mintsoft @mrkev 

I think the problem is that the factor is based on  the meter on Convert::Pluggable 
I found it on https://github.com/elohmrow/Convert-Pluggable/blob/master/lib/Convert/Pluggable.pm
```
{
            'unit'      => 'meter',
            'factor'    => '1',
            'aliases'   => ['meters', 'metre', 'metres', 'm'],
            'type'      => 'length',
        },
        {
            'unit'      => 'kilometer',
            'factor'    => '0.001',
            'aliases'   => ['kilometers', 'kilometre', 'kilometres', 'km', 'kms', 'klick', 'klicks'],
            'type'      => 'length',
        },
        {
            'unit'      => 'centimeter',
            'factor'    => '100',
            'aliases'   => ['centimeters', 'centimetre', 'centimetres', 'cm', 'cms'],
            'type'      => 'length',
        },
        {
            'unit'      => 'millimeter',
            'factor'    => '1000',
            'aliases'   => ['millimeters', 'millimetre', 'millimetres', 'mm', 'mms'],
            'type'      => 'length',
        },
        {
            'unit'      => 'mile',
            'factor'    => 1/1609.344,
            'aliases'   => ['miles', 'statute mile', 'statute miles', 'land mile', 'land miles', 'mi'],
            'type'      => 'length',
        },
        {
            'unit'      => 'yard',
            'factor'    => '1.0936133',
            'aliases'   => ['yards', 'yd', 'yds', 'yrds'],
            'type'      => 'length',
        },
        {
            'unit'      => 'foot',
            'factor'    => '3.28084',
            'aliases'   => ['feet', 'ft', 'international foot', 'international feet', 'survey foot', 'survey feet'],
            'type'      => 'length',
        },
        {
            'unit'      => 'inch',
            'factor'    => '39.3701',
[...]

Issue #1754, opened 2015-11-11
Conversions: Cups not triggering on production


tagawa:
For example this query doesn't trigger a conversion:
https://duckduckgo.com/?q=4+cups+in+quarts

Although cups seem to be in the upstream code:
https://github.com/elohmrow/Convert-Pluggable/blob/33c0275e0b358ac841b4a10a3915dc18b1cb6307/lib/Convert/Pluggable.pm

mintsoft:
@tagawa I've just checked this and the goodie fires correctly, however it doesn't in production!

![image](https://cloud.githubusercontent.com/assets/978048/11192742/ad87bd7c-8c9a-11e5-95d1-568689a8b1e9.png)


@moollaza @jagtalon I can't debug this any further from here; it looks like a live-only problem!

moollaza:
@mintsoft I suspect our version of Convert::Pluggable is behind in production. Will make a task

Issue #1729, opened 2015-11-01
Negative degree conversion to Kelvins do not work


Bujiraso:
When converting from degree scales (e.g. `55C in Fahrenheit`), if negative numbers are used while converting **to** Kelvins, no instant answer appears to be triggered, and incorrect results are shown
![negativetokelvin](https://cloud.githubusercontent.com/assets/1975771/10869980/ab253796-8095-11e5-8847-1b8af350e75e.png)

Conversions **from** Kelvins work -- including no answer shown for negative Kelvin (non-sensical) input.

Conversions of positive degrees to Kelvins and conversions between Celsius and Fahrenheit all work well, even with negative numbers. 
![negativetof](https://cloud.githubusercontent.com/assets/1975771/10869998/19201edc-8096-11e5-8cc6-194453318cd0.png)

mintsoft:
@Bujiraso Thanks for the report, I've had a quick look into what's going on here and it looks like there's an issue within Convert::Pluggable that's causing this behaviour. I'll see if I can dig through it again and track it down later on

cc @elohmrow

elohmrow:
@mintsoft what version of C::P is currently in use? i recall a (possibly) similar issue recently https://github.com/duckduckgo/zeroclickinfo-goodies/issues/1413#issuecomment-141917103 ...

elohmrow:
@mintsoft it seems most problems with C::P come from the temperature conversions ... i think my strategy of first converting Y to F then from F to X has some logical hole(s) somewhere.  it may be better for now to simply re-write that routine in C::P using something like http://search.cpan.org/~mopy/Convert-Temperature-0.03/lib/Convert/Temperature.pm for now.  i will implement that w/n 24 hrs and release a new C::P if you think my assessment is more or less accurate.

mintsoft:
@elohmrow I think it's 0.030; I've just upgraded to that on my dev environment and issue still persists.

WRT to temperature conversion, there *shouldn't* be any technical problem with going via another unit. For example Kelvin is the SI unit of temperature and is an absolute measure of temperature (i.e. 0K is 'no kinetic energy of particles'), whilst Celsius & Fahrenheit are standardised around other things. So practically, it should be acceptable to always convert anything into Kelvin and then back to another unit, this is true of any other unit (Fahrenheit for example) as long as the appropriate accuracy is maintained throughout the conversion.

[...]

Issue #1413, opened 2015-08-18
Rankine temperature conversion is incorrect


ahstro:
As of now, `10 rankine in fahrenheit`  outputs `-441.670 °F`, when, according to this conversion, `[°F] = [°R] − 459.67`, found [here](https://en.wikipedia.org/wiki/Conversion_of_units_of_temperature), it should equal `-449.67`.
I searched through both [`Conversions.pm`](https://github.com/duckduckgo/zeroclickinfo-goodies/blob/master/lib/DDG/Goodie/Conversions.pm) and the imported [`Convert::Pluggable`](https://github.com/elohmrow/Convert-Pluggable), but just when I thought I found the right place in the latter, I realized that the [conversion process](https://github.com/elohmrow/Convert-Pluggable/blob/master/lib/Convert/Pluggable.pm#L46) seemed correct.

I'm baffled and Perl is not my primary, nor secondary, language, so I'm going to leave this issue here. Also, CC @elohmrow, since you created the conversion module, although I'm not sure that's where the problem lies.

elohmrow:
i will look in a bit - thanks!

elohmrow:
i've taken a peek - and it seems *very* odd.  F<=>R is actually the *easiest* of the conversions in that set, but it simply is not working.  tracing it out, running new tests, changing the code (it really needs updating, and i had started to, but there does not seem to be much call for it) - nothing.  i'll have to look again, or, perhaps someone else copied on here might see it?

seisfeld:
Dunno if that helps, but Celsius to Rakine also gives wrong answer (-263.150 °C but should give -267.5944444 °C).

Source to double check: https://converterin.com/temperature/rankine-scale-to-celsius-degrees.html

elohmrow:
yeah its just rankine ... basically the way I implemented it is fahrenheit is base, every unit turns first to fahrenheit, then gets converted.

so since R<=>F is not working, no rankine conversions will :(

/bda

seisfeld:
This looks like a rounding or precision error. The answer is exactly off by -0.8 F per 1 R:

DDG: 1 R -> -457.870 F
Correct: -458.67 F
Difference: -0.8

DDG: 10 R -> -441.670 F
Correct: -449.67 F
Difference: -8

elohmrow:
@seisfeld *very nice* catch!<br />
[...]

PR #1996, opened 2016-01-01
Conversions: "Convert The Future Into The Present"

mintsoft:
Hello Everyone.

We've been discussing for a while what we want to do around unit conversion. I've been evaluating a number of API's and npm modules as @moollaza knows. I've not found any which cover the same breadth of units that we do currently and that are easily extensible for us to add more later. I feel that given the profile of this IA, it's important for us to be able to customise, fix and maintain it quickly and easily.

I was in conversation with @GuiltyDolphin about what we can do for the future of both Calculator and Conversions. It seems like a natural progression would be to be able to handle queries such as `33 inches + 5 cm in yards`. Extending the grammar for the calculator in https://github.com/duckduckgo/zeroclickinfo-goodies/pull/1927 for units should be relatively trivial, especially if it is handed a list or collection of units from somewhere; however it means we need control over the units and conversions handling etc. 

I think this leads us down a path of modularising both Conversions and Calculator into a collection of Roles (or similar) so we can easily assemble what we need. 

We were also discussing about interactivity of these IA's; we have a couple of choices. One is to re-implement them entirely in JS (relatively trivial for Conversions) however much more difficult for Calculator as grammar parsing isn't very mature in JS. @GuiltyDolphin suggested that there was a conversation around being able to 50-50 it, so that we can build a front-end in JS and do some AJAX off to the backend for processing. This seems like an ideal compromise to me, but I don't believe the IA architecture currently supports such an arrangement.

I've assumed that either way we could do with centralising and tidying this up a bit so I've resurrected an old branch where I brought everything back together and tidied things out in order to kick start the conversation.

This branch:
* Fixes the temperature conversions (includes passing tests from: https://github.com/duckduckgo/zeroclickinfo-goodies/pull/1731)
* Thus fixes https://github.com/duckduckgo/zeroclickinfo-goodies/issues/1413
* Includes https://github.com/duckduckgo/zeroclickinfo-goodies/pull/1992
* Can be easily extended by adding units to the `ratios.yaml` : i.e. resolves: https://github.com/duckduckgo/zeroclickinfo-goodies/issues/828
* Adds support for queries like `How to convert meters to inches` as suggested by @GuiltyDolphin 
* Fixes https://github.com/duckduckgo/zeroclickinfo-goodies/issues/1754
* Fixes (by expanding the existing bodge, until ambiguous units can be resolved, I'd like to tackle that as a separate unit of work) https://github.com/duckduckgo/zeroclickinfo-goodies/issues/1767
* Fixes https://github.com/duckduckgo/zeroclickinfo-goodies/issues/1847 by increasing accuracy of definition rather than rounding the result off sooner
* Fixes https://github.com/duckduckgo/zeroclickinfo-goodies/issues/1729

I would like (if we agree to go down this path) to separate out prefixes from the rest (so all units can be expressed as kiloX or megaX) etc which would simplify things massively and help the modularity.

I'm keen to hear what people think about this, I think we've been a bit stuck-in-the-mud around conversions for too long now.

@duckduckgo/community-leaders @moollaza @jagtalon 

===============
https://duck.co/ia/view/conversions

### DIFF
--- a/lib/DDG/Goodie/Conversions.pm
+++ b/lib/DDG/Goodie/Conversions.pm
@@ -71,14 +71,14 @@ handle query_lc => sub {
     
     # hack - convert "oz" to "fl oz" if "ml" contained in query
     s/(oz|ounces)/fl oz/ if(/ml/ && not /fl oz/);
-warn("LINE 74");    
+#warn("LINE 74");    
     # guard the query from spurious matches
     return unless $_ =~ /$guard/;
-warn("LINE 77");    
+#warn("LINE 77");    
     my @matches = ($+{'left_unit'}, $+{'right_unit'});
     return if ("" ne $+{'left_num'} && "" ne $+{'right_num'});
     my $factor = $+{'left_num'};
-warn("LINE 81");
+#warn("LINE 81");
     # Compare factors of both units to ensure proper order when ambiguous
     # also, check the <connecting_word> of regex for possible user intentions 
     my @factor1 = (); # conversion factors, not left_num or right_num values
@@ -109,7 +109,7 @@ warn("LINE 81");
         }
     }
 
-warn("LINE 112");
+#warn("LINE 112");
     # if the query is in the format <unit> in <num> <unit> we need to flip
     # also if it's like "how many cm in metre"; the "1" is implicitly metre so also flip
     # But if the second unit is plural, assume we want the the implicit one on the first
@@ -127,7 +127,7 @@ warn("LINE 112");
         @matches = ($matches[1], $matches[0]);
     }
     $factor = 1 if ($factor =~ qr/^(a[n]?)?$/);
-warn("LINE 130");
+#warn("LINE 130");
     my $styler = number_style_for($factor);
     return unless $styler;
 
@@ -137,9 +137,9 @@ warn("LINE 130");
         'to_unit' => $matches[1],
     });
 
-warn("LINE 139");
+#warn("LINE 139");
     return if !$result->{'result'};
-warn("LINE 141");
+#warn("LINE 141");
     my $formatted_result = sprintf("%.${precision}f", $result->{'result'});
 
     # if $result = 1.00000 .. 000n, where n <> 0 then $result != 1 and throws off pluralization, so:
@@ -192,7 +192,7 @@ sub looks_plural {
 sub convert_temperatures {
     my ($from, $to, $in_temperature) = @_;
 
-    warn("'$from' '$to' '$in_temperature'\n");
+    #warn("'$from' '$to' '$in_temperature'\n");
 
     my $kelvin;
     # Convert to SI (Kelvin)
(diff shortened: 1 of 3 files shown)


daxtheduck:
Metadata placeholder post
            
---
            
**How to find your IA page**
- New Instant Answer
              
- Existing Instant Answer

GuiltyDolphin:
*subscribing again* @mintsoft Very fresh!

jagtalon:
@mintsoft oh my god! This is exciting!

jagtalon:
I'll be testing it out a bunch today. <3

javathunderman:
Exciting!

moollaza:
@mintsoft this looks really promising. I just realized that if the backend continues to return plaintext results, we could potentially leverage the DDG API to build interactivity into the frontend very easily.

Or if absolutely necessary we could likely build a service to handle conversions and return the structured answer as JSON.

I'll get this onto beta so we can give it a test!

mintsoft:
@moollaza Using the DDG API sounds like a good idea actually; at least for an initial version; it'd fit into what @GuiltyDolphin and I were discussing for future revisions

mintsoft:
@moollaza Any idea why the deploy's failed?

mintsoft:
@jagtalon @duckduckgo/community-leaders 

I think this is the most I'd like to do in this PR. I've already discussed with @moollaza about the js side of the things and I'm going to spike something out this weekend hopefully.

There are several issues still here that remain (ambiguous unit resolution is 'fixed' with the bodge around: https://github.com/duckduckgo/zeroclickinfo-goodies/pull/1996/files#diff-e40dae3e0dc357950d79ca5c2e964b3eR74) SI and Binary prefixes are added manually (so only some units support them etc). However I'd like to tackle each of these things (js, ambiguity, prefixes) as separate PRs if this approach is settled upon.

gautamkrishnar:
:heart:

mintsoft:
I've also just gone through all the issues and updated the description with the issues that are solved with this PR

lib/DDG/Goodie/Conversions.pm
```diff
@@ -78,26 +72,20 @@ handle query_lc => sub {
     $_ =~ s/ degrees (celsius|fahrenheit)/ $1/;
```
moollaza: looks like we don't support `1 degree X to Y` or `X degree(s) (kelvin|rankine) to Y`?

moollaza: Can fix in next PR -- the live version has this same behaviour anyway...

mintsoft: Fixed in: https://github.com/duckduckgo/zeroclickinfo-goodies/commit/567b651a3b0ca7d3e48284b115fa8732f9dbdc26

moollaza:
@mintsoft this LGTM! 

Let's get those Travis tests passing again!

PR #1998, opened 2016-01-02
Regexp: Improvements!

GuiltyDolphin:
Which of these do you prefer?

screenshot from 2016-01-02 16 08 13

**OR**

screenshot from 2016-01-02 16 10 16

Yeah, I thought so. The first one is clearly superior ;)

## To-do

- [x] Support Perl-style matching (`foo =~ /foo/`).
- [x] Improve ordering of results (full match first, numbered matches in order etc...).
- [ ] Ensure that the Goodie won't try and compute really complex regexes (maybe time limit or just detect complexity?). 
- [x] Support basic modifiers (ignore case, for example).
- [x] Expand key width a bit - named captures are wrapping too early.

### Bugs

- [x] Matches like `regexp /foo/ foo` will allocate `1` to the first numbered match.
- [x] Uppercases the named captures (i.e, `Named Capture <x>` becomes `Named Capture <X>`).

---
[IA Page](https://duck.co/ia/view/regexp)

### DIFF
--- a/lib/DDG/Goodie/Regexp.pm
+++ b/lib/DDG/Goodie/Regexp.pm
@@ -21,10 +21,18 @@ sub get_full_match {
     return substr(shift, $-[0], $+[0] - $-[0]);
 }
 
+# Ensures that the correct numbered matches are being produced.
+sub real_number_matches {
+    my ($one, @numbered) = @_;
+    # If the first match isn't defined then neither are the others!
+    return defined $one ? @numbered : ();
+}
+
 sub get_match_record {
     my ($regexp, $str) = @_;
     my $compiler = Safe->new->reval(q{ sub { qr/$_[0]/ } });
     my @numbered = $str =~ compile_re($regexp, $compiler) or return;
+    @numbered = real_number_matches($1, @numbered);
 		my $matches = {};
 		$matches->{'Full Match'} = get_full_match($str);
     foreach my $match (keys %+) {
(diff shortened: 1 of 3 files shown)


daxtheduck:
Metadata placeholder post
            
---
            
**How to find your IA page**
- New Instant Answer
              
- Existing Instant Answer

jagtalon:
@GuiltyDolphin clearly the first one is best. :)

I'm excited! Ping us when you're ready.

GuiltyDolphin:
@jagtalon Ping!

This looks pretty good! I think that last point is fairly moot - I haven't managed to get it to even blink with complicated regexps!

moollaza:
deploying to beta...

lib/DDG/Goodie/Regexp.pm
```diff
@@ -9,26 +10,88 @@ use Safe;
...
 
-    return join( ' | ', @results ), heading => 'Regexp Result' if @results;
-    return;
+    return $matches,
+        structured_answer => {
+            id   => 'regexp',
+            name => 'Answer',
+            data => {
+                title       => "Regular Expression Match",
+                subtitle    => "Match regular expression /$regexp/$modifiers on $str",
```
moollaza: @GuiltyDolphin this is looking perfect, only bug is that we need to escape the input

`regex /(.+)/ <script>alert()</script>` breaks this :+1:

moollaza:

lib/DDG/Goodie/Regexp.pm
GuiltyDolphin: @moollaza I'm not getting that - are you sure that's not the old version?

PR #2026, opened 2016-01-09
Italic transliteration text for language template

wongalvis:
image

### DIFF
--- a/share/goodie/cheat_sheets/language.handlebars
+++ b/share/goodie/cheat_sheets/language.handlebars
@@ -4,7 +4,7 @@
       <div class="cheatsheet__key">{{key}}</div>
       <div class="cheatsheet__value text--secondary">{{val}}</div>
       {{#if trn}}
-        <div class="cheatsheet__translit text--secondary">{{trn}}</div>
+        <div class="cheatsheet__translit text--secondary"><em>{{trn}}</em></div>
       {{/if}}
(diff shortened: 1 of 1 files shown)


daxtheduck:
Metadata placeholder post
            
---
            
**How to find your IA page**
- New Instant Answer
              
- Existing Instant Answer

gautamkrishnar:
:+1: LGTM

moollaza:
@wongalvis we tend to use font-styling pretty lightly (usually opting for colour differentiation), but I can see the need to separate these visually.

@abeyang what are your thoughts on this? We have a new cheat sheet template that add's a 3rd line of text, providing a transliteration. How should we differentiate this from the actual translation? This format is only used when the translation is provided in a non-latin character set.

abeyang:
Hey @wongalvis , like @moollaza mentioned, we generally don't use italics -- but given the use case, I think this is fine. Thanks.

moollaza:
@abeyang thanks!

@wongalvis I think instead of CSS, we should use markup to indicate the text should be styled differently. Let's wrap the text inside the `<div>` with an `<em>` -- our site CSS will italicize the text as a result.

wongalvis:
@moollaza It's now updated.

moollaza:
@wongalvis Thanks! This LGTM :+1:

PR #2036, opened 2016-01-11
A javascript-focused template group

zachthompson:
Add a new template group with a javascript file in the configuration.  Also added a semi-modified Perl module template that puts the query remainder directly in the data, i.e. the processing will be done in the js.

See if this makes sense or if we just want to have people roll through the optional templates and select js to do the same thing.

### DIFF
--- a/template/lib/DDG/Goodie/Minimal.pm
+++ b/template/lib/DDG/Goodie/Minimal.pm
@@ -10,10 +10,10 @@ use strict;
 zci answer_type => '<: $ia_id :>';
 
 # Caching - https://duck.co/duckduckhack/spice_advanced_backend#caching-api-responses
-zci is_cached   => 1;
+zci is_cached => 1;
 
 # Triggers - https://duck.co/duckduckhack/goodie_triggers
-triggers any => 'triggerWord', 'trigger phrase';
+triggers any => 'triggerword', 'trigger phrase';
 
 # Handle statement
 handle remainder => sub {
--- a/template/lib/DDG/Goodie/Minimal.pm
+++ b/template/lib/DDG/Goodie/Minimal.pm
@@ -1,0 +1,46 @@
+package DDG::Goodie::<: $ia_package_name :>;
+
+# ABSTRACT: Write an abstract here
+# Start at https://duck.co/duckduckhack/goodie_overview if you are new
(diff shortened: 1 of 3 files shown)


daxtheduck:
Metadata placeholder post
            
---
            
**How to find your IA page**
- New Instant Answer
              
- Existing Instant Answer

template/lib/DDG/Goodie/Minimal.pm
```diff
@@ -0,0 +1,46 @@
...
+            # Name - Used for Answer Bar Tab
+            # Value should be chosen from existing Instant Answer topics
+            # see https://duck.co/duckduckhack/display_reference#codenamecode-emstringem-required
+            name => 'Answer',
+
+            data => {
+                remainder => $_
+            },
+
+            templates => {
```
moollaza: is this required? might as well define this once in the JS?

template/lib/DDG/Goodie/Minimal.pm
```diff
@@ -0,0 +1,46 @@
...
+use DDG::Goodie;
+use strict;
+
+zci answer_type => '<: $ia_id :>';
+
+# Caching - https://duck.co/duckduckhack/spice_advanced_backend#caching-api-responses
+zci is_cached   => 1;
+
+# Triggers - https://duck.co/duckduckhack/goodie_triggers
+triggers any => 'triggerWord', 'trigger phrase';
```
moollaza: just realized we have an uppercase letter here :frowning: 

That's my bad -- we should change that to `triggerword`

template/lib/DDG/Goodie/Minimal.pm
moollaza: Seems like it's required...

moollaza:
@zachthompson this LGTM. It would be great if the Perl didn't have to pass along a `template` obj so that the JS would handle *everything*.

Otherwise this works great.

zachthompson:
Totally agree about having to define the template here.  Had the same thought.  Unfortunately that's how we currently determine whether it's templated or not so we can treat it differently.  With autotemplating maybe going away completely, we should revisit and see if we can change this in the future.

moollaza:
:+1:

PR #2045, opened 2016-01-12
Binary to unicode converter

zachthompson:

Converts binary into unicode characters much like the [Ascii to Binary IA](https://duck.co/ia/view/ascii).  However, it has some additional features:

1. Converts to unicode outside of the ascii range.
2. Will convert a list of binary strings separated by spaces.
3. Retains the original query's spacing.

Every now and again you have some funny person that encodes a sentence in binary and you want to quickly decode it.  Here's your IA.  I don't want to decode it letter by letter.

Geeks and the curious.

https://duck.co/ia/view/ascii  We could potentially decommission ascii.

---
IA Page: https://duck.co/ia/view/bin2unicode

### DIFF
--- a/lib/DDG/Goodie/Bin2Unicode.pm
+++ b/lib/DDG/Goodie/Bin2Unicode.pm
@@ -4,24 +4,47 @@ package DDG::Goodie::Bin2Unicode;
 use DDG::Goodie;
 use strict;
 
-no warnings 'non_unicode';
+no warnings qw'non_unicode overflow portable';
 
 zci answer_type => 'bin2unicode';
 zci is_cached   => 1;
 
-triggers query => qr{^([01\s]{8,})(?:(?:to\s+)?(?:unicode|text|ascii))?$};
+triggers query => qr{^([01\s]{8,})(?:\s+(?:to\s+)?(?:unicode|text|ascii))?$};
+
+my $MAX_CODE_PT = 1114111;
 
 handle matches => sub {
+    my $q = $_; # orginal query
+    my $bin_string = shift @_; # captured binary string
 
-    my $bin = my $unicode = shift;
-    $unicode =~ s/([01]+)/chr(oct("0b$1"))/ge;
-    return "Binary '$bin' converted to unicode is '$unicode'",,
+    my $want_ascii = $q =~ /\bascii\b/;
+
+    my @bins = $bin_string =~ /([01]+|\s+)/g;
+    my $str;
+    for my $b (@bins){
+        if($b =~ /^[01]+$/){
+		    # Overflow/non-portable warnings expected
+            my $i = oct("0b$b");
+            # Assume ascii if out of range or explicitly requested.
+            # This will work for characters all in the same string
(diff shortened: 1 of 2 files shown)


daxtheduck:
Metadata placeholder post
            
---
            
**How to find your IA page**
- New Instant Answer
              
- Existing Instant Answer

lib/DDG/Goodie/Bin2Unicode.pm
```diff
@@ -0,0 +1,36 @@
...
+
+use DDG::Goodie;
+use strict;
+
+no warnings 'non_unicode';
+
+zci answer_type => 'bin2unicode';
+zci is_cached   => 1;
+
+triggers query => qr{[ 01]+(?:(?:to\s+)?unicode|text|ascii)?};
```
mintsoft: couldn't this be `triggers end => qw(unicode text ascii)`

zachthompson: @mintsoft It could but those are all optional.  So "01001101 01110101 01100011 01101000" as well as "01001101 01110101 01100011 01101000 to text" will work.  It's an open question I guess if we want the former to trigger.  I think we do ;)

mintsoft: That's fair enough I can certainly see the use-case of the pure-binary triggering

moollaza: @zachthompson I think we'll want to anchor this regex? `201 to text` triggers this.

It may also help to add a capture to the regex and `handle matches` below?

zachthompson:
Some fairly major changes:

* Cover some edge cases, e.g. erroneously trying to parse hex
* Default to ascii when exceeding the max code point.  This allows for long strings of multiple characters (no spaces) like the original.
* Tests

zachthompson:
BTW, this should now cover all of the space that Ascii did.  So if this seems like a more robust route, we might as well remove it as part of the PR.

moollaza:
@zachthompson this is looking good but I think we need to make sure the rendering unicode is visible? and exists:

1111_1111_to_unicode_at_duckduckgo_and_any_do
0000000000_to_unicode_at_duckduckgo_and_any_do

I'm just throwing 1's and 0's at it right now to see what happens...

t/Bin2Unicode.t
```diff
@@ -0,0 +1,65 @@
...
+use strict;
+use warnings;
+
+zci answer_type => 'bin2unicode';
+zci is_cached   => 1;
+
+sub gen_struc_ans {
+    my ($in, $bin, $str, $asc) = @_;
+
+    return "Binary '$bin' converted to " . $asc ? 'ascii' : 'unicode' . " is '$str'",,
```
moollaza: @zachthompson now that I see the double comma again `,,` is that intentional? (I just removed it in the PM, was pushing a commit to link the IA page and PR)

zachthompson:
Give it a shot now. Comma was a typo.

moollaza:
@zachthompson Thanks! This LGTM :+1:

Issue #853, opened 2014-12-17
DuckDuckGo Info IA has links of the wrong colour (like normal text). 


HackOrQuack:
Acording to the [Design Style Guide](https://duck.co/duckduckhack/design_styleguide):
#### Font Color
All text should be colored black, #000000, unless otherwise necessary. Coloring text grey, #808080 is also a good way to emphasize/differentiate text. However, no bolded text is allowed. We reserve bolding to highlight exact matches in our results. Furthermore, all links should be colored blue, #1168CE.
#### Rollover / Hover
Upon hovering, all links should have text-decoration: underline and cursor: pointer.

But at the moment links from the DuckDuckGo.pm IA aren't the right colour.
![DuckDuckGo Info](https://cloud.githubusercontent.com/assets/8239935/5478299/17ec2152-862b-11e4-9163-85b02a6e2dad.png)

https://duck.co/ia/view/duck_duck_go

altern8tif:
@moollaza @sdougbrown Is this still outstanding?  Not sure whether the link color should be blue or black since the style guide has changed.

HackOrQuack:
Though I can no longer see mention of link colour in the [general style guide](https://duckduckgo.com/styleguide) or the [IA style guide](https://duck.co/duckduckhack/design_styleguide), I can see links in both. The colour of links varies (always blue) but it is never the same as normal text. This is intuitive and helps users to see where they can click. This isn't the case with this IA so I think this is still a problem.

moollaza:
Yes this is still an issue. We need to fix this internally though as it affects more than just this IA. 

@abeyang, I assume these links should be grey?

bsstoner:
@abeyang or @chrismorast do you guys have any guidance here?

bsstoner:
Maybe we need a different UI for these types of IA's where it's pretty much just navigational.

abeyang:
@bsstoner I've just pinged you internally re: this.

tagawa:
For low-vision users and users with touch devices, having a permanent underline would be good IMO to indicate text is a link. A similar alternative is possible but underlines are easily understood by any level of user.

![ddg_tor_link](https://cloud.githubusercontent.com/assets/94173/11053894/a377a2f8-87a8-11e5-9deb-b2ded0c87356.png)

tagawa:
[...]

PR #2053, opened 2016-01-14
Update structure + format of DDG goodie

bsstoner:
To address #853

I restructured things according to @abeyang's mockup and got the first one ("about") working.

We need to finish going through all the responses and rewrite the text so it makes sense in the new layout. I could use some help if anyone's up for it: https://github.com/duckduckgo/zeroclickinfo-goodies/blob/892f9ea0c0cacfbff6faccbb9a1089a24ce8b178/share/goodie/duck_duck_go/responses.yml

After that, just need to update the tests!

@jagtalon @moollaza @chrismorast

### DIFF
--- a/lib/DDG/Goodie/DuckDuckGo.pm
+++ b/lib/DDG/Goodie/DuckDuckGo.pm
@@ -20,9 +20,17 @@ foreach my $keyword (keys %$responses) {
     my $response = $responses->{$keyword};
 
     if ($response->{title}) {
-        $response->{text} = $response->{title}.' '.$response->{url};
+        $response->{text} = $response->{title};
+        if ($response->{url}) {
+            $response->{text} .= ' '.$response->{url};
+        }
     } else {
         $response->{text} = '';
+    }
+
+    # if there's no subtitle, make the subtitle the url text:
+    unless ($response->{subtitle}) {
+        $response->{subtitle} = $response->{url};
     }
 
     if (ref($response->{aliases}) eq 'ARRAY') {
--- a/lib/DDG/Goodie/DuckDuckGo.pm
+++ b/lib/DDG/Goodie/DuckDuckGo.pm
@@ -59,6 +59,7 @@ handle remainder => sub {
         id => 'duck_duck_go',
         data => {
             title => $response->{title},
+            image => $response->{image},
             subtitle_text => $response->{subtitle},
             subtitle_url => $response->{url}
         },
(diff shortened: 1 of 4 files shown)


daxtheduck:
Metadata placeholder post
            
---
            
**How to find your IA page**
- New Instant Answer
              
- Existing Instant Answer

lib/DDG/Goodie/DuckDuckGo.pm
```diff
@@ -58,11 +45,22 @@ handle remainder => sub {
...
+        id => 'duck_duck_go',
+        data => {
+            title => $response->{title},
+            subtitle_text => $response->{subtitle},
+            subtitle_url => $response->{url}
+        },
+        templates => {
+            group => 'text',
+            options => {
+                subtitle_content => 'DDH.duck_duck_go.subtitle_content'
```
moollaza: We should be able to create the url if we pass in an object thanks to our `subtitle` helper:

```perl
subtitle => { url => $response->{url}, text => $response->{subtitle}, className => "tx--14" }
```

I need to confirm this still...

moollaza: Yup, it works :smile:

moollaza: Ah, but we can't add in the favicon this way...

moollaza:
I've done a first pass here: https://gist.github.com/moollaza/eefed2b30803b1d14529 -- open to feedback/suggestions

//cc @zekiel

bsstoner:
I merged @moollaza's gist and made some revisions. Some of these feel outdated and there's probably a bunch that are missing, i.e. "duckduckgo slack", "duckduckgo meetups", etc.. But even if it's not perfect, I think it's a big improvement over what's in prod currently.

@zekiel or others, feel free to jump in and add revisions: https://github.com/duckduckgo/zeroclickinfo-goodies/blob/brian/duckduckgo-goodie/share/goodie/duck_duck_go/responses.yml

lib/DDG/Goodie/DuckDuckGo.pm
bsstoner: yea, so annoying... trying to do this PR made me realize how difficult the templates are to work with.

moollaza: I'm both happy and sad to hear this :+1:

bsstoner:
:+1: Thanks @zekiel, just verified links with ids like that work.

Anyone else want to make revisions? IMO the fact that the links are clearer is big enough improvement to just merge and we can iterate on any other content changes.

@moollaza / @jagtalon want to do a final review and merge?

jagtalon:
@bsstoner checking it out!

jagtalon:
@bsstoner Ugh, sometimes an oxford comma is used, and sometimes it isn't. What are we, barbarians?!

I kid. It looks good! :+1: 

@moollaza What do you think about adding the edge browser?

PR #2060, opened 2016-01-14
DateMath: Improvements

### LINKED ISSUE #2056: DateMath.pm could/should handle time too?
I feel that queries such as `123456789 seconds ago` should work and be handled by probably `DateMath.pm`

I guess the location will be needed to do timezone adjustments etc

-----
https://duck.co/ia/view/date_math

GuiltyDolphin:
## Changes

A quick overview of what has been updated/changed.

### Updates to `DateMath`

1. Added support for hours, minutes, and seconds.

 This means you can now do queries such as `today + 7 hours`, and get, for example `16 Jan 2016 11:20:55 EST`.

2. Added support for using `time` as a triggering word.

 Instead of just being able to do `date today`, `date now` etc..., you can use `time` instead of `date`.

 When `time` is used, a 24hr clock is automatically displayed in output, regardless of the type of operation (adding days or hours etc...).

3. Provide support for specifying the time in input.

 Example: `01 Jan 2012 00:07:00 CET - 7 minutes`, yields `01 Jan 2012 00:00:00 CET`. The time-zone is optional and should hopefully be automatically detected.

### Changes to the `Dates` role

1. Added support for a new date format.

 It is essentially the `HTTP` format without the day, and supporting any TZ.

 An example of `HTTP` format might be: `Sun, 01 Jan 2012 00:00:00 UTC`. An example of the new format would be `01 Jan 2012 00:00:00 BST`.

 I think this format is more flexible - as well as being completely unambiguous.

2. Allow formatting of dates using the above format.

 I noticed that `Dates` only allowed output in terms of the day, month and year - `%d %b %Y`. I've added an optional flag (that does not affect any existing users of the Role) that tells `Dates` to output in terms of the above format `%d %b %Y %T %Z`. This allows full 24hr clock and time-zone (will try the three-letter variant, but fall back on offset) to be displayed.

3. Additional exports.

 Provide access to the `time_24h` and `time_12h` regular expressions - they are useful.

### Changes to `t/SunInfo`

1. Corrected test times.

 In testing, the main-used time-zone is `EST` (due to test location). The `SunInfo` tests were given in `UTC`, and due to my updates to the `Dates` role, the time-zone was being detected as being `EST`, so I've just corrected them to the 5hr difference.

 Please let me know if I've now made the tests incorrect - I did some testing and it seemed fine.

## To-Do

- [x] Convert to template.
- [x] Resolve #2056
- [x] Update to use NumberStyler
- [x] Support 24-hour clock notation (`01 Jan 2012 00:00:17 + 3 seconds`).

---
[IA Page](https://duck.co/ia/view/date_math)


daxtheduck:
Metadata placeholder post
            
---
            
**How to find your IA page**
- New Instant Answer
              
- Existing Instant Answer

GuiltyDolphin:
@mintsoft I've implemented the suggested functionality.

Queries such as `1 hour ago`, `3 seconds from now` etc... will all work.

Does not currently support specifying the time from which calculations should be made - I'll have a quick look into this.

t/00-roles.t
```diff
@@ -89,11 +89,17 @@ subtest 'Dates' => sub {
             # RFC850
             '08-Feb-94 14:15:29 GMT' => 760716929,
             # date(1) default
-            'Sun Sep  7 15:57:56 EDT 2014' => 1410119876,
```
mintsoft: Any particular reason the timezones have been changed here?

t/DateMath.t
```diff
@@ -7,138 +7,82 @@ use Test::MockTime qw( :all );
...
+
+ddg_goodie_test([ qw( DDG::Goodie::DateMath ) ],
+    # 2012 Jan tests
+    'Jan 1 2012 plus 32 days'       => build_test(@overjan),
+    'January 1 2012 plus 32 days'   => build_test(@overjan),
+    'January 1, 2012 plus 32 days'  => build_test(@overjan),
+    'January 1st 2012 plus 32 days' => build_test(@overjan),
+    '32 days from January 1st 2012' => build_test(@overjan),
+    # Relative (to today)
+    'date January 1st'         => build_test('01 Jan 2014', 'January 1st'),
```
mintsoft: This `date January 1st` query seems like a useless answer to me

t/SunInfo.t
```diff
@@ -5,10 +5,12 @@ use warnings;
 use Test::More;
 use DDG::Test::Goodie;
 use utf8;
+use Test::MockTime qw(set_fixed_time);
```
mintsoft: I'm guessing these updates shouldn't be in this PR?

t/00-roles.t
GuiltyDolphin: Ugh... Woops, thanks for noticing - I was updating some of the new tests and must've got that one accidently (works though ;))

t/SunInfo.t
GuiltyDolphin: It's because of my updates to the `Dates` role - it now actually performs some additional TZ checks, rather than just defaulting to `UTC` (or none). I tested out the `SunRise` package and the times were consistent with what I'd expect for UTC - but the tests should be in `EST`.

t/DateMath.t
GuiltyDolphin: Maybe - It gives the year though.

I might be able to silence queries that just add to a single date - but it might be a bit of a pain. I'll have a look.

GuiltyDolphin:
@mintsoft, @moollaza I've updated the PR comment to give a quick overview of the changes - hopefully this will better explain my reasoning.

t/DateMath.t
mintsoft: True, although outputting the year seems like something only someone working on the DatesRole would want to do in order to test it!

This is hardly a dealbreaker if it's awkward to silence it unless @duckduckgo/community-leaders disagree

GuiltyDolphin: No worries, the fix was actually pretty easy ^^

Basically will not accept queries where there is no action (such as adding days) **and** the specified date is not relative (such as `3 days ago`).

gautamkrishnar:
:+1:

moollaza:
@GuiltyDolphin @mintsoft this is looking pretty good so far. I'll deploy to Beta.

Once that's ready, would you guys like to test this with our nifty screenshot comparison tool?

https://www.npmjs.com/package/ddg-screen-diff

If you install this it enables you to take local screenshots and compare them with production, or compare two remote computed (i.e. beta and production).

If you can collect a list of the id's of the IA's that depend on this code we can very easily build a JSON file (a "group") and have this tool take screenshots to compare beta and production. It will highlight any differences between the results. Ideally we shouldn't see any in most cases ;)

You'd need to make a `date-goodies.json` file using the following format:

```json
[{
    "command": "ia",
    "commandValue": "abc"
}, {
    "command": "ia",
    "commandValue": "alt_calendars"
}, {
    "command": "ia",
    "commandValue": "anagram"
}]
```
and then once you've installed the tool, navigate to the dir containing your group.json file and run:
`ddg-screen-diff group "date-goodies" beta prod` --- this should take a bunch of screenshots and create an `index.html` file which you'll want to open. It'll show you rows of 3 photos, where the third is the visual-diff of the first two :)

GuiltyDolphin:
@moollaza Sweet - checking it out.

moollaza:
>Okay, I've done as you suggested and run it against all the Goodies that used the Dates role. There were no differences bar those for DateMath - which are expected.

Awesome! Sounds like everything went smoothly then?

>I did notice some issues with some of the IA's though - e.g., not triggering on both beta+production, triggering different IA etc.. (I'll have a further look into these).

Can you give me examples? Some discrepency is normal, especially if it’s a Fathead or Wiki that shows. They're a little faster to trigger, but if a certain IA should always show, we can set the `signal` property to `'high'` in the `meta` obj (for those that use a full structured answer)

```js
meta: {
    sourceName: "...",
    sourceUrl: "...".
    signal: "high"
}
```

GuiltyDolphin:
@moollaza The `SunInfo` Goodie has `sunrise` as its query, which is triggering the `Sunrise` Spice.

The screenshots for `NameDays` show no IA result - the `ia` parameter is being set to `namedays` (`&ia=namedays`) rather than `answer`. The `NameDays` Goodie does not use templates so this might be why?

And yes - as far as I can tell it all went smoothly!

andrey-p:
Hey @GuiltyDolphin, I saw the issue you reported yesterday in the `ddg-screen-diff` repo. That's really cool! I'll respond to it later today.

In the meantime, if you bump into anything else, feel free to make an issue for it. I initially built the tool for internal use, so there may be some, um, awkward things about using it locally.

GuiltyDolphin:
@andrey-p Will do - it'll be great to see it used around the place - will be very useful for developing IAs!

moollaza:
>The SunInfo Goodie has sunrise as its query, which is triggering the Sunrise Spice.

Crap! Thanks for pointing that out. We definitely adjust the Spice. It should only trigger when a place is specified.

>The screenshots for NameDays show no IA result - the ia parameter is being set to namedays (&ia=namedays) rather than answer. The NameDays Goodie does not use templates so this might be why?

This was an issue with the Metadata. I've changed the tab name in the metadata to "Answer". It should now open the tab correctly, once the metadata is live.

Thanks for this!

ScreapDK:
Off-topic: I think Dax the Duck is having a little trouble with the last couple query examples in his comment above, as he [made them appear as one query](https://beta.duckduckgo.com/?q=1/1/2012%20plus%205%20months,%20January%20first%20minus%20ten%20days,%20in%205%20weeks,%202%20weeks%20ago,%201%20month%20from%20today).

PR #2112, opened 2016-01-26
UltimateAnswer: Expand query space

tyleryasaka:
Thought it might be nice to expand the query space a bit. I didn't realize this goodie existed at first because I couldn't get it to trigger.

If you type in 'answer to' in DDG, the second suggestion is currently 'answer to life the universe and everything'.

And if you type this query into Google, it will output 42 as the answer. So I think it makes sense that the query should also trigger this goodie in DDG.

IA: https://duck.co/ia/view/ultimate_answer

### DIFF
--- a/lib/DDG/Goodie/UltimateAnswer.pm
+++ b/lib/DDG/Goodie/UltimateAnswer.pm
@@ -18,7 +18,7 @@ zci is_cached   => 1;
 
 handle remainder => sub {
     
-    return unless $_ =~ qr/^(what is)?( )?(the)?$/i;
+    return unless $_ =~ qr/^(what is )?(the)?$/i;
 
     my $answer = 'Forty-two';
 
--- a/lib/DDG/Goodie/UltimateAnswer.pm
+++ b/lib/DDG/Goodie/UltimateAnswer.pm
@@ -3,13 +3,38 @@ package DDG::Goodie::UltimateAnswer;
 
 use strict;
 use DDG::Goodie;
-triggers start => 'what is the ultimate answer', 'what is the ultimate answer to life the universe and everything', 'what is the answer to the ultimate question of life the universe and everything', 'answer to life the universe and everything';
+
+triggers any =>
+    'life the universe and everything',
+    'ultimate answer',
+    'answer to the ultimate question';
 
 zci answer_type => 'ultimate_answer';
 zci is_cached   => 1;
 
-handle remainder => sub {
-    return unless ($_ eq '' || $_ eq '?');
+handle query => sub {
+    
+    return if (
+        # 'life the universe and everything' is the name of a book in the series.
+        # Want to restrict this query a bit.
+        (
+            $_ =~ /life the universe and everything/ &&
+            !($_ =~ /question|answer|what is|meaning/)
+        ) ||
+        # 'ultimate answer' can refer to other topics (e.g. 'ultimate answer to kings')
+        # Also want to restrict this query.
+        (
+            $_ =~ /ultimate answer/ &&
+            !( $_ =~ /^(what is the )?ultimate answer(\?)?$/) &&
+            !( $_ =~ /life the universe and everything/ )
+        ) ||
+        # Restrict 'answer to the ultimate question' just to be safe
+        (
+            $_ =~ /answer to the ultimate question/ &&
+            !( $_ =~ /^((what is the )?(ultimate )?answer to the ultimate question(\?)?)$/) &&
+            !( $_ =~ /life the universe and everything/)
+        )
+    );
 
     my $answer = 'Forty-two';
 
--- a/lib/DDG/Goodie/UltimateAnswer.pm
+++ b/lib/DDG/Goodie/UltimateAnswer.pm
@@ -12,7 +12,7 @@ triggers any =>
 zci answer_type => 'ultimate_answer';
 zci is_cached   => 1;
 
-handle query => sub {
+handle query_lc => sub {
     
     return if (
         # 'life the universe and everything' is the name of a book in the series.
@@ -40,8 +40,8 @@ handle query => sub {
 
     return $answer,
         structured_answer => {
-            id        => 'ultimateanswer',
-            name      => 'Ultimate Answer',
+            id        => 'ultimate_answer',
+            name      => 'Answer',
             data      => {
                 title      => $answer,
                 subtitle   => 'The answer to the ultimate question of life, the universe and everything.'
--- a/lib/DDG/Goodie/UltimateAnswer.pm
+++ b/lib/DDG/Goodie/UltimateAnswer.pm
@@ -3,7 +3,7 @@ package DDG::Goodie::UltimateAnswer;
 
 use strict;
 use DDG::Goodie;
-triggers start => 'what is the ultimate answer', 'what is the ultimate answer to life the universe and everything', 'what is the answer to the ultimate question of life the universe and everything';
+triggers start => 'what is the ultimate answer', 'what is the ultimate answer to life the universe and everything', 'what is the answer to the ultimate question of life the universe and everything', 'answer to life the universe and everything';
 
 zci answer_type => 'ultimate_answer';
 zci is_cached   => 1;
--- a/lib/DDG/Goodie/UltimateAnswer.pm
+++ b/lib/DDG/Goodie/UltimateAnswer.pm
@@ -28,7 +28,7 @@ handle remainder => sub {
             name      => 'Answer',
             data      => {
                 title      => $answer,
-                subtitle   => 'The answer to the ultimate question of life, the universe and everything.'
+                subtitle   => 'The Answer to the Ultimate Question of Life, the Universe, and Everything.'
             },
             meta      => {
                 sourceName => 'Wikipedia',
--- a/lib/DDG/Goodie/UltimateAnswer.pm
+++ b/lib/DDG/Goodie/UltimateAnswer.pm
@@ -4,37 +4,21 @@ package DDG::Goodie::UltimateAnswer;
 use strict;
 use DDG::Goodie;
 
-triggers any =>
-    'life the universe and everything',
+triggers end =>
+    'meaning of life the universe and everything',
+    'answer to life the universe and everything',
+    'ultimate answer to life the universe and everything',
+    'answer to the question of life the universe and everything',
+    'answer to the ultimate question of life the universe and everything',
     'ultimate answer',
     'answer to the ultimate question';
 
 zci answer_type => 'ultimate_answer';
 zci is_cached   => 1;
 
-handle query_lc => sub {
+handle remainder => sub {
     
-    return if (
-        # 'life the universe and everything' is the name of a book in the series.
-        # Want to restrict this query a bit.
-        (
-            $_ =~ /life the universe and everything/ &&
-            !($_ =~ /question|answer|what is|meaning/)
-        ) ||
-        # 'ultimate answer' can refer to other topics (e.g. 'ultimate answer to kings')
-        # Also want to restrict this query.
-        (
-            $_ =~ /ultimate answer/ &&
-            !( $_ =~ /^(what is the )?ultimate answer(\?)?$/) &&
-            !( $_ =~ /life the universe and everything/ )
-        ) ||
-        # Restrict 'answer to the ultimate question' just to be safe
-        (
-            $_ =~ /answer to the ultimate question/ &&
-            !( $_ =~ /^((what is the )?(ultimate )?answer to the ultimate question(\?)?)$/) &&
-            !( $_ =~ /life the universe and everything/)
-        )
-    );
+    return unless $_ =~ qr/^(what is)?( )?(the)?$/i;
 
     my $answer = 'Forty-two';
 
--- a/lib/DDG/Goodie/UltimateAnswer.pm
+++ b/lib/DDG/Goodie/UltimateAnswer.pm
@@ -39,11 +39,21 @@ handle query => sub {
     my $answer = 'Forty-two';
 
     return $answer,
-      structured_answer => {
-        input     => [],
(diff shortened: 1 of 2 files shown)


daxtheduck:
@tyleryasaka Hey!
        
While the team and the community review your submission, we'd like to make sure you look good! If you'd like a free DuckDuckGo tshirt, please send over the following info to merch@duckduckgo.com, and be sure to put "Dev Merch" in the subject line:

Name:
Street Address:
City:
State / Province:
Country:
Phone (Non-US Orders only): 
T-shirt size:
GitHub username:
Link to this Pull Request:

Thanks!

daxtheduck:
Metadata placeholder post
            
---
            
**How to find your IA page**
- New Instant Answer
              
- Existing Instant Answer

GuiltyDolphin:
@tyleryasaka I completely agree - `What is the meaning of life, the Universe, and everything?` and such like should trigger this Goodie.

I'll have a look over this.

lib/DDG/Goodie/UltimateAnswer.pm
```diff
@@ -3,7 +3,7 @@ package DDG::Goodie::UltimateAnswer;
 
 use strict;
 use DDG::Goodie;
-triggers start => 'what is the ultimate answer', 'what is the ultimate answer to life the universe and everything', 'what is the answer to the ultimate question of life the universe and everything';
+triggers start => 'what is the ultimate answer', 'what is the ultimate answer to life the universe and everything', 'what is the answer to the ultimate question of life the universe and everything', 'answer to life the universe and everything';
```
GuiltyDolphin: @tyleryasaka Hmm... I think it needs to trigger with things like `what is the meaning of life, the universe, and everything`, `meaning of life...` etc...

You could perhaps have an `any` trigger for `what is the meaning of life` which would trigger both of those.

Also, pull these out into a list so it is easier to read.

t/UltimateAnswer.t
```diff
@@ -20,6 +20,14 @@ ddg_goodie_test(
...
         }
     ),
+    'answer to life the universe and everything' => test_zci(
+        'Forty-two',
+        structured_answer => {
+            input     => [],
+            operation => 'The answer to the ultimate question of life, the universe and everything',
+            result    => 'Forty-two',
+        }
+    ),
```
GuiltyDolphin: @tyleryasaka After you've done as mentioned in my other comment, could you add a few more tests for relevant queries?

Would like to see at least one for each of the triggers.

tyleryasaka:
@GuiltyDolphin Ok, I'll work on your suggestions.

tyleryasaka:
@GuiltyDolphin you don't think "what is the meaning of life" is too general? Right now that query results in an IA for a very thorough and deep wikipedia entry for "meaning of life". I wonder if some people might be bothered for such a serious question to return such a silly result. As opposed to the more specific question "what is the meaning of life the universe and everything", which would almost always be an intentional reference to Hitchhiker's Guide, in which case we could expect the user to be pleasantly surprised.
I'm fine with whatever, those are just some thoughts I have.

GuiltyDolphin:
@tyleryasaka I meant use `meaning of life` as a trigger that would catch those terms, you can then use a regular expression in the `handler` to make sure the query is valid, returning nothing if it isn't.

The alternative is to verbosely list out all of the triggers and then use those directly.

tyleryasaka:
Ok. Would it not make more sense to do an `any` on `life the universe and everything`, and then restrict further in the handler? Also if we went this route it would mess up some of the current queries it matches, like `what is the ultimate answer`. Currently the handler restricts the results to exact matches of the triggers. We could add some more triggers and then have a more complex handler to accommodate the previous query matches...
It would be nice to hear from one of the previous contributors of this goodie to make sure I understand the reasoning behind setting it up the way it is now.

GuiltyDolphin:
@tyleryasaka Having `start` with `what is` makes sense as you likely wouldn't want the `what is` to come later.

Using the `any` triggers `ultimate answer` and `life the universe and everything` would likely have you covered, you'd only need to do a minimal amount of guarding (for `meaning of...`, optionally `what is` with `meaning of...` etc...).

tyleryasaka:
@GuiltyDolphin the GoogieRole looks neat. If you could use another contributor I would be down to pitch in (assuming I could find the time).

So if I understand correctly, you are saying I could catch `ultimate answer` and `life the universe and everything` with `any` triggers and then restrict the results to those containing `meaning of` or `what is` and possibly a few other things. Something like that? I'll play around with it.

GuiltyDolphin:
@tyleryasaka Mind updating this to use [templates](http://docs.duckduckhack.com/frontend-reference/templates-overview.html)? (Most likely just want `text`).

tyleryasaka:
@GuiltyDolphin Ok I'll just hold off on the GoodieRole for now. If I have time I may try to give you some feedback.

Added a text template. See what you think.

lib/DDG/Goodie/UltimateAnswer.pm
```diff
@@ -3,22 +3,57 @@ package DDG::Goodie::UltimateAnswer;
...
     my $answer = 'Forty-two';
 
     return $answer,
-      structured_answer => {
-        input     => [],
-        operation => 'The answer to the ultimate question of life, the universe and everything',
-        result    => $answer
-      };
+        structured_answer => {
+            id        => 'ultimateanswer',
```
GuiltyDolphin: The `id` should match the IA page (`ultimate_answer`).

lib/DDG/Goodie/UltimateAnswer.pm
```diff
@@ -3,22 +3,57 @@ package DDG::Goodie::UltimateAnswer;
...
+            name      => 'Ultimate Answer',
```
GuiltyDolphin: Use `Answer`.

lib/DDG/Goodie/UltimateAnswer.pm
```diff
@@ -3,22 +3,57 @@ package DDG::Goodie::UltimateAnswer;
...
+    'life the universe and everything',
+    'ultimate answer',
+    'answer to the ultimate question';
 
 zci answer_type => 'ultimate_answer';
 zci is_cached   => 1;
 
-handle remainder => sub {
-    return unless ($_ eq '' || $_ eq '?');
+handle query => sub {
```
moollaza: @tyleryasaka You'll want to use `handle query_lc` here so queries with uppercase letters also work :+1:

moollaza: or, if you use `triggers end` you could `handle remainder` and validate that the remainder matches `what is the (ultimate)? (answer to)?`

I think that approach might be a little more straight forward?

--

An alternative is to just generate a list of verbose triggers and handle things like the IA currently does.

Happy either way.

BTW, I figured I should point out we also have a very similar result from our Constants Goodie: https://github.com/duckduckgo/zeroclickinfo-goodies/blob/master/share/goodie/constants/constants.yml#L226

/cc @GuiltyDolphin

GuiltyDolphin: @moollaza That one is triggered by `42` so shouldn't cause any issues (essentially the inverse of this).

tyleryasaka: @moollaza I am experimenting with either using `triggers end` and `handle remainder` or just using a verbose list of triggers. It is tricky because the only ways I can think of implementing it that way off the top of my head would limit the query space from what it is now. But maybe I am trying to make the answer too intelligent.

lib/DDG/Goodie/UltimateAnswer.pm
```diff
@@ -3,22 +3,41 @@ package DDG::Goodie::UltimateAnswer;
...
+    'ultimate answer',
+    'answer to the ultimate question';
 
 zci answer_type => 'ultimate_answer';
 zci is_cached   => 1;
 
 handle remainder => sub {
-    return unless ($_ eq '' || $_ eq '?');
+    
+    return unless $_ =~ qr/^(what is)?( )?(the)?$/i;
```
GuiltyDolphin: Could use `(what is )?(the)?`, unless I'm missing something.

GuiltyDolphin:
@moollaza Lets get this on Beta.

lib/DDG/Goodie/UltimateAnswer.pm
```diff
@@ -3,22 +3,41 @@ package DDG::Goodie::UltimateAnswer;
...
-triggers start => 'what is the ultimate answer', 'what is the ultimate answer to life the universe and everything', 'what is the answer to the ultimate question of life the universe and everything';
+
+triggers end =>
+    'meaning of life the universe and everything',
+    'answer to life the universe and everything',
+    'ultimate answer to life the universe and everything',
+    'answer to the question of life the universe and everything',
+    'answer to the ultimate question of life the universe and everything',
+    'ultimate answer',
+    'answer to the ultimate question';
```
GuiltyDolphin: We could move these into a JSON file, then slurp that up.

lib/DDG/Goodie/UltimateAnswer.pm
```diff
@@ -3,22 +3,41 @@ package DDG::Goodie::UltimateAnswer;
...
-        input     => [],
-        operation => 'The answer to the ultimate question of life, the universe and everything',
-        result    => $answer
-      };
+        structured_answer => {
+            id        => 'ultimate_answer',
+            name      => 'Answer',
+            data      => {
+                title      => $answer,
+                subtitle   => 'The answer to the ultimate question of life, the universe and everything.'
```
GuiltyDolphin: Use `The Answer to the Ultimate Question of Life, the Universe, and Everything`.

t/UltimateAnswer.t
```diff
@@ -9,17 +9,93 @@ use DDG::Test::Goodie;
...
+        structured_answer => $structuredAnswer
+    ),
+    'what is the answer to the ultimate question' => test_zci(
+        $answer,
+        structured_answer => $structuredAnswer
+    ),
+    'what is the answer to the ultimate question?' => test_zci(
+        $answer,
+        structured_answer => $structuredAnswer
+    ),
```
GuiltyDolphin: Add a tests for `Answer to the Ultimate Question of Life, the Universe, and Everything` and `The Answer to the Ultimate Question of Life, the Universe, and Everything`.

lib/DDG/Goodie/UltimateAnswer.pm
tyleryasaka: What would be the benefit of that?

GuiltyDolphin: Hmm... We can probably leave it for now, although you should probably move it out into a separate list.

If any more triggers are added though, it'll be worth moving it just to keep the file clean.

tyleryasaka: @GuiltyDolphin Ok I'll go ahead and move it out. Not quite sure where to put the file though. Is this in the documentation somewhere, or is there another goodie that does this?

GuiltyDolphin: Additional files go in the `share` directory, under the name of the instant answer - in this case it'd be `share/goodie/ultimate_answer/FILE`, you can then use `share('FILE')->slurp()` in the Perl file to get the contents.

It might be worth just using a plain text file for this - now I think about it; you don't really need to do more than just read the contents into a list.

GuiltyDolphin:
@moollaza Can we get this on Beta?

GuiltyDolphin:
@tyleryasaka How's this going?

Looking good so far - I'll take a look at it on Beta (when it has bee deployed) and let you know if I spot any regressions.

/cc @moollaza

moollaza:
Deploying to beta...

lib/DDG/Goodie/UltimateAnswer.pm
```diff
@@ -3,22 +3,41 @@ package DDG::Goodie::UltimateAnswer;
...
 use strict;
 use DDG::Goodie;
-triggers start => 'what is the ultimate answer', 'what is the ultimate answer to life the universe and everything', 'what is the answer to the ultimate question of life the universe and everything';
+
+triggers end =>
+    'meaning of life the universe and everything',
+    'answer to life the universe and everything',
+    'ultimate answer to life the universe and everything',
+    'answer to the question of life the universe and everything',
+    'answer to the ultimate question of life the universe and everything',
```
GuiltyDolphin: Humm, didn't trigger on https://beta.duckduckgo.com/?q=answer+to+the+ultimate+question+of+life%2C+the+universe%2C+and+everything

Hmm.. I tried a few queries out and it didn't like them unless I prefixed them with `what is`, perhaps some regex triggers need to be sorted?

tyleryasaka: @GuiltyDolphin that query passes the test case. I think maybe the problem is that wikipedia answer gets "preference"? Idk how DDG handles all that.

In any case that query returns pretty much the same answer, just with a link to wikipedia. IMHO that's ok.

Which queries did you try?

GuiltyDolphin: @tyleryasaka Seems to be working fine now - not sure why it wasn't doing it before.

moollaza:
@tyleryasaka thanks again for all the hard work you've put into improving this Ultimate Answer Goodie. You've done a great job so far! We at DuckDuckGo, and the DuckDuckHack community really appreciate you taking the time to contribute.

It looks like this is just about finished?  If you're stuck or need any help, I'd love to chat if you're around this week or next. Please message me on slack (@moollaza) or send me an email (moollaza@duckduckgo.com) if you'd be interested in talking. 

Thanks again!

tyleryasaka:
@moollaza As far as I know it's ready to go. Is there anything else we need to do @GuiltyDolphin ?

GuiltyDolphin:
@tyleryasaka @moollaza Yeah, this looks good to me, thanks (sorry for the delay). :dolphin:

Issue #1905, opened 2015-12-15
'list' as triggering  word for cheat sheet.


jophab:
Shall we add the word 'list ' as  triggering word for cheat sheet.
It will be helpful in case of IAs which list something.
for eg : #1884 

---

IA Page: https://duck.co/ia/view/cheat_sheets

jophab:
@moollaza  @jdorweiler  @tagawa

moollaza:
@jophab thanks for the suggestion. I'd like to hold off on adding more triggers to cheat sheets for now. We'll be revisiting our approach to triggering cheat sheets in the new year to ensure they're triggering as well as possible.

gautamkrishnar:
@jophab that could be a good idea.

gautamkrishnar:
@moollaza I think, We must add some more trigger words like:-
* help
* quickstart
* jumpstart

vCillusion:
I would like to add one more thing can we include some additional metadata for categorization. See the following case.
Example: Trigger for language lets say 'Javascript' 
Query : javascript syntax -> should display the cheatsheet

tagawa:
Perhaps this can be closed, thanks to #2170 ?

GuiltyDolphin:
@tagawa I've left this one open for now as I haven't actually added a `list` trigger yet. It's not quite clear *where* it would make sense, and could easily lead to massive overtriggering (`python list` -> `python cheat sheet`). Perhaps this is something we could look into further if we improve the integration of metadata and cheat sheets.

pjhampton:
Hey all

[DuckDuckHack is now in Maintenance Mode](https://duckduckhack.com/) and from now on, we are only accepting issues and PRs for essential bugs and bug fixes.

Unfortunately, with the above in mind, this isn't something we can action and will be closed.

We appreciate you taking the time to contribute and apologize for not being able to triage this issue.

Issue #1528, opened 2015-09-08
CheatSheets: Support trigger on queries like "Cheat sheet for X"


moollaza:
Currently we only trigger cheat sheets when one of the trigger words comes immediately before or after one of our JSON filenames (or an alias). It would be great to support queries like "cheat sheet for vim" or "shortcuts for emacs".

Let's add some additional triggers! This will require the triggering logic to be adjusted slightly.

gautamkrishnar:
That's a good feature.... We must add this one because most of the users aren't tech savvy. So they will input the search queries in a natural language rather than typing specific trigger words:+1:

GuiltyDolphin:
Whoops, this one isn't solved quite yet!

GuiltyDolphin:
On hold pending some discussion on handling this in general.

PR #2170, opened 2016-01-30
Make Cheat Sheets more Awesome

GuiltyDolphin:
Well, they deserve it.

#### General Triggering

**The current State**

At the moment, we have many cheat sheets each designed to convey different information. Some provide language references, others keyboard shortcuts, as well as many others.

Currently, there is no discrimination when it comes to triggering - language cheat sheets will triggers on `keyboard shortcuts` just as those *intended* to convey keyboard shortcuts will. This is not ideal as it causes a lot of over-triggering (and potentially confusion!). I don't know about you, but typing in `jQuery key bindings` should not cause a language reference to appear!

**The Fix**

To fix this I've changed the way triggers work by defining categories that each have triggers associated with them; `math` triggering on `formulae`, `keyboard` triggering on `shortcuts` etc...

Initially the `template_type` of the cheat sheet will cause a default category to be assigned - thus the `keyboard` template type gets assigned the triggers for `keyboard`. Additional categories may be specified in an `additional_categories` attribute when it makes sense to trigger from another category.

#### Custom Triggers

I've seen mentions of "Oh, can I get my cheat sheet to trigger on X?" mentioned around, and often the answer is either "Use an alias" or "No".

This feature aims to fix this by providing a means of specifying additional triggers for cheat sheets.

The supported triggers are `start`, `end`, `any`, and `startend. By default the name (or an alias) of the cheat sheet is still required to trigger, by this may be turned off.

**Example**

Say we have a cheat sheet, `historical artifacts`, that displays some notable artifacts from ancient history.

Obviously the category for this cheat sheet should be `reference`. Now the cheat sheet will trigger on `historical artifacts cheat sheet`, `historical artifacts reference` etc...

Maybe we want to have it trigger on `ancient artifacts`? No problem, we just add the `historical artifacts` alias.

But hang on... Now we want it to trigger on `list of historical artifacts` and `list of ancient artifacts`, but none of the existing categories allow this!

~~Now the custom triggers come into play, by specifying:~~

```JavaScript
"triggers": {
    "start": "list of"
}
```
~~in the attributes for the cheat sheet, the desired behavior can now be achieved!~~

The new method of implementation is through the `triggers.json` file, as discussed in the comments (I might add that here).

## To-Do

- [x] Allow specifying custom triggers in JSON
- [x] Allow specifying multiple categories
- [x] Fix all the lovely `$key` warnings (need to PR `DuckDuckGo`)

---
[IA Page](https://duck.co/ia/view/cheat_sheets)


daxtheduck:
Metadata placeholder post
            
---
            
**How to find your IA page**
- New Instant Answer
              
- Existing Instant Answer

lib/DDG/Goodie/CheatSheets.pm
```diff
@@ -5,25 +5,66 @@ use JSON::XS;
...
 
+sub generate_triggers {
+    my @triggers = @all_triggers;
+    foreach my $trigger_set (values %standard_triggers) {
+        push @triggers, @{$trigger_set};
+    }
+    return uniq @triggers;
+}
+
+triggers startend => generate_triggers();
```
moollaza: not that it really matters, but you could use an anonymous function here :+1: 

```perl
triggers startend => sub {
    my @triggers = @all_triggers;
    foreach my $trigger_set (values %standard_triggers) {
        push @triggers, @{$trigger_set};
    }
    return uniq @triggers;
}
```

moollaza: At least I think so...

GuiltyDolphin: Huh, cool. I've changed it up anyway so might not still be as relevant.

GuiltyDolphin:
@duckduckgo/community-leaders Any suggestions for the categories? We can have a bit of fun and add some new ones if you like ;)

russellholt:
cc @zachthompson

zachthompson:
Hi @GuiltyDolphin, please update the description with what this PR is attempting to solve/enhance when you have enough of it complete.

GuiltyDolphin:
@zachthompson Will do - it's basically done - just adjusting now. I need review on [the standard triggers](https://github.com/GuiltyDolphin/zeroclickinfo-goodies/blob/cheatsheet-triggers/lib/DDG/Goodie/CheatSheets.pm#L21) which are what will actually trigger each category (I'll describe this in main comment).

GuiltyDolphin:
@zachthompson Want to have a look over this so far? The only pressing issue is the output warnings (I've done a PR over at `duckduckgo` to fix these).

zachthompson:
> Cheat sheet triggers are now assigned based on categories (rather than all cheat sheets triggering on all the triggers).

Can you expand on what you mean by "triggering on all the triggers"?  I don't follow how the change enhances this.

> Cheat sheets may now also specify custom triggers (only start, end, any, and startend). 

In which cases would "custom triggers" like this be used?  How would someone specify one?

GuiltyDolphin:
@zachthompson I've updated the comment to be a bit more comprehensive, hopefully it'll better convey what I'm trying to achieve.

zachthompson:
Thanks for the update.  I will give it a look.

lib/DDG/Goodie/CheatSheets.pm
```diff
@@ -55,25 +163,66 @@ sub getAliases {
...
     # If needed we could jump through a few more hoops to check
     # terms against file names.
-    open my $fh, $aliases->{join(' ', split /\s+/o, lc($_))} or return;
+    my ($was_additional, $who) = who_triggered($remainder, $req->matched_trigger);
+    open my $fh, $who or return;
 
     my $json = do { local $/;  <$fh> };
     my $data = decode_json($json);
+    unless ($was_additional || any { $_ eq $req->matched_trigger } @all_triggers) {
+        my $template_type = ($data->{template_type});
```
zachthompson: Cheat sheets without `template_type`, e.g. [gnome](https://github.com/duckduckgo/zeroclickinfo-goodies/blob/master/share/goodie/cheat_sheets/json/gnome.json), break this.

GuiltyDolphin: Why doesn't it have a template type? Surely it should be `keyboard`?

zachthompson:
Hey @GuiltyDolphin, I've given this a look over the last day.  There are some architectural changes I'd like to consider that will help us down the road, particularly if we decided to migrate some of the metadata out of the module and into a DB.  Several other syntactical modifications as well.

lib/DDG/Goodie/CheatSheets.pm
```diff
@@ -5,30 +5,113 @@ use JSON::XS;
...
-    'shortcut', 'shortcuts',
-    'symbol', 'symbols',
-    'key bindings', 'keys', 'default keys'
+sub maybe_plural {
+    my $phrase = shift;
+    return ($phrase, $phrase . 's');
+}
+
+my %standard_triggers = (
+    'keyboard' => [
```
zachthompson: Let's pull this into a triggers.json file so we don't need to edit the module to add new ones.

lib/DDG/Goodie/CheatSheets.pm
```diff
@@ -5,30 +5,113 @@ use JSON::XS;
...
-    'character', 'characters',
-    'cheat sheet', 'cheatsheet',
-    'command', 'commands',
-    'example', 'examples',
-    'guide', 'help',
-    'quick reference', 'reference',
-    'shortcut', 'shortcuts',
-    'symbol', 'symbols',
-    'key bindings', 'keys', 'default keys'
+sub maybe_plural {
```
zachthompson: This will need to be much more robust, e.g. https://en.wikipedia.org/wiki/English_plurals#Regular_plurals.  I say we drop it and just specify plurals longhand.

lib/DDG/Goodie/CheatSheets.pm
```diff
@@ -5,30 +5,113 @@ use JSON::XS;
...
+    'cheat sheet',
+    'cheatsheet',
 );
 
+sub generate_standard_triggers {
+    my @triggers = @all_triggers;
+    foreach my $trigger_set (values %standard_triggers) {
+        push @triggers, @{$trigger_set};
+    }
+    return uniq @triggers;
```
zachthompson: The uniqueness can be done above by setting each trigger as a key to a hash value, then returning the keys.  We can remove List::MoreUtils.

lib/DDG/Goodie/CheatSheets.pm
```diff
@@ -5,30 +5,113 @@ use JSON::XS;
...
+
+my %additional_triggers = (
+    any      => [],
+    end      => [],
+    start    => [],
+    startend => [],
+);
+
+# Used for determining who triggered something.
+my %trigger_lookup = ();
```
zachthompson: don't need the empty list, e.g. `my %trigger_lookup;`

lib/DDG/Goodie/CheatSheets.pm
```diff
@@ -5,30 +5,113 @@ use JSON::XS;
...
 
+sub generate_standard_triggers {
+    my @triggers = @all_triggers;
+    foreach my $trigger_set (values %standard_triggers) {
+        push @triggers, @{$trigger_set};
+    }
+    return uniq @triggers;
+}
+
+my %additional_triggers = (
```
zachthompson: `my %additional_triggers;` will suffice and also get rid of the messages.

lib/DDG/Goodie/CheatSheets.pm
GuiltyDolphin: @zachthompson T'was just a helper function so I didn't have to write so many duplicates. Wasn't intended to be a robust solution to plurals.

lib/DDG/Goodie/CheatSheets.pm
```diff
@@ -5,30 +5,113 @@ use JSON::XS;
...
+
+# Used for determining who triggered something.
+my %trigger_lookup = ();
+
+sub add_triggers {
+    my ($options, $type, $triggers) = @_;
+    my $existing = $additional_triggers{$type};
+    $triggers = [$triggers] unless ref $triggers eq 'ARRAY';
+    foreach my $trigger (@{$triggers}) {
+        if (any { $_ eq $trigger } @{$existing}) {
```
zachthompson: Would be preferable if `$existing` were a hash so we could `$existing->{$trigger}` rather than use `any`.  There are places below where this applies as well.

lib/DDG/Goodie/CheatSheets.pm
```diff
@@ -5,30 +5,113 @@ use JSON::XS;
...
+            die "Trigger '$trigger' already in use!\n";
+        };
+    };
+    my @new_triggers = (@{$existing}, @{$triggers});
+    $additional_triggers{$type} = \@new_triggers;
+    @trigger_lookup{@{$triggers}} = map { $options } (1..@{$triggers});
+}
+
+sub cheat_sheet_file_for {
```
zachthompson: This really should already be handled by `$aliases`

lib/DDG/Goodie/CheatSheets.pm
```diff
@@ -5,30 +5,113 @@ use JSON::XS;
...
+    my $name = shift;
+    $name =~ s/ /-/g;
+    return "$name.json";
+}
+
+sub cheat_names_equal {
```
zachthompson: Again, should be handled by `$aliases`

lib/DDG/Goodie/CheatSheets.pm
GuiltyDolphin: @zachthompson Gnnnh... How did I not spot that. Thanks!

lib/DDG/Goodie/CheatSheets.pm
```diff
@@ -38,15 +121,40 @@ sub getAliases {
...
+                my %options = (%default_options, %{$options});
+                add_triggers(\%options, $type, $triggers);
+            };
+        };
+
         if ($data->{'aliases'}) {
             foreach my $alias (@{$data->{'aliases'}}) {
-                $results{lc($alias)} = $file;
+                my $lc_alias = lc $alias;
+                unless (cheat_names_equal $lc_alias, $defaultName) {
```
zachthompson: Not sure why this is being done.  Checking in `%results` here seems to be fine.  if `$results{$lc_alias} ne $file` we know we have a problem.

lib/DDG/Goodie/CheatSheets.pm
zachthompson: Remove parens, not a list.

lib/DDG/Goodie/CheatSheets.pm
```diff
@@ -55,25 +163,66 @@ sub getAliases {
...
+    my $remainder = shift;
     # If needed we could jump through a few more hoops to check
     # terms against file names.
-    open my $fh, $aliases->{join(' ', split /\s+/o, lc($_))} or return;
+    my ($was_additional, $who) = who_triggered($remainder, $req->matched_trigger);
+    open my $fh, $who or return;
 
     my $json = do { local $/;  <$fh> };
     my $data = decode_json($json);
+    unless ($was_additional || any { $_ eq $req->matched_trigger } @all_triggers) {
```
zachthompson: Make it `%all_triggers` for hash look up.

lib/DDG/Goodie/CheatSheets.pm
```diff
@@ -55,25 +163,66 @@ sub getAliases {
...
+        return (0, $file);
+    };
+}
+
 handle remainder => sub {
+    my $remainder = shift;
     # If needed we could jump through a few more hoops to check
     # terms against file names.
-    open my $fh, $aliases->{join(' ', split /\s+/o, lc($_))} or return;
+    my ($was_additional, $who) = who_triggered($remainder, $req->matched_trigger);
```
zachthompson: Above this line somewhere:

```Perl
my $matched_trigger = $req->matched_trigger;
```
The value doesn't change so we can store in a temporary variable rather than calling the method repeatedly.

lib/DDG/Goodie/CheatSheets.pm
```diff
@@ -55,25 +163,66 @@ sub getAliases {
...
     my $data = decode_json($json);
+    unless ($was_additional || any { $_ eq $req->matched_trigger } @all_triggers) {
+        my $template_type = ($data->{template_type});
+        my @categories = ($template_type);
+        push @categories, @{$data->{additional_categories}}
+            if defined $data->{additional_categories};
+        my $matched = 0;
+        foreach my $category (@categories) {
+            my @triggers = @{$standard_triggers{$category}};
+            if (any { $_ eq $req->matched_trigger } @triggers) {
```
zachthompson: Convert `@triggers` to a hash.  Use `$matched_trigger` instead of method.

lib/DDG/Goodie/CheatSheets.pm
GuiltyDolphin: @zachthompson Can you expand? Do you mean the custom triggers in general?

zachthompson: So the whole idea behind `$aliases` is to do lookups:
```
this-that-other => some.json
this that other => some.json
this-other      => some.json
```
We know if something comes through that tries to do this:
```
this-other => other.json
```
that we have a problem since the files aren't equal.  We already have the file name at line 124 so why are we subbing and reconstituting it?  Even custom aliases should go through this check fine by using `%results` in `getAliases` appropriately.

GuiltyDolphin: Hmm, I see what you mean, but this is for the custom triggers, which are stored separately and can have there own clashes. I'm not sure they'll be caught by the alias checks (correct me if I'm wrong).

lib/DDG/Goodie/CheatSheets.pm
zachthompson: Yeah, that's part of why I suggested it.  Would like to avoid serializing code.  There are modules that will pluralize correctly but it's probably more overhead than necessary, e.g. 
```JSON
"make_plural":["char", "character", "command"]
```

GuiltyDolphin: @zachthompson Some of the plurals seem a bit odd `char` and `chars` for example - I would consider the singular to be incorrect given the context. `X char` versus `X chars`. I've left those in because there were in the original but my instinct would be to remove them - thoughts?

lib/DDG/Goodie/CheatSheets.pm
zachthompson: Not quite sure I follow.  This is being used for the `aliases afaik, no? For example, line 149.  We're trying to make sure that default/custom aliases don't step on one another.

lib/DDG/Goodie/CheatSheets.pm
```diff
@@ -38,15 +121,40 @@ sub getAliases {
...
+                add_triggers(\%options, $type, $triggers);
+            };
+        };
+
         if ($data->{'aliases'}) {
             foreach my $alias (@{$data->{'aliases'}}) {
-                $results{lc($alias)} = $file;
+                my $lc_alias = lc $alias;
+                unless (cheat_names_equal $lc_alias, $defaultName) {
+                    die "Cannot use an alias that is another cheat sheet's name ($lc_alias) in $file"
```
zachthompson: These two `die` statements are essentially doing the same thing: checking whether an alias is already mapped.

lib/DDG/Goodie/CheatSheets.pm
```diff
@@ -38,15 +78,40 @@ sub getAliases {
...
 
+        if ($data->{triggers}) {
+            my %default_options = (
+                require_name => 1,
+            );
+            my $options = {
+                file => $file,
+            };
+            if ($data->{triggers}->{options}) {
+                my $trigger_options = delete $data->{triggers}->{options};
```
zachthompson: Just use this in the `if` statement.  Can drop the second `->` as well.

lib/DDG/Goodie/CheatSheets.pm
```diff
@@ -38,15 +78,40 @@ sub getAliases {
...
+                my %new_options = (%{$trigger_options}, %{$options});
+                $options = \%new_options;
+            };
```
zachthompson: extra semi-colon

lib/DDG/Goodie/CheatSheets.pm
```diff
@@ -38,15 +78,40 @@ sub getAliases {
...
+        my $defaultName = $name =~ s/-/ /gr;
         $defaultName =~ s/.json//;
 
         $results{$defaultName} = $file;
 
+        if ($data->{triggers}) {
+            my %default_options = (
+                require_name => 1,
+            );
+            my $options = {
```
zachthompson: So the `file` option is only necessary if they've specified trigger options?  We can make this a normal hash so we don't have to deref it below.

lib/DDG/Goodie/CheatSheets.pm
```diff
@@ -38,15 +78,40 @@ sub getAliases {
...
+                file => $file,
+            };
+            if ($data->{triggers}->{options}) {
+                my $trigger_options = delete $data->{triggers}->{options};
+                my %new_options = (%{$trigger_options}, %{$options});
```
zachthompson:
```Perl
%options = (%options, %{$trigger_options});
```

lib/DDG/Goodie/CheatSheets.pm
```diff
@@ -38,15 +78,40 @@ sub getAliases {
...
+                $options = \%new_options;
+            };
+            while (my ($type, $triggers) = each $data->{triggers}) {
+                my %options = (%default_options, %{$options});
```
zachthompson: Can be done once outside the `while` loop.  There it could be `my $options = {}` so we take the ref once.

lib/DDG/Goodie/CheatSheets.pm
GuiltyDolphin: @zachthompson Ooohh.. That function is so I can retrieve the file name that *would* be produced by an alias - so I can check the file doesn't already exist ***and*** isn't the file for the current cheat sheet.

The reason for having the function (I believe) is that the error needs to happen for the cheat sheet defining the alias - not the cheat sheet who's name it is (thus reserving the name for that cheat sheet). Otherwise the first cheat sheet that attempted to define that name - through aliases or otherwise, would encounter the error.

zachthompson: You should be able to handle that check in `getAliases` with something like:
```Perl
if(exists $results{$alias} && ($results{$alias} ne $file) ){
    die "Collision between $results{$alias} and $file both using alias $alias";
}
```

lib/DDG/Goodie/CheatSheets.pm
GuiltyDolphin: The `file` option is included in the options hash for comparisons - that's why I've put it at the end of `my %new_options = ... (..., %options)`, so the file is included (and they cannot overwrite).

zachthompson: Gotcha, but it's ok for them to overwrite `require_name`?

GuiltyDolphin:
@zachthompson Thanks for the review! Now I know to use hashes for existence checks ;)

Hopefully I've covered all you suggested (I've replied to a couple of things - let me know if you disagree/have any thoughts).

lib/DDG/Goodie/CheatSheets.pm
```diff
@@ -38,15 +72,37 @@ sub getAliases {
...
+            };
+        }
+
         if ($data->{'aliases'}) {
             foreach my $alias (@{$data->{'aliases'}}) {
-                $results{lc($alias)} = $file;
+                my $lc_alias = lc $alias;
+                if (defined $results{$lc_alias}
+                    && $results{$lc_alias} ne $file) {
+                    die "Cannot use an alias that is another cheat sheet's name ($lc_alias) in $file"
```
zachthompson: I don't think we need this distinction.  If we get here, we definitely want to `die`.  All we need to know are the two files names so we can deal with one of them.

GuiltyDolphin: You mean don't prioritize for the cheat sheet who *has* the name that is being used? Just tell them which two files are clashing?

zachthompson: No, we will prioritize for the one with the file name. So if we get
```
vim.json and vi.json both using alias 'vi'
```
it's pretty clear who wins.

GuiltyDolphin: Ah, I see what you mean, alrighty then.

lib/DDG/Goodie/CheatSheets.pm
```diff
@@ -5,30 +5,58 @@ use JSON::XS;
...
-    'shortcut', 'shortcuts',
-    'symbol', 'symbols',
-    'key bindings', 'keys', 'default keys'
-);
+my $triggers_json = share('triggers.json')->slurp();
+my %standard_triggers = %{decode_json($triggers_json)};
+
+# All cheat sheets are triggered by these.
+my %all_triggers;
+@all_triggers{(
```
zachthompson: How about moving these as well?

share/goodie/cheat_sheets/triggers.json
```diff
@@ -0,0 +1,41 @@
+{
+    "keyboard": [
```
zachthompson: If we changed these to hashes we could potentially turn them off/on:
```JSON
{
  "quick reference":1,
  "reference":0,
   etc
}
```

lib/DDG/Goodie/CheatSheets.pm
```diff
@@ -5,30 +5,58 @@ use JSON::XS;
...
-    'cheat sheet', 'cheatsheet',
-    'command', 'commands',
-    'example', 'examples',
-    'guide', 'help',
-    'quick reference', 'reference',
-    'shortcut', 'shortcuts',
-    'symbol', 'symbols',
-    'key bindings', 'keys', 'default keys'
-);
+my $triggers_json = share('triggers.json')->slurp();
```
zachthompson: Let's consolidate all of this trigger building, e.g. to line 52, into something like `build_triggers`, including the standard`startend`.  It will be a bit more intuitive and we can simplify:
```Perl
my $triggers = build_triggers;
while(my ($type, $trig) = each %$triggers){
    triggers $type => ( @$trig );
}
```
We could even put the `while` inside `build_triggers` since we don't need the triggers after they're used.  The return value could be `$trigger_lookup` which you are using during run time.

lib/DDG/Goodie/CheatSheets.pm
```diff
@@ -5,30 +5,58 @@ use JSON::XS;
...
+    @existing{@{$additional_triggers{$type}}} = 1
+        if defined($additional_triggers{$type});
+    foreach my $trigger (@triggers) {
+        if (exists($existing{$trigger})) {
+            die "Trigger '$trigger' already in use!\n";
+        };
+    };
+    my @new_triggers = (keys %existing, @triggers);
+    $additional_triggers{$type} = \@new_triggers;
+    @trigger_lookup{@triggers} = map { $options } (1..@triggers);
```
zachthompson: As each trigger is verified above, just add it to this structure as a last step:
```Perl
$trigger_lookup{$trigger} = ...
```

lib/DDG/Goodie/CheatSheets.pm
```diff
@@ -5,30 +5,58 @@ use JSON::XS;
...
+my %all_triggers;
+@all_triggers{(
+    'help',
+    'cheat sheet',
+    'cheatsheet'
+)} = undef;
+
+sub generate_standard_triggers {
+    my %triggers;
+    @triggers{keys %all_triggers} = undef;
```
zachthompson: Shouldn't need to do it this way if we structure the JSON correctly.  Just iterate through the hash and add them to the master triggers hash and the lookup.

share/goodie/cheat_sheets/triggers.json
GuiltyDolphin: Nice idea - might even allow for mixing categories but having some of the triggers switched off - neat.

zachthompson: One other tweak that *might* be useful here is grouping them under trigger type, e.g. all `startend` initially.  This would allow us to change the type for the defaults if necessary.  When we finally have the triggers loaded, I think the structure we're look for is
```Perl
$triggers->{type}{trigger}
```
This file sort of breaks that pattern since we're assuming they're all `startend`.

There's also the idea that we could force triggers for specific IAs to be specified here so they centralized in one place and easily scanned.  Then we wouldn't need to load each JSON file to get custom triggers.  @GuiltyDolphin @Zaahir, what do think about this idea?

GuiltyDolphin: Hmm... You mean IAs that are essentially cheat sheets with a bit more... Gusto?

CC @moollaza

zachthompson: I don't...think so :D  Which part are you talking about @GuiltyDolphin?

GuiltyDolphin: @zachthompson The last part, specifying triggers for certain IAs in the JSON.

Do you mean exclusively custom triggers for the JSON cheat sheets, or also IAs that act like cheat sheets but have to be separate due to additional functionality?

zachthompson: Yes, just the JSON cheat sheets.  Part of our discussion has been about not wanting to encourage cheat sheet contributors to even think about triggers.  It only needs to be done in rare cases and would probably be better done by staff/the more experienced.  We would still need to document it and then people start making bad assumptions, e.g. I need one.  So adding custom triggers to the individual cheat sheets sets us up for headaches, whereas centralizing them here might limit it to the more experienced.

I also don't like having to root through the JSON to find which cheat sheets are doing this.

GuiltyDolphin: Ah, okay. So maybe we could have them (or the experienced) specify custom categories that are named after the cheat sheet in the triggers file. Will want to be able to specify the location (`start`, `end`, `startend` etc...)...

Maybe we can use the trigger positions as the values? Then `0` can still be disabled, but `start` or whatever will cause it to trigger in that position.

zachthompson: Categories might be one way.  I was just thinking a small adjustment to the structure like:
```JSON
"default_triggers":{
  "startend":{
    "template_type1":{
        "trigger1":1
    }
  }
},
"my_special_cheat_sheet":{
  "start":{
     "a custom trigger for":1
   }
},
"another_non_standard_cheat_sheet":{
   "any":{
    "crazy query":1
   },
   "start":{
      "In the beginning..":1
    }
}
```

GuiltyDolphin: Ah, okay.

Categories would be a bit more flat:
```JavaScript
"keyboard": {
    "keyboard shortcuts": "startend"
},
"my_cheat_sheet_category": {
    "my trigger and then...": "start",
    "disabled for now": 0
}
```

But yeah, maybe having all the defaults in one place is a good idea!

zachthompson: The custom triggers would need to link to a specific cheat sheet, i.e. the ID, so we could automatically derive it.  Otherwise they would have to add something to their cheat sheet to link it to the category.

GuiltyDolphin: Yeah, I was thinking they would specify the category as one of the `additional_categories`, but I think your idea will work better (less maintenance).

GuiltyDolphin:
@zachthompson I've moved the additional triggers into the main `triggers.json` file - I'm working on turning the trigger sets into hashes. Need to clean this up a bit...

GuiltyDolphin:
@zachthompson I've converted the triggers to be hashes.

I'm thinking maybe it would be worth splitting the one-to-one relationship between categories and template types. It would be fairly trivial to implement (the infrastructure is essentially already there) and would hopefully make it easier to add new triggers to multiple template types. 

What do you think?

GuiltyDolphin:
@moollaza @zachthompson I'm mostly done here (I think). I'm happy to open another PR after and implement anything else and tidy up a bit more, but given we have a lot of cheat sheets to review (and this implements a lot of tests to reduce the amount of "do this, not like that!" we need to do) I'd say I'm happy for final review.

Once that's done, I'll open another PR for minor tweaks and tidying.

zachthompson:
@GuiltyDolphin For now I think we'll hold off on this change. The _complexity_ of the updated solution raises concerns with respect to _maintainability_ and, to a lesser extent,  performance, which has reverted to before we started tweaking it.  It would need to be a simpler solution to move in this direction as there's a good chance we'll consider migrating triggers out of the modules at some point.

GuiltyDolphin:
@zachthompson Hmm... Okay, you're referring to the complexity of the Perl, right? Rather than how new triggers would be instantiated?

I think I'll pull the tests out to a separate PR then, and work on simplifying the solution here.

zachthompson:
@GuiltyDolphin  I'm referring to mostly how the JSON is processed and used.  I think you were pretty close after our initial round of tweaks.  From reviewing what you're trying to accomplish, it seems to me that it comes down to two things:

1. Create a hash to initialize the triggers (this can be discarded)
2. Create a lookup (probably another hash) to filter queries based on trigger

It should be pretty straightforward to create both at the same time in `get_triggers` (or whatever you name it) from the JSON.

It's probably worthwhile to reconsider what the easiest way would be to filter the query since it will determine the structure of #2 above. You have three conditions:

1. ALL (scalar)
2. template_type (array)
3. specific ID (scalar)

These should be easily testable in an `if-elsif` with what you get back from the lookup hash.

The performance was initially a bit over twice as slow as the current implementation.  We had tweaked it down to ~60% slower but it has now gone back to 2x.  Like I said, the lesser of the concerns if it provides useful functionality but simplifying will definitely improve it.

GuiltyDolphin:
@zachthompson I see - thanks, I'll have a bit of mull over this before I start messing around with the code again (don't want to make it worse!). I've pushed up to `gd/cheatsheet-triggers`, so if you wanted to tweak anything feel free.

zachthompson:
:+1: 
I'm hopeful it can be simplified to:
```Perl
my $trigger_lookup = get_triggers();
```

GuiltyDolphin:
@zachthompson Yeah, I'm thinking of just removing any support for custom triggers (initially), getting it to work really well with the standard triggers then trying to re-introduce the custom triggers.

Hopefully that'll simplify the process.

GuiltyDolphin:
@zachthompson Phew! It looks *a lot* better now.

Hopefully much more robust, and it is definitely simpler!

It should deploy to Beta fine now - I've removed the tests which were causing it to fail (moved to #2281) and changes to misc cheat sheets. It just modifies core functionality now.

/cc @moollaza

share/goodie/cheat_sheets/triggers.json
```diff
@@ -0,0 +1,61 @@
+{
+    "standard": {
```
GuiltyDolphin: Humm... Not sure I need this anymore. The distinction between custom triggers and standard triggers is based solely upon the name `X_cheat_sheet` vs `some_category`. The ID of the cheat sheet is used to specify the custom triggers, so it should be fairly easy to maintain.

share/goodie/cheat_sheets/triggers.json
```diff
@@ -0,0 +1,61 @@
...
+          }
+      },
+      "standard": {
+          "startend": {
+              "cheat sheet": 1,
+              "cheatsheet": 1,
+              "help": 1
+          }
+      }
+    }
```
GuiltyDolphin: Example of custom triggers:

```json
"id_of_cheat_sheet": {
    "startend": {
        "some disabled trigger": 0,
        "not disabled": 1,
        "disabled with more options": {
            "disabled": 1
        },
        "does not require the name of the cheat sheet": {
            "require_name": 0
        }
    }
}

Can use `start`, `end`, `startend`, and `any`.

**Comment by zachthompson** [comment]
@GuiltyDolphin It's a bit better but I think still a bit too complicated for what you're trying to do.  The performance of `remainder` is [slightly better](https://zach.duckduckgo.com/nytprof/cheatsheets-2170-mod3/CheatSheets-pm-8-line.html#209) (here is [master](https://zach.duckduckgo.com/nytprof/cheatsheets-master/CheatSheets-pm-8-line.html#58) for reference.)

Yesterday I was playing around with [this solution](https://github.com/duckduckgo/zeroclickinfo-goodies/compare/zt/cs-trig?w=1) based on your json file and retaining custom triggers.

**Comment by GuiltyDolphin** [comment]
@zachthompson I've removed some more code that didn't *really* need to be there.

It doesn't seem much more complex than your solution to me - there's a few extra checks (to allow for a bit of extra customizing), but apart from that, unless I'm missing something it doesn't seem that bad.

Any improvements on time with the recent removals?

**Comment by zachthompson** [comment]
The complexity is still there, only shifted.  Now we have two files we need to consider when adding a trigger, settings, options, and many areas of the code where it is not at all clear why it doing what it is.  I will make a few comments with respect to these.

File: lib/DDG/Goodie/CheatSheets.pm [hunk 29]
```diff
@@ -5,30 +5,92 @@ use JSON::XS;
...
-);
-
-sub getAliases {
+# Instantiate triggers as defined in 'triggers.json', return a hash that
+# allows category and/or cheat sheet look-up based on trigger.
+sub generate_triggers {
+    my $aliases = shift;
+    my $triggers_json = share('triggers.json')->slurp();
+    my $json_triggers = decode_json($triggers_json);
+    my $trigger_lookup = make_all_triggers($aliases, $json_triggers);
```
zachthompson: Why do we need `make_all_triggers`?  That's what this sub should be doing.

lib/DDG/Goodie/CheatSheets.pm
```diff
@@ -5,30 +5,92 @@ use JSON::XS;
...
+    return $trigger_lookup;
+}
+
+sub make_all_triggers {
+    my ($aliases, $spec_triggers) = @_;
+    # This will contain the actual triggers, with the triggers as values and
+    # the trigger positions as keys (e.g., 'startend' => ['foo'])
+    my %triggers = ();
```
zachthompson: Don't need to initialize like this.  I've mentioned this before.

lib/DDG/Goodie/CheatSheets.pm
```diff
@@ -5,30 +5,92 @@ use JSON::XS;
...
+    # This will contain a lookup from triggers to categories and/or files.
+    my $trigger_lookup = {};
+
+    # Default settings for custom triggers.
```
zachthompson: Why do we need these?

lib/DDG/Goodie/CheatSheets.pm
```diff
@@ -5,30 +5,92 @@ use JSON::XS;
...
+    my %defaults = (
+        require_name => 1,
+        full_match   => 1,
+    );
+
+    while (my ($name, $trigger_setsh) = each $spec_triggers) {
+        if ($name =~ /cheat_sheet$/) {
```
zachthompson: What is the purpose of this entire `if` and the use of `aliases`?  If it's not in aliases, it will get kicked out in remainder anyhow.

lib/DDG/Goodie/CheatSheets.pm
```diff
@@ -5,30 +5,92 @@ use JSON::XS;
...
+            $file =~ s/_/ /g;
+            $file = $aliases->{$file};
+            die "Bad ID: '$name'" unless defined $file;
+            $name = $file;
+        }
+        while (my ($trigger_type, $triggersh) = each $trigger_setsh) {
+            while (my ($trigger, $opts) = each $triggersh) {
+                next if $opts == 0;
+                # Normalize options to use default options where not provided.
+                my %opts = %defaults;
```
zachthompson: This `opts` functionality, why do we need this?

lib/DDG/Goodie/CheatSheets.pm
```diff
@@ -5,30 +5,92 @@ use JSON::XS;
...
+    }
+    while (my ($trigger_type, $triggers_a) = each %triggers) {
+        triggers $trigger_type => (keys %{$triggers_a});
+    }
+    return $trigger_lookup;
+
+}
+
+# Parse the category map defined in 'categories.json'.
+sub get_category_map {
```
zachthompson: Why the introduction of categories?

lib/DDG/Goodie/CheatSheets.pm
GuiltyDolphin: It is to get it to point at the file directly - rather than have additional checks when handling queries.

The purpose is to check if the set of triggers is standard or non-standard - you use the ID of the cheat sheet for the name of the additional triggers.

lib/DDG/Goodie/CheatSheets.pm
```diff
@@ -38,42 +100,91 @@ sub getAliases {
...
+    open my $fh, $file or return;
     my $json = do { local $/;  <$fh> };
     my $data = decode_json($json);
+    return $data;
+}
+
+# Attempt to retrieve the JSON data based on the used trigger.
+sub get_cheat_json {
+    my ($remainder, $req) = @_;
+    my $trigger = $req->matched_trigger;
```
zachthompson: `matched_trigger` doesn't normalize, e.g. it will return `Cheat     Sheet` if that's what the user specified.

lib/DDG/Goodie/CheatSheets.pm
```diff
@@ -38,42 +100,91 @@ sub getAliases {
...
+    my $file;
+    my $lookup = $trigger_lookup->{$trigger};
+    if ($lookup->{is_custom}) {
+        return if $lookup->{options}{full_match} && $remainder ne '';
```
zachthompson: What is this use case of requiring no remainder?

GuiltyDolphin:
@zachthompson By the way, I've run some tests using NYTProf (same benches as you) and the new version is consistently coming out on top (about 1 second faster on average), you getting anything like this?

lib/DDG/Goodie/CheatSheets.pm
GuiltyDolphin: In terms of casing?

lib/DDG/Goodie/CheatSheets.pm
GuiltyDolphin: They were there all along (just as template types).

See the triggers for the `math` category - `math` doesn't have a template type, but can be specified as an additional category.

lib/DDG/Goodie/CheatSheets.pm
GuiltyDolphin: The standard one - this is the general default.

In terms of why you *would* want a remainder, well, if the custom trigger is quite specific and is likely to be relevant to queries with extra phrasing like `what is` around it.

If you feel we shouldn't have this I can remove it - it does show that additional options *do* work though.

zachthompson:
I've just been looking at the performance of `remainder` since that's what will mostly matter when triggering and only in a general sense to make sure it's not way out of bounds.  The latest changes are hovering around [~103-106ms](https://zach.duckduckgo.com/nytprof/cheatsheets-2170-mod3/CheatSheets-pm-8-line.html#209).

lib/DDG/Goodie/CheatSheets.pm
GuiltyDolphin: For specifying things like:

* `require_name`, boolean: When 'true' means that the name of the cheat sheet (or one of the aliases) ***must*** be in the remainder, when it is not specified it allows for standalone triggers which are independent of the name.
* `full_match`, boolean: When 'true' there must be no remainder left in the query after processing, otherwise remainder is allowed.

In short: It is to allow additional options in the triggers - seeing as they do not have direct access to Perl to do any processing themselves.

lib/DDG/Goodie/CheatSheets.pm
GuiltyDolphin: See my other comments regarding `$opts` - this is to ensure the sensible default values are followed.

lib/DDG/Goodie/CheatSheets.pm
GuiltyDolphin: Whoops, yeah, sorry :blush:

lib/DDG/Goodie/CheatSheets.pm
GuiltyDolphin: Humm... Sorry, they can be merged.

My personal preference (likely due to mostly using functional programming) is to have lots and lots of functions that do *one* thing, and *one* thing only - so I sometimes pull out bits of code - I'm still learning the Perl idioms.

I'll merge that into `make_all_triggers` (or the other way round).

lib/DDG/Goodie/CheatSheets.pm
zachthompson: Let me see if I understand.  This is saying return if `full_match` (I don't even know what this means) is true *and* there is a `remainder` or what is the same
```Perl
return if $lookup->{options}{full_match} && $remainder;
```
So this would get triggered by "some totally custom cheat sheet trigger" that matches the entire trigger, looking up what is essentially an `alias` for the file?

lib/DDG/Goodie/CheatSheets.pm
zachthompson: And irregular spacing.  Should have quoted it:
```
my crazy Cheat    Sheet

will return a trigger of

Cheat    Sheet
```

lib/DDG/Goodie/CheatSheets.pm
zachthompson: I'm all for that *if* they're going to be used elsewhere and/or they break the complexity down into logical chunks.

lib/DDG/Goodie/CheatSheets.pm
zachthompson: Can you give an example of both of these and how they are specified?

lib/DDG/Goodie/CheatSheets.pm
GuiltyDolphin: Kindah - `full_match` just means, as you say, that the trigger matches the *whole* query, rather than just part.

This isn't *super* important to the implementation, it is the highest level and doesn't affect anything else so it can be removed if you want. It is more a proof of concept of the options than anything else.

lib/DDG/Goodie/CheatSheets.pm
GuiltyDolphin: Huh? Are you saying that *should*, or *shouldn't* trigger it?

GuiltyDolphin:
@zachthompson The way I see the different components used is:

**Users:**

1. The *template type* is specified in your `cheatsheet.json` file, this should hopefully automatically cover most triggering.
2. Maybe some preset triggers *are* relevant to your cheat sheet, but are not default for your chosen *template type*, in that case you specify additional *categories*.
3. Even then it isn't enough, so you *(get permission and then)* add custom triggers to the `triggers.json` file under the id of your cheat sheet.

**Maintainers**

1. New standard triggers to be added (or a new category): done by adding new triggers to the `triggers.json` file, or adding a new category with the appropriate triggers.
2. New *template type*, or changing the triggers for an existing *template type*: the templates in `categories.json` are modified to change which *categories* are allocated to each *template type*.

**Options**

After this discussion it's looking less and less like the options will *actually* be much more useful than all this stuff provided by the rest of the trigger system.

I'll be happy to remove the options, which should simplify the code and solution somewhat (basically a few commits back); and if we ever want them... We can just re-add them I guess.

What do you say?

zachthompson:
@GuiltyDolphin We had the discussion about `template_type` vs. `category` as well.  Right now they are sort of the same but it's not guaranteed that we won't want to have `template_type`s in the future that don't map easily to triggers.  So I understand the need for categories.

With respect to the custom triggers, I would still like to see:

1. how each type looks in the JSON
2. an example of where a `full_match` would be useful

lib/DDG/Goodie/CheatSheets.pm
zachthompson: Also, couldn't this simplify to `full_match` only?  If `full_match` is specified it means there's no remainder and hence no `name`.  Otherwise `name` is by default required (and will fail if it doesn't match anyhow.)

lib/DDG/Goodie/CheatSheets.pm
```diff
@@ -38,42 +95,91 @@ sub getAliases {
...
-handle remainder => sub {
-    # If needed we could jump through a few more hoops to check
-    # terms against file names.
-    open my $fh, $aliases->{join(' ', split /\s+/o, lc($_))} or return;
+my $category_map = get_category_map();
+
+my $trigger_lookup = generate_triggers($aliases);
+
+# Retrieve the categories that can trigger the given cheat sheet.
+sub supported_categories {
```
zachthompson: In general, we shouldn't need to call this every time through.  These categories don't change.  It should be static and built during the trigger/alias generation.

GuiltyDolphin:
@zachthompson Oh, custom triggers... I think I had a previous comment... [here we go](https://github.com/duckduckgo/zeroclickinfo-goodies/pull/2170#discussion-diff-52844144); that should show the *general* syntax - although if we remove the options then it is just the top two that apply.

Yeah... I had a bit of a think and I'm not sure how useful it is (hence my suggestion with removing the `options`). Looking back I think the combination of aliases with *general* custom triggers should suffice for almost everything - but if a case comes up that requires options I guess we can reconsider.

**Syntax in `cheatsheet.json`**

```json
...
"template_type": "reference",
"categories": ["math"],
...
```

**In `triggers.json`**

```json
...
"id_cheat_sheet": {
    "start": {
        "enabled": 1
    },
    "any": {
        "disabled": 0
    }
},
...
```

**In `categories.json`**

```json
...
"name of template": {
    "category 1": 1,
    "disabled category": 0
},
...
```

zachthompson:
Got it.  Give it a shot and let's see how it looks.

GuiltyDolphin:
@zachthompson Done. I'm *dying* to check this out on Beta :stuck_out_tongue:

zachthompson:
This is getting closer.  Like the triggers, I don't think we want the categories in individual json files.  Aliases can be there since they're totally free form.  Categories, on the other hand, require that the user knows which are available in the json file.  This will be problematic if we move to a cheat sheet creator.  As with the triggers, it also sort of encourages cheat sheet creators to use the feature.

I'll try to make some comments about this and a few other things I've seen in the latest update soon.

GuiltyDolphin:
@zachthompson Humm... I see what you mean... There's less scope for error if we fully generate the triggers from the `template_type`.

I was planning on documenting the available categories in the docs though - so hopefully people'd be able to find out the available categories.

Triggers on the other hand - you'd always need special permission for those - especially since you'd fail the `number of files test` for cheat sheets on Beta without getting some special access.

lib/DDG/Goodie/CheatSheets.pm
```diff
@@ -5,31 +5,77 @@ use JSON::XS;
...
+# trigger.
+sub generate_triggers {
+    my $aliases = shift;
+    my $triggers_json = share('triggers.json')->slurp();
+    my $json_triggers = decode_json($triggers_json);
+    # This will contain the actual triggers, with the triggers as values and
+    # the trigger positions as keys (e.g., 'startend' => ['foo'])
+    my %triggers;
+    # This will contain a lookup from triggers to categories and/or files.
+    my $trigger_lookup = {};
```
zachthompson: Like above don't need to initialize.  Probably better as a normal hash which you take the reference of when returning.

lib/DDG/Goodie/CheatSheets.pm
```diff
@@ -5,31 +5,77 @@ use JSON::XS;
...
+    my @additional_categories = @{$data->{categories}}
+        if defined $data->{categories};
+    my %categories = %{$category_map->{$template_type}};
+    my @categories = (@additional_categories,
+                      grep { $categories{$_} } (keys %categories));
+    return \@categories;
+}
+
+# Initialize aliases and categories.
+sub get_aliases_categories {
```
zachthompson: I don't think we want to conflate aliases and categories.  I'll make a comment about the structure of json on that file.

share/goodie/cheat_sheets/triggers.json
```diff
@@ -0,0 +1,59 @@
+{
```
zachthompson: In general, this file has become complex enough that we probably want to move it to YAML so we can document here what the options are.  The categories file can probably be moved into it as well.  I imagine a high-level structure like:
```
default_triggers
additional_categories
custom_triggers (and categories)
```
Each of these sections needs to have reasonable documentation on what is allowed/available.  What do you think, @GuiltyDolphin?  To me this seems like fairly detailed config info we understand now but will become hazy as time goes on.

share/goodie/cheat_sheets/categories.json
```diff
@@ -0,0 +1,32 @@
+{
+    "keyboard": {
+        "keyboard": 1,
+        "reference": 1,
+        "standard": 1
```
zachthompson: `standard` should be added by default.  If we want to allow a category to say `standard = 0` that would be different.  But that's probably too much complexity already.

share/goodie/cheat_sheets/triggers.json
GuiltyDolphin: @zachthompson Humm, sure.

To clarify, do you mean something like:

```yaml
standard triggers (categories)
mapping between templates and categories
custom triggers
custom categories
```

I'm all for moving it out so we can better document it - but I'd like to further discuss the structure of this first.

lib/DDG/Goodie/CheatSheets.pm
```diff
@@ -38,42 +84,73 @@ sub getAliases {
...
+
+# Attempt to retrieve the JSON data based on the used trigger.
+sub get_cheat_json {
+    my ($remainder, $req) = @_;
+    my $trigger = (lc $req->matched_trigger) =~ s/(\t|\s{2,})/ /gr;
+    my $lookup = $trigger_lookup->{$trigger};
+    my $file = $aliases->{join(' ', split /\s+/o, lc($remainder))} or return;
+    my $data = read_cheat_json($file) or return;
+    return $data if defined $lookup->{$file};
+    my @allowed_categories = @{$categories->{$file}};
```
zachthompson: I think one of the reasons this is slightly slower is because every query has to be checked against an array of categories, except custom triggers (and we don't have any atm.)  The default triggers that apply to everything shouldn't need to do this either, e.g. see the [logic here](https://github.com/duckduckgo/zeroclickinfo-goodies/compare/zt/cs-trig#diff-a3bab390041f33c0ff8542d706b02da1R116).

lib/DDG/Goodie/CheatSheets.pm
GuiltyDolphin: Humm... I was putting them together so we didn't have to map through all the files more than once...

Oh, in your other comment did you mean to have additional categories for the cheat sheets specified *in* the YAML?

lib/DDG/Goodie/CheatSheets.pm
```diff
@@ -38,42 +84,73 @@ sub getAliases {
...
+    foreach my $category (@allowed_categories) {
+        return $data if defined $lookup->{$category};
+    }
+}
+
+handle remainder => sub {
+    my $remainder = shift;
+
+    my $data = get_cheat_json($remainder, $req) or return;
```
zachthompson: Not a big deal but `get_cheat_json` is probably not complex enough to be on it's own.

lib/DDG/Goodie/CheatSheets.pm
```diff
@@ -38,42 +84,73 @@ sub getAliases {
...
+    open my $fh, $file or return;
     my $json = do { local $/;  <$fh> };
     my $data = decode_json($json);
+    return $data;
+}
+
+# Attempt to retrieve the JSON data based on the used trigger.
+sub get_cheat_json {
+    my ($remainder, $req) = @_;
+    my $trigger = (lc $req->matched_trigger) =~ s/(\t|\s{2,})/ /gr;
```
zachthompson: Slow.  Probably better done as in line 129.

lib/DDG/Goodie/CheatSheets.pm
zachthompson: Yeah, exactly.  I think we want to keep the individual cheat sheet files simple for now.

share/goodie/cheat_sheets/categories.json
GuiltyDolphin: @zachthompson Sure - I was just putting it there for consistency - I'll move it out.

share/goodie/cheat_sheets/triggers.json
zachthompson: Fair enough.  What I meant by `additional_categories` was essentially what's in the categories.json.  Breaking the custom into two keys would probably be fine although it wouldn't be hard to keep them in one, e.g.
```YAML
id_cheat_sheet:
  additional_categories: ["off topic", "entertainment"]
  triggers:
  - great stuff here
  - you want to read this
```
I'll leave that up to you, as long as it's as simple as can be.

share/goodie/cheat_sheets/categories.json
GuiltyDolphin: Humm.. Not 100% sure but I think we need to keep it like this with my recent changes - at least for now.

GuiltyDolphin:
@zachthompson I've redone the trigger matching *(again)*, looking better?

The file actually seems a bit more compact now - three distinct routines + the handler.

share/goodie/cheat_sheets/categories.json
zachthompson: That's fine, more important stuff to worry about.

share/goodie/cheat_sheets/triggers.json
GuiltyDolphin: Humm.. With the YAML the enable/disable looks pretty ugly...

```YAML
category:
  startend:
    trigger1: 1
    trigger2: 0
```

I'm thinking it might be best just to have them as a list (as you have done). I don't think being able to enable/disable them will actually be hugely beneficial (we can just remove/add them - or comment them out tbh).

zachthompson: Yeah, I thought about that too.  We can always look at the history to find triggers we may have removed.  I'm ok with it either way.

GuiltyDolphin:
@zachthompson How's that for the overall structure (YAML)?

In terms of the documentation - I feel that referring to the `categories` section is probably best (from the `custom_triggers` section) to show what categories are available - no point duplicating the information.

share/goodie/cheat_sheets/triggers.yaml
```diff
@@ -0,0 +1,88 @@
+categories:
```
zachthompson: Let's comment the structure here as well, maybe with a reference to [word trigger](http://docs.duckduckhack.com/backend-reference/triggers-handle-functions.html).  Also, note that standard are the default triggers for all cheat sheets.  Actually, let's just change that to default.  Standard doesn't quite convey the same message.

zachthompson:
Overall, I think the structure looks fine.  I would lean towards over-documenting to make it easier for us to "do the right thing" in the future ;)

GuiltyDolphin:
@zachthompson Humm.. How about if I give a description of what the categories are for, use cases, stuff like that in the category section, then refer to that when I'm talking about categories in the custom triggers section? Considering the custom triggers use the same trigger format as well as specify triggers.

share/goodie/cheat_sheets/triggers.yaml
GuiltyDolphin: Sure, was just the first name I came up with ;)

Want to knock #1905 on the head while we're at it?

GuiltyDolphin:
@zachthompson I know that once this goes on Beta it will probably be tested to death (as it affects a lot of cheat sheets), but what say you about having a special `test-cheatsheet.json` file (or a directory with test files) for ensuring different cheat sheet features work?

share/goodie/cheat_sheets/triggers.yaml
zachthompson: Sure, are you thinking a new category?

GuiltyDolphin: @zachthompson Yeah, for things that contain a list reference - e.g `holiday items list`, `list of things to take on holiday` etc...

I don't think that second one would be supported, but there is an issue #1528 that wants this sort of behavior - maybe we could try and get that one in this PR too? (I'll have a mull over how that would be done - given custom triggers and whatnot).

lib/DDG/Goodie/CheatSheets.pm
```diff
@@ -5,31 +5,85 @@ use JSON::XS;
...
-    'example', 'examples',
-    'guide', 'help',
-    'quick reference', 'reference',
-    'shortcut', 'shortcuts',
-    'symbol', 'symbols',
-    'key bindings', 'keys', 'default keys'
-);
-
-sub getAliases {
+my $yaml_triggers = share('triggers.yaml')->slurp();
```
zachthompson: Use `LoadFile` for these two lines.

share/goodie/cheat_sheets/triggers.yaml
```diff
@@ -0,0 +1,142 @@
...
+    startend:
+      - "commands"
+      - "symbols"
+      - "chars"
+      - "characters"
+
+# You shouldn't need to touch this.
+#
+# Specifies the categories that are automatically assigned to any
+# cheat sheet using the given template_type.
```
zachthompson: Shouldn't this mention custom categories as well?

GuiltyDolphin:
@zachthompson I've updated the docs a bit, hopefully a bit clearer now.

How's this looking in terms of the triggering? Happy with the current method?

And I think we can deploy to beta again now, @moollaza, there shouldn't be any conflicts.

lib/DDG/Goodie/CheatSheets.pm
```diff
@@ -5,31 +5,84 @@ use JSON::XS;
 use DDG::Goodie;
 use DDP;
 use File::Find::Rule;
+use JSON;
```
zachthompson: Why did this change from XS?

GuiltyDolphin: @zachthompson My bad, must have accidentally changed it.

lib/DDG/Goodie/CheatSheets.pm
```diff
@@ -5,31 +5,84 @@ use JSON::XS;
...
+use YAML qw(LoadFile);
```
zachthompson: Even though not as significant, we should use YAML::XS as well.

lib/DDG/Goodie/CheatSheets.pm
```diff
@@ -5,31 +5,83 @@ use JSON::XS;
...
+
+    # This will contain the actual triggers, with the triggers as values and
+    # the trigger positions as keys (e.g., 'startend' => ['foo'])
+    my %triggers;
+    # This will contain a lookup from triggers to categories and/or files.
+    my %trigger_lookup;
+
+    while (my ($name, $trigger_setsh) = each %spec_triggers) {
+        my $is_standard = 1;
+        if ($name =~ /cheat_sheet$/) {
```
zachthompson: This can be moved directly to line 60, e.g.
```Perl
if(/cheat_sheet$/){ ...
```

zachthompson:
@GuiltyDolphin looking pretty good now, save for a few minor tweaks.

@moollaza Do the groupings of triggers look reasonable to you?  The only one that really stands out for me is `guide` which seems a bit broader than `language`.

zachthompson:
The bash cheat sheet is interesting.  Probably should be `template_type: terminal`, though some of the opening commands don't look great with that template.   `bash commands` should definitely trigger it.

GuiltyDolphin:
@zachthompson Hmm.. Yeah, I agree. I'll do something about that - I'm going to clear up triggers a bit.

share/goodie/cheat_sheets/triggers.yaml
```diff
@@ -0,0 +1,233 @@
...
+    - "default"
+    - "keyboard"
+    - "reference"
+  language:
+    - "default"
+    - "language"
+  links:
+    - "default"
+    - "links"
+  math:
```
GuiltyDolphin: Just realised, `math` isn't a template type :/

share/goodie/cheat_sheets/triggers.yaml
```diff
@@ -0,0 +1,155 @@
...
+  terminal:
+    startend:
+      - "commands"
+      - "symbols"
+      - "chars"
+      - "characters"
+
+# You shouldn't need to touch this.
+#
+# Specifies the categories that are automatically assigned to any
```
moollaza: Seems like a word is missing here? What "Specifies"?

share/goodie/cheat_sheets/triggers.yaml
```diff
@@ -0,0 +1,155 @@
...
+#   additional_categories:
+#     - "math"
+#   triggers:
+#     startend:
+#       - "formula sheet"
+#       - "equation sheet"
+#
+# And now it will be able to trigger on 'calculus equations',
+# 'calculus formula sheet', 'equation sheet calculus' etc...
+custom_triggers:
```
moollaza: Is the expectation that this will be used sparingly or often? If it's that latter I'm a little worried about merge conflicts when multiple PRs are trying to add their own triggers

share/goodie/cheat_sheets/triggers.yaml
```diff
@@ -0,0 +1,155 @@
...
+      - "cheatsheet"
+      - "help"
+      - "cheat sheet"
+  terminal:
+    startend:
+      - "commands"
+      - "symbols"
+      - "chars"
+      - "characters"
+
```
moollaza: I think we need a category like "gaming" for triggers like "cheats, cheat codes, codes"

share/goodie/cheat_sheets/triggers.yaml
```diff
@@ -0,0 +1,155 @@
+# Categories represent sets of triggers that can be
+# assigned to templates or individual cheat sheets.
+#
+# Each category consists of a name, along with a set
+# of trigger types (specifically WORD triggers - start,
+# end, startend, and any) and the associated triggers.
```
moollaza: Are we sure we want to support `any` and `start`?

I can only think of a few `start` triggers: `how to`, `list of`, `syntax for`, `help with` -- all of which have a secondary word indicating the trigger is coming before the subject

These all still feel like `end` triggers (which is what we're currently using).

share/goodie/cheat_sheets/triggers.yaml
GuiltyDolphin: It's referring to the `template_map`, which is the target of this section of the documentation. I can clarify if you wish.

share/goodie/cheat_sheets/triggers.yaml
GuiltyDolphin: @moollaza I think we're aiming for semi-sparingly - most cheat sheets should be fine with the defaults (which we can tweak when we wish), but any that want additional triggers should probably get permission first - there has to be a *good* reason for straying from the defaults.

Seem right @zachthompson ?

share/goodie/cheat_sheets/triggers.yaml
GuiltyDolphin: @moollaza There's a `cheats` category.

share/goodie/cheat_sheets/triggers.yaml
GuiltyDolphin: @moollaza It'd actually be more work to start excluding triggers - I think generally people we want to use `startend`, but excluding `start` and `end` may be a pain down the line.

I think the current default is `startend`.

My only *real* worry is with `any` - which doesn't *really* work due to the way the `matched_trigger` is handled in DDG - where it returns the remainder because it doesn't isolate the actual trigger.

zachthompson:
@GuiltyDolphin `matched_trigger` isn't super-robust. I've made a note of it here - https://github.com/duckduckgo/duckduckgo/issues/177

share/goodie/cheat_sheets/triggers.yaml
moollaza: > I think the current default is startend.

My bad, you're correct :+1:

share/goodie/cheat_sheets/triggers.yaml
```diff
@@ -0,0 +1,230 @@
...
+  # Introduces basic translations for a human language.
+  language:
+    startend:
+      - "phrases"
+      - "translations"
+    ignore:
+      - "simple"
+      - "basic"
+      - "example"
+      - "list of"
```
GuiltyDolphin: @moollaza @zachthompson I was going through the aliases on #2382 and noticed virtually all the language cheat sheets had `english to X` as an alias (where `X` is the cheat sheet name). I could easily add that to this then remove the aliases from the cheat sheets, what do you say? (My only concern is if there were some non-English to X cheat sheet - but I don't think we support those?).

GuiltyDolphin:
@moollaza @zachthompson I've gone ahead with moving the `english to` aliases into the `triggers.yaml` file. If you've got any issues please let me know. I'm not sure it covers all the original triggers (`english to X cheat sheet`), but I could perhaps get around this through a global trigger mechanism? It does let you do stuff like `basic english to X phrases` though :dolphin:.

***Note:*** The tests are failing because of a bad cheat sheet - it is fixed in #2281 so it should be fine when that gets merged.

zachthompson:
@GuiltyDolphin Is this ready for review again?

GuiltyDolphin:
@zachthompson Sure - what're your thoughts on the `english to` 'alias'?

At the moment it will trigger with `english to X phrases` (and similar), but not `english to X cheat sheet` as the 'ignore' sections aren't shared between categories.

If I were to add support for shared ignores it would require finding out specifically which category triggered - which might start making the code a bit more complex? But yeah, that's the only thing bugging me at the moment.

lib/DDG/Goodie/CheatSheets.pm
```diff
@@ -46,34 +102,50 @@ sub getAliases {
...
-    open my $fh, $aliases->{join(' ', split /\s+/o, lc($_))} or return;
+    my $remainder = shift;
+
+    my $trigger = join(' ', split /\s+/o, lc($req->matched_trigger));
+    my $lookup = $trigger_lookup{$trigger};
+    if (exists $trigger_lookup{ignore}{$trigger}) {
+        foreach my $ignore (keys %{$trigger_lookup{ignore}{$trigger}}) {
+            $remainder =~ s/\b$ignore\b//;
+        }
+        $remainder =~ s/^\s+//;
```
zachthompson:
```Perl
s/^\s*(.+?)\s*$/$1/;
```

zachthompson:
@GuiltyDolphin How about we delve into the [ignore](https://github.com/GuiltyDolphin/zeroclickinfo-goodies/commit/620f0a051cd4cff8d2dec4a96cf463724af188de) functionality in a follow up PR?  It [affected](https://zach.duckduckgo.com/nytprof/cheatsheet-2170-latest/CheatSheets-pm-8-line.html#116) performance from where we were [previously](https://zach.duckduckgo.com/nytprof/cheatsheets-2170-mod8/CheatSheets-pm-8-line.html#118) and I'm not sure we want to get that complex with the triggers yet.

GuiltyDolphin:
@zachthompson Sure - I'm happy to do that, then we can focus on where we want to go with any additional triggering mechanisms.

GuiltyDolphin:
@zachthompson I've removed all support for the 'ignore' option. I'll make a PR once this is finished for any follow-up/improvements we want to make to the triggering (including 'ignore').

zachthompson:
:+1:  I'm going to try and look at this today.

lib/DDG/Goodie/CheatSheets.pm
```diff
@@ -5,29 +5,74 @@ use JSON::XS;
...
+# trigger.
+sub generate_triggers {
+    my $aliases = @_;
+
+    # Initialize categories
+    my %categories;
+    my $category_map = $trigger_data->{template_map};
+    my %spec_triggers = %{$trigger_data->{categories}};
+    # Initialize custom triggers
+    foreach (my ($id, $spec) = each ($trigger_data->{custom_triggers} || {})) {
```
zachthompson: Should be `while`.

share/goodie/cheat_sheets/triggers.yaml
```diff
@@ -0,0 +1,183 @@
...
+#     startend:
+#       - "equation sheet"
+#       - "formula sheet"
+#
+# And now it will be able to trigger on 'calculus equations',
+# 'calculus formula sheet', 'equation sheet calculus' etc...
+#
+# NOTE: IDs should be in alphabetical order (and the categories and triggers) as
+# much as possible to ensure it is as easy as possible to locate triggers.
+custom_triggers:
```
zachthompson: Let's comment out for now.

zachthompson:
Thank you.

GuiltyDolphin:
@zachthompson Sweet. I'll go sort out my branches for the 'continuation'.

PR #2189, opened 2016-01-31
Interactive Calculator WIP

GuiltyDolphin:
Thanks to @sevastos for a head start - I'm not much of a designer!

This builds upon #1927, so I'll need to re-base onto that when it gets merged... Fun fun fun...

## Current State

- Interfaces with the calculator from #1927, so anything that supports should work out of the box (as long as it is enabled to pass text).
- Pretty decent support for unary functions.
- No support for binary functions.

## To-Do (and musings / workings)

- [ ] Enable switching between output forms (fractions, decimal, (maybe mixed fractions?)).
- [ ] Assume the first unfilled argument/operand is the previous answer (default 0).
- [ ] Output HTML (or Unicode characters?) from Perl to improve styling.
- [ ] Trigger on things such as `calculator`, `interactive calculator` etc...

### Bugs

- [ ] Sometimes the `backspace` functionality is failing for nested functions with multiple values (e.g., `sin(sin(cos(777)))`).

### Plan of Action

1. [ ] Go over the current system and implement a rigid system for traversal that is able to cope with nested scopes as well as multiple fields.
2. [ ] Throw away as much functionality as is not actually needed at this point (in terms of the JS). This should hopefully make working the the code a lot easier.
3. [ ] Improve display to support an arbitrary number of functions, while making it more modular at the same time (the css and handlebars needs refactoring).

### Results

The JS sends requests to the Perl file (through the DDG API - I know! That's a bit *spicy*) which then validates the input and performs the calculations to produce a result.

I'm thinking it should produce HTML as well as plain text for both result and formatted input. Also it should provide each result type (fraction, decimal etc...) in separate fields.

Not sure how much of that would actually be required (might just be the formatted input?) as the Calculator has to run through the text if it is to be put back into the display. My main concern would be for functions like `sin`, which would likely not be processed correctly in the JS (and I don't want to implement anything to test for this - all the non-display hard work should be done in the Perl).
 
### Display

Display tips, code snippets etc are welcome! I'm not super experienced with CSS and styling in general.

- [ ] Tabs

I'm kinda thinking that it'd be good to have several tabs in the display which the user can switch between. These would cover things like trig, possibly things like currencies and bases, as well as a general tab (scientific?) that would be default and cover general use.

- [ ] Indicate current field

Need a way to display which field is active so the user knows where they have to move to get out of functions etc...

I think I might be able to do this by setting a class on the current field when generating the HTML based on the cursor position. Gonna look into this.

---
[IA Page](https://duck.co/ia/view/calculator)


daxtheduck:
Metadata placeholder post
            
---
            
**How to find your IA page**
- New Instant Answer
              
- Existing Instant Answer

GuiltyDolphin:
Continued on #2848.

PR #2196, opened 2016-02-01
Regex: Remove last test temporarily

moollaza:
It's throwing warning on production machines about an uninitialized regex

/cc @jdorweiler @GuiltyDolphin


GuiltyDolphin:
@moollaza Aye... But when I find something that breaks it I have to try and fix it and include the test ;) Even if it wouldn't trigger on production!

PR #2210, opened 2016-02-02
Cheat Sheet : Express brackets using double backslash

vikashvverma:
There were some brackets that were not properly expressed. This PR expresses brackets with double backslash.

IA Page : https://duck.co/ia/view/php_cheat_sheet

### DIFF
--- a/share/goodie/cheat_sheets/json/php.json
+++ b/share/goodie/cheat_sheets/json/php.json
@@ -88,10 +88,10 @@
             "key": "(...)"
         }, {
             "val": "in range (a, b or c)",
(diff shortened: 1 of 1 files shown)


moollaza:
@vikashvverma escaping isn't necessary when the string contains no spaces. It's only used to delimit keys when the string contains space (and hence multiple keys). If a string contains no spaces, it's taken literally

moollaza:
The IA currently renders correctly, so I will close this.
php_cheat_sheet_at_duckduckgo

vikashvverma:
Yeah sure

PR #2219, opened 2016-02-03
Add template_type to gnome cheat sheet

zachthompson:
Noticed when testing today that the gnome cheat sheet doesn't have a `template_type`.


moollaza:
Thanks! This LGTM :+1:

PR #2242, opened 2016-02-06
TimezoneConverter : Fixes #1279

### LINKED ISSUE #1279: TimezoneConverter: Improve triggering
Originally reported on Duck.co: https://duck.co/ideas/idea/6047/time-zone-conversion-triggering

---

"7:00 p.m. ET to PT" doesn't return the IA though "7:00 PM EDT to PDT" does.

Other queries that should trigger the Time Zones conversion but don't are:
- "7:00 pm eastern to pacific"
- "7:00 PM ET to PT"
- "7:00 pm ET in pacific"
- "7:00 pm EDT in pacific"

---

IA Page:  http://duck.co/ia/view/timezone_converter

vadi95:
Added more triggers for pacific and eastern time zones. Fixes #1279

-----------
IA Page: http://duck.co/ia/view/timezone_converter

### DIFF
--- a/lib/DDG/Goodie/TimezoneConverter.pm
+++ b/lib/DDG/Goodie/TimezoneConverter.pm
@@ -12,7 +12,22 @@ use POSIX qw(fmod);
 
 my %timezones = DDG::GoodieRole::Dates::get_timezones();
 
-triggers any => lc for keys %timezones;
+my $dt = DateTime->now();
+my $utc_hour= $dt->hour;
+$dt->set_time_zone( 'America/Los_Angeles' );
+my $pt_hour= $dt->hour;
+my $pt_offset = $utc_hour - $pt_hour;
+if($pt_offset < 0){
+    $pt_offset = $pt_offset + 24;
+}
+my $pacific = "PST";
+my $eastern = "EST";
+if ($pt_offset == 7) {
+    $pacific = "PDT";
+    $eastern = "EDT";
+}
+
+triggers any => lc for (keys %timezones, "PT", "PACIFIC", "ET", "EASTERN");
 
 zci is_cached   => 1;
 zci answer_type => 'timezone_converter';
@@ -33,6 +48,9 @@ sub parse_timezone {
     # Normalize
     $timezone = uc $timezone;
     $timezone =~ s/\s+//g;
+
+    $timezone =~ s/Pacific|^PT$/$pacific/ig;
+    $timezone =~ s/Eastern|^ET$/$eastern/ig;
 
     my ($name, $modifier, $minutes) = $timezone =~ /\A(\w+)(?:([+-]\d+)(?::(\d\d))?)?\z/;
 


lib/DDG/GoodieRole/Dates.pm
```diff
@@ -114,6 +114,7 @@ my %tz_offsets = (
     DFT   => '+0100',
     EASST => '-0500',
     EAST  => '-0600',
+    EASTERN   => '-0400',
```
GuiltyDolphin: @vadi95 Why are you modifying the GoodieRole?

If you want to modify triggers, please do so through the actual [Instant Answer](https://github.com/duckduckgo/zeroclickinfo-goodies/blob/master/lib/DDG/Goodie/TimezoneConverter.pm).

daxtheduck:
## [TimezoneConverter](https://duck.co/ia/view/timezone_converter)

**Description**: convert times between timezones

**Example Query**: [10:00AM MST to PST](https://beta.duckduckgo.com/?q=10:00AM MST to PST), [19:00 UTC to EST, 1am UTC to PST](https://beta.duckduckgo.com/?q=19:00 UTC to EST, 1am UTC to PST)

**Tab Name**: Answer

**Source**:  

*These are the important fields from the IA page.  Please check these for errors or missing information and update the [IA page](https://duck.co/ia/view/timezone_converter)*

---
*This is an automated message which will be updated as changes are made to the [IA page](https://duck.co/ia/view/timezone_converter)*

vadi95:
Hi! @GuiltyDolphin as you can see here triggers depend on the timezone offsets in GoodieRole/Date.pm https://github.com/duckduckgo/zeroclickinfo-goodies/blob/master/lib/DDG/Goodie/TimezoneConverter.pm#L15

lib/DDG/GoodieRole/Dates.pm
```diff
@@ -185,6 +187,7 @@ my %tz_offsets = (
     NZST  => '+1200',
     OMST  => '+0700',
     ORAT  => '-0500',
+    PACIFIC   => '-0700',
```
GuiltyDolphin: This is not a valid TZ abbreviation, nor is `EASTERN`.

lib/DDG/GoodieRole/Dates.pm
```diff
@@ -195,6 +198,7 @@ my %tz_offsets = (
     PMST  => '-0300',
     PONT  => '+1100',
     PST   => '-0800',
+    PT   => '-0700',
```
GuiltyDolphin: We cannot use `PT` as it is a general term for `PST` and `PDT` - it does not have a fixed UTC offset.

lib/DDG/GoodieRole/Dates.pm
```diff
@@ -124,6 +125,7 @@ my %tz_offsets = (
     EGT   => '-0100',
     EIT   => '+0900',
     EST   => '-0500',
+    ET   => '-0400',
```
GuiltyDolphin: We cannot use `ET` as it is a general term for `EST` and `EDT` - it does not have a fixed offset.

GuiltyDolphin:
@vadi95 Given none of the TZ's you specified are valid, I would advise you to add the triggers to the Goodie rather than the GoodieRole. Just use a list in combination with the existing triggers.

vadi95:
@GuiltyDolphin Thanks for pointing that out :). So I'll assume the PST offset for 'PT' and 'pacific' keywords and EST for 'ET' and 'eastern'? Or should we display both offsets?

GuiltyDolphin:
@vadi95 If you're up to it displaying the appropriate offset given the current time of year would be nice.

It should probably say which it used so `... assuming PST` or something similar. An alternative would be to normalize it to the current TZ and display that, so instead of `PT` it would display `PST` or `PDT`.

lib/DDG/Goodie/TimezoneConverter.pm
```diff
@@ -12,6 +12,25 @@ use POSIX qw(fmod);
 
 my %timezones = DDG::GoodieRole::Dates::get_timezones();
 
+my $dt = DateTime->now();
+
+my $utc_hour= $dt->hour;
+$dt->set_time_zone( 'America/Los_Angeles' );
```
mintsoft: Is this attempting to discover if it's daylight sayings time currently?

lib/DDG/Goodie/TimezoneConverter.pm
```diff
@@ -39,6 +58,21 @@ sub parse_timezone {
     # If timezone doesn't exist, skip it
     return unless defined $timezones{$name};
 
+    if($timezone eq "PT" || $timezone eq "PACIFIC"){
```
mintsoft: if these keys have been added to `%timezones` I don't understand why this is needed? Or is it just for display?

vadi95: This is just for display

lib/DDG/Goodie/TimezoneConverter.pm
vadi95: Yeah

mintsoft:
Hmm looking over this, maybe a cleaner option would be to substitute the input? That way we wouldn't need to hack the output either, we can just replace `s/PACIFIC/PST/` or whatever

mintsoft:
@vadi95 Ping! Just checking in, are you still interested in finishing this off?

vadi95:
@mintsoft @GuiltyDolphin I have added a commit. Please check when you're free :)

GuiltyDolphin:
@vadi95 The main issue with this solution is that it does not *extend* to other regions. If we are going ahead with this I think the solution should support all time zones that the Role can process - in which case it might be better if we handled this directly in the Role, allowing the 'generic' form of a TZ to be parsed *into* the ST or DT for a particular region. What we should ***not*** do however, is treat them as time zones - which they are not (truly), they are special cases that can *sometimes* represent a particular time zone - and precedence should always be given to a *real* time zone.

TL;DR: The solution should support all time zones - not just EDT/ET and PST/PT; and thus I think it would be better handled directly within the Dates Role.

moollaza:
@vadi95 thanks again for improving the Timezone Converter Goodie! You’ve been doing a really good job here and we'd like you to help make DuckDuckGo even better for yourself and others interested in Time Zones.

When you're ready to move on, please: 

- Create a new Instant Answer for searches you do related to Time Zones

OR

- Test live Time Zone Instant Answers (or those in development) for bugs and suggest improvements

Detailed instructions explaining how you can help are available at the top of this [page](https://github.com/duckduckgo/duckduckgo/wiki/Creating-or-Improving-Instant-Answers).

Thanks again!

mintsoft:
Ping! @vadi95 Just checking in to see if you're still interested in finishing this up?

mintsoft:
@vadi95 I'm going to close this for now; feel free to gimme a ping to reopen it whenever you want :+1:

PR #2255, opened 2016-02-10
Created IsAwesome Goodie

vCillusion:


moollaza:
@vCillusion thanks a lot for making this. Unfortunately, we're no longer accepting them for PRs though. They're intended to be a tutorial for you to build to see how Goodies work.

Thanks again! I would suggest you build a new Goodie or tackle some bugs!

PR #2262, opened 2016-02-11
Remove Metadata

GuiltyDolphin:
Some metadata was creeping back in...

/cc @moollaza


moollaza:
Thanks! This LGTM :+1:

PR #2274, opened 2016-02-12
Remove Ascii which has been replaced by Bin2Unicode

zachthompson:
@moollaza I think we want to ditch this if we're rolling with [Bin2Unicode](https://github.com/duckduckgo/zeroclickinfo-goodies/pull/2045).


moollaza:
Thanks! This LGTM :+1:

PR #2281, opened 2016-02-13
CheatSheets: Add more tests

### LINKED ISSUE #1844: Cheat Sheets (Controller): Need to ensure aliases don't match existing cheat sheet names, or aliases
CheatSheets.t (or similar) should test that the aliases for each cheat sheet don't match the name of an existing cheat sheet, or the aliases of an existing cheat sheet.

Names and aliases must be unique across cheat sheets.

For example `ubuntu.json` had `"aliases" : [ "ubuntu unity" ]`, while there was already an `ubuntu-unity.json` cheat sheet.

This caused ambiguity in triggering and so the tests failed randomly (depending which Cheat Sheet triggered).

------

GuiltyDolphin:
Adds more tests to the `CheatSheets` Controller.

Also fixes #1844 and cleans up some cheat sheets that were violating tests.

### Tests

* ~~**Warning** if tabs are used for indenting.~~
* ~~**Warning** if an indent that is not a multiple of 4 is used.~~
* ~~**Warning** if trailing white-space is detected.~~
* **Critical** if an alias is used that duplicates the cheat sheet name.
* **Critical** if an alias is used that contains uppercase characters.
* **Critical** if the ID does not follow the expected naming conventions.
* **Critical** if the file name does not follow the expected naming conventions.
* **Critical** if no `template_type` is specified.
* **Dies** if an alias is used that is the name of another cheat sheet.
* **Dies** if an alias is used that is already in use.

---
[IA Page](https://duck.co/ia/view/cheat_sheets)


daxtheduck:
## [Cheat Sheets (Controller)](https://duck.co/ia/view/cheat_sheets)

**Description**: The base controller for all Cheat Sheet Goodies. This is the actual Instant Answer that gets triggered, and decides which Cheat Sheet to display based on the query. This Goodie also handles the parsing of each Cheat Sheet's JSON and the display of the Cheat Sheet.

**Example Query**: [vim cheatsheet](https://beta.duckduckgo.com/?q=vim cheatsheet), [](https://beta.duckduckgo.com/?q=)

**Tab Name**: Answer

**Source**:  

*These are the important fields from the IA page.  Please check these for errors or missing information and update the [IA page](https://duck.co/ia/view/cheat_sheets)*

---
*This is an automated message which will be updated as changes are made to the [IA page](https://duck.co/ia/view/cheat_sheets)*

mintsoft:
@GuiltyDolphin This is conflicted now, although I was thinking another test should be if an aliases is added which is the same as the cheatsheet id!

GuiltyDolphin:
@mintsoft Isn't that one in there? It might be in #2170 if it's not.

GuiltyDolphin:
@zachthompson Any thoughts on the tests? This roughly goes hand-in-hand with #2170 - improving the general process.

lib/DDG/Goodie/CheatSheets.pm
```diff
@@ -38,15 +39,21 @@ sub getAliases {
...
+        my $name = File::Basename::fileparse($file);
+        my $defaultName = $name =~ s/-/ /gr;
         $defaultName =~ s/.json//;
 
         $results{$defaultName} = $file;
 
         if ($data->{'aliases'}) {
             foreach my $alias (@{$data->{'aliases'}}) {
-                $results{lc($alias)} = $file;
+                my $lc_alias = lc $alias;
```
zachthompson: So why does this particular "test" need to be done here every time we load CheatSheets?  Couldn't this be in the CheatSheets.t?

GuiltyDolphin: @zachthompson We're building up the aliases so we already have something to check against - we could probably do the same in `CheatSheets.t` but it would involve duplicating the entire logic AFAIK. There might be a better way though, any ideas?

zachthompson: We're already iterating over the json, extracting the default name, etc.  I don't think too much logic would need to be duplicated.  Most of the logic is this change in fact.

GuiltyDolphin: @zachthompson Whoops, my bad, yeah, I was looking at `CheatSheetsJSON.t` rather than `CheatSheets.t`, looks like I should be able to move it out.

zachthompson: :+1: I think that's the right place to alert people of potential conflicts.

GuiltyDolphin: @zachthompson Alrighty, I've updated it so the tests are now in the `CheatSheets.t` file.

share/goodie/cheat_sheets/json/db2.json
```diff
@@ -1,5 +1,5 @@
 {
-    "id": "ibm_db2_cheat_sheet",
+    "id": "db2_cheat_sheet",
```
zachthompson: Why are we changing this?  We will need to change the id on the IA page.

t/CheatSheets/CheatSheets.t
```diff
@@ -15,20 +15,31 @@ sub getTests {
...
+        my $defaultName = $name =~ s/-/ /gr;
         $defaultName =~ s/.json//;
 
+        $aliases{$defaultName} = $file;
+
         $tests{$defaultName." cheat sheet"} = test_zci(build_answer($data));
 
         if ($data->{'aliases'}) {
             foreach my $alias (@{$data->{'aliases'}}) {
+                my $lc_alias = lc $alias;
```
zachthompson: If this is in the test file, why is this change still here?

t/CheatSheets/CheatSheetsJSON.t
```diff
@@ -55,6 +96,16 @@ foreach my $path (glob("$json_dir/*.json")){
     $temp_pass = (my $url = $json->{metadata}{sourceUrl})? 1 : 0;
     push(@tests, {msg => "sourceUrl is not undef $name", critical => 1, pass => $temp_pass, skip => 1});;
 
+    ### Alias tests ###
+    my $has_aliases = exists $json->{aliases};
```
zachthompson: Put this in the `if` below, e.g. `if(my $aliases = $json->{aliases})`

t/CheatSheets/CheatSheetsJSON.t
```diff
@@ -55,6 +96,16 @@ foreach my $path (glob("$json_dir/*.json")){
...
+    my $has_file_name_alias = 0;
+    if ($has_aliases) {
+        my @aliases = map { lc } @{$json->{aliases}};
+        $has_file_name_alias = scalar (grep { $_ eq $defaultName } @aliases);
```
zachthompson: use `List::Util 'first'`

t/CheatSheets/CheatSheetsJSON.t
```diff
@@ -28,6 +44,16 @@ foreach my $path (glob("$json_dir/*.json")){
     $temp_pass = ($json = decode_json($content))? 1 : 0;
     push(@tests, {msg => 'content is valid JSON', critical => 1, pass => $temp_pass});
 
+    ### Code style tests ###
```
zachthompson: I'm on the fence with these.  *If* we're going to do them, they should give the line numbers on which the warning was generated.

@moolaza You deal with these PRs every day.  Are these formatting warnings useful to you and contributors?

share/goodie/cheat_sheets/json/db2.json
GuiltyDolphin: @zachthompson Yeah, that or the file. `db2.json -> db2_cheat_sheet`. Would you rather I moved the file?

t/CheatSheets/CheatSheetsJSON.t
GuiltyDolphin: @zachthompson I'm getting a bit more fency - having a look at the logs it turns out there is a lot of violation, so it is a bit harder to find the *real* issues.

If we had them all cleaned up, then it might prevent future offenses, (and it encourages people to follow the preferred style should they move on to a full Goodie), but I wouldn't have too much issue with removing them.

I could do line numbers (it might be a bit slower), but what happens with multiple offenses? First occurrence, or do we spam them with line numbers?

mintsoft: if we're moving to content managed here, then this'll become redundant anyway. Personally I'm less bothered about the JSON 'style' as it were

GuiltyDolphin: @mintsoft @zachthompson Enough for me, I'll remove the style tests.

t/CheatSheets/CheatSheetsJSON.t
```diff
@@ -39,6 +55,21 @@ foreach my $path (glob("$json_dir/*.json")){
     $temp_pass = (!exists $json->{description} && !$json->{description})? 0 : 1;
     push(@tests, {msg => "has description (optional but suggested)", critical => 0, pass => $temp_pass});
 
+    ### ID tests ###
+    my $cheat_id = $json->{id};
+    if ($cheat_id) {
+        my $expected = file_name_to_id $file_name;
+        $temp_pass = $cheat_id eq $expected;
+        push(@tests, {msg => "bad id '$cheat_id', should be '$expected'", critical => 1, pass => $temp_pass});
+
```
zachthompson: These two are effectively the same test.  Both either pass or fail.  Since we don't really know which value is incorrect, the file name or the id, let's do one test.   We should alert that the id and file name don't match with a generic example.

t/CheatSheets/CheatSheetsJSON.t
```diff
@@ -55,6 +86,17 @@ foreach my $path (glob("$json_dir/*.json")){
     $temp_pass = (my $url = $json->{metadata}{sourceUrl})? 1 : 0;
     push(@tests, {msg => "sourceUrl is not undef $name", critical => 1, pass => $temp_pass, skip => 1});;
 
+    ### Alias tests ###
+    if (my $aliases = $json->{aliases}) {
+        my @aliases = @{$aliases};
+        if (first { lc $_ ne $_ } @aliases) {
```
zachthompson: I think `/[:upper:]/` would be more explicit, and slightly faster.

t/CheatSheets/CheatSheetsJSON.t
GuiltyDolphin: @zachthompson Yeah... I wasn't really sure which one to do seeing as the we don't know at this point whether the ID is correct of the file name.

We could have something along the lines of `Invalid ID 'ID' for file name 'file name'` or similar?

mintsoft: That sounds reasonable

zachthompson: Sounds good. One consideration is to slant the message away from the ID, e.g. not "Invalid ID".  What we're trying to avoid is people changing the IA page ID unnecessarily.  "File name doesn't match ID" type language would be preferable if we don't want to be completely neutral.

GuiltyDolphin: Yeah - that language seems about right.

t/CheatSheets/CheatSheetsJSON.t
```diff
@@ -55,6 +86,17 @@ foreach my $path (glob("$json_dir/*.json")){
     $temp_pass = (my $url = $json->{metadata}{sourceUrl})? 1 : 0;
     push(@tests, {msg => "sourceUrl is not undef $name", critical => 1, pass => $temp_pass, skip => 1});;
 
+    ### Alias tests ###
+    if (my $aliases = $json->{aliases}) {
+        my @aliases = @{$aliases};
+        if ("@aliases" =~ /[[:upper:]]/) {
+            push(@tests, {msg => "uppercase detected in alias - aliases should be lowercase"});
```
GuiltyDolphin: @zachthompson @mintsoft Do you think I should get rid of the bit after the `-`?

zachthompson: Well, it seems like it should be obvious what needs to be done but....I could see someone saying, "Yeah, so?" :smiling_imp:

GuiltyDolphin: @zachthompson Hehe... Well, I guess I'll leave it in for now :wink:

t/CheatSheets/CheatSheetsJSON.t
```diff
@@ -10,15 +10,32 @@ use Test::More;
...
 my $json;
 
+sub file_name_to_id {
+    my $file_name = shift;
+    $file_name =~ s/\.json//;
+    $file_name =~ s/-/_/g;
+    return $file_name . "_cheat_sheet";
+}
+
+sub id_to_file_name {
```
zachthompson: Don't think we need this any longer.

t/CheatSheets/CheatSheetsJSON.t
```diff
@@ -55,6 +80,17 @@ foreach my $path (glob("$json_dir/*.json")){
     $temp_pass = (my $url = $json->{metadata}{sourceUrl})? 1 : 0;
     push(@tests, {msg => "sourceUrl is not undef $name", critical => 1, pass => $temp_pass, skip => 1});;
 
+    ### Alias tests ###
+    if (my $aliases = $json->{aliases}) {
```
zachthompson: Let's make both of these fail and I think we're good.  Anything else you're planning on updating?

mintsoft:
@GuiltyDolphin FYI you're conflicted here

GuiltyDolphin:
@mintsoft Thanks for the heads up!

@zachthompson I think that's all - these were the main issues that were cropping up often; if I've missed anything then its not too much trouble for someone to add later on.

zachthompson:
Thank you.

PR #2296, opened 2016-02-15
Bitsum: Fixes #1902

### LINKED ISSUE #1902: Bitsum: Hamming weight result is confusing
@kste 

The Hamming weight output is the same as the search query which doesn't seem correct. E.g. the [Hamming weight for 11101 is 4](https://en.wikipedia.org/wiki/Hamming_weight) but the IA says 11101.

------
IA Page:  http://duck.co/ia/view/bitsum

swapagarwal:
Show the base used to compute the answer.

Fixes #1902

IA Page: http://duck.co/ia/view/bitsum

### DIFF
--- a/lib/DDG/Goodie/Bitsum.pm
+++ b/lib/DDG/Goodie/Bitsum.pm
@@ -24,28 +24,32 @@ handle remainder => sub {
     my $hex_number;
     my $binstring;
     my $base;
+    my $base_string;
 
     # Construct binary representation for both hex and decimal numbers
     if( $input_number =~ /^0x/) {
         $hex_number = substr($input_number, 2);
         $binstring = unpack ('B*', pack ('H*',$hex_number));
-	$base = 'hexadecimal';
+        $base = 16;
+        $base_string = 'Hexadecimal';
     } elsif( $input_number =~ /^0b/) {
         $binstring = substr($input_number, 2);
-	$base = 'binary';
+        $base = 2;
+        $base_string = 'Binary';
     } else {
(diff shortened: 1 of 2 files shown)


daxtheduck:
Metadata placeholder post
            
---
            
**How to find your IA page**
- New Instant Answer
              
- Existing Instant Answer

lib/DDG/Goodie/Bitsum.pm
```diff
@@ -23,20 +23,25 @@ handle remainder => sub {
...
+	$base = 'binary';
     } else {
         $binstring = Math::BigInt->new($input_number)->as_bin();
+	$base = 'decimal';
     }
     
     # Count ones
     my $result = $binstring =~ tr/1/1/;
-    
+    $input_number .= ' (input treated as '.$base.')';
```
moollaza: I think we should change the result to `(Base 10, Decimal)`.

It would look like:

>**9**
> Hamming Weight Calculation: 11101 (Base 10, Decimal)

/cc @abeyang -- any thoughts? We want to make it more clear what base the input has been interpreted as. I.e. 11101 in base 10 is `0010101101011101` in binary which has a bitsum of 9. It's possible to assume we've taken `11101` as binary and we're showing an incorrect result (because it would be 4, not 9)

swapagarwal: Yep, that looks better! On it.

moollaza:
@swapagarwal this is looking much better :+1:

moollaza:
bitsum_11101_at_duckduckgo

PR #2379, opened 2016-02-20
Conversions: Tests for weird 999 problem

mintsoft:
Currently on production this answer is wrong:

On my older version of `Math::BigFloat` it had the same behaviour, however updating the dependency fixes it. It's probably worth having these tests in anyway


GuiltyDolphin:
Thanks! :dolphin:

mintsoft:
Cheers @GuiltyDolphin

PR #2405, opened 2016-02-21
New Star Trek Code Phrase

swapagarwal:
A Star Trek catchphrase

IA Page: https://duck.co/ia/view/beam_me_up_scotty

### DIFF
--- a/lib/DDG/Goodie/BeamMeUpScotty.pm
+++ b/lib/DDG/Goodie/BeamMeUpScotty.pm
@@ -1,0 +1,25 @@
+package DDG::Goodie::BeamMeUpScotty;
+# ABSTRACT: A Star Trek catchphrase
+
+use DDG::Goodie;
+use strict;
+
+zci answer_type => 'beam_me_up_scotty';
+zci is_cached => 1;
+
+triggers start => 'beam me up scotty', 'beam us up scotty';
+
+handle remainder => sub {
+    return unless ($_ eq '');
+
+    my $answer = 'Aye, aye, captain.';
+
+    return $answer,
+      structured_answer => {
+        input     => ['Beam me up, Scotty'],
+        operation => 'Code phrase',
+        result    => $answer
+      };
+};
+
+1;
--- a/lib/DDG/Goodie/BeamMeUpScotty.pm
+++ b/lib/DDG/Goodie/BeamMeUpScotty.pm
@@ -10,7 +10,7 @@ zci is_cached => 1;
 triggers start => 'beam me up scotty', 'beam us up scotty';
 
 handle remainder => sub {
-    return unless ($_ eq '');
+    return unless $_ eq '';
 
     my $answer = 'Aye, aye, captain.';
 
--- a/t/BeamMeUpScotty.t
+++ b/t/BeamMeUpScotty.t
@@ -1,0 +1,33 @@
+#!/usr/bin/env perl
+
+use strict;
+use warnings;
+use Test::More;
+use DDG::Test::Goodie;
+
+zci answer_type => "beam_me_up_scotty";
+zci is_cached   => 1;
+
+ddg_goodie_test(
+    [qw( DDG::Goodie::BeamMeUpScotty )],
+    'beam me up scotty' => test_zci(
+        'Aye, aye, captain.',
+        structured_answer => {
+            input     => ['Beam me up, Scotty'],
+            operation => 'Code phrase',
+            result    => 'Aye, aye, captain.'
+        }
+    ),
+    'beam us up scotty' => test_zci(
+        'Aye, aye, captain.',
+        structured_answer => {
+            input     => ['Beam me up, Scotty'],
+            operation => 'Code phrase',
+            result    => 'Aye, aye, captain.'
+        }
+    ),
+    'beam me up' => undef,
+    'scotty beam us up' => undef,
+);
+
+done_testing;


daxtheduck:
## [Beam Me Up Scotty](https://duck.co/ia/view/beam_me_up_scotty)

**Description**: Beam me up, Scotty is a catchphrase that made its way into popular culture from the science fiction television series Star Trek. It comes from the command Captain Kirk gives his chief engineer, Montgomery Scotty Scott, when he needs to be transported back to the Starship Enterprise.

**Example Query**: [Beam me up](https://beta.duckduckgo.com/?q=Beam me up), [Scotty](https://beta.duckduckgo.com/?q=Scotty)

**Tab Name**: Answer

**Source**:  

*These are the important fields from the IA page.  Please check these for errors or missing information and update the [IA page](https://duck.co/ia/view/beam_me_up_scotty)*

---
*This is an automated message which will be updated as changes are made to the [IA page](https://duck.co/ia/view/beam_me_up_scotty)*

lib/DDG/Goodie/BeamMeUpScotty.pm
```diff
@@ -0,0 +1,25 @@
...
+use DDG::Goodie;
+use strict;
+
+zci answer_type => 'beam_me_up_scotty';
+zci is_cached => 1;
+
+triggers start => 'beam me up scotty', 'beam us up scotty';
+
+handle remainder => sub {
+    return unless ($_ eq '');
```
moollaza: FWIW we don't typically use the `()` in this context:

e.g.
```perl
return unless $_ eq '';
x = 1 if y == 2;
```

GuiltyDolphin:
@swapagarwal Can we get this to support more phrases? Goodies should ideally be answering multiple queries (more than 10, at least).

I'm personally not a huge fan of (static) short-phrase 'easter-egg' Goodies, but I like the idea of a Star-Trek phrases Goodie. Would it be possible to expand this to cover queries such as star-dates or maybe include references to seasons/authors? You could look around on places like [cpan](https://metacpan.org/search?q=star%20trek) for inspiration and packages that might help you achieve bigger things with this Goodie.

moollaza:
>I'm personally not a huge fan of (static) short-phrase 'easter-egg' Goodies,

@GuiltyDolphin I was contemplating this yesterday: Should we have an "easter eggs" Goodie that uses a YAML file to map all these single (or very niche) queries to an answers. Implementation wise that approach makes the most sense because we have several Goodies that do this now.

However, I think the major benefit to the individual approach is that we have an IA page for each which is searchable and shareable, we also get to credit each "IA" to a developer and assign a maintainer. A single IA page for all easter eggs wouldn't allow us to do this and we wouldn't be able to list out all the example queries either (depending how big it grows).

We could also handle this in some hybrid approach like CheatSheets but that would require internal changes.

For now I think the current approach is acceptable, though I agree these IAs don't really follow protocol.

@zachthompson what are your thoughts on this?

GuiltyDolphin:
@moollaza Yeah, I agree people should be able to contribute in whichever way - but from a development stand-point it makes it more difficult to find other IAs as well as filling-up and slowing-down the tests.

TBH I think your first suggestion (the single Goodie) is a pretty decent approach - people would still receive credit through the `contributers` on duck.co.

Seeing as we're working on the CheatSheets atm, maybe the approach could be similar to however that gets handled? I don't think they really warrant a full Perl-backed approach (overkill!).

swapagarwal:
@GuiltyDolphin I like the idea of star-dates; should that be included in this goodie itself? (Is there a goodie that returns static answers for some queries and calculates for others?)
@moollaza I'm with @GuiltyDolphin on making an easter eggs cheatsheet kind of goodie that would have to be handled internally in a different way.

GuiltyDolphin:
@swapagarwal If you wanted to go down the star-dates route (just for returning random star-dates), I think that in itself would be a Goodie.

The [Caesar Cipher](https://duck.co/ia/view/caesar_cipher) Goodie returns static answers for certain queries, but performs the calculations for the majority. But the results it returns are all for the same specific topic - we should keep 'unrelated' results separate.

swapagarwal:
@GuiltyDolphin Well, star-dates need not be random! It can be calculated for the current time or user specified time. Wait, should we put it in a new goodie or add it as a timezone itself? :stuck_out_tongue_winking_eye:

GuiltyDolphin:
@swapagarwal Huuurh? I think that would be pretty awesome - I was just looking at the cpan page I linked to and the was a random generator.

Calculating star-dates based on current time is much more Goodie! Please feel free to create an IA page and *make it so!*

moollaza:
@swapagarwal @GuiltyDolphin it sounds/looks like this IA is done then?

I'm happy to merge if you are :+1:

swapagarwal:
@GuiltyDolphin I'm working on the [stardate goodie](https://github.com/duckduckgo/zeroclickinfo-goodies/pull/2711). Do have a look!
@moollaza I'll be happy to. I'll take a look and get back to you on slack.

moollaza:
@swapagarwal Thanks! This LGTM :+1:

Would you be willing to update it to the new structured_answer format now?

I will merge this for now, but we'd appreciate if you could update it (and others if interested!) as we're on a missing to replace this old approach to templates. Please see #2767 for more details and instructions on what to do!

swapagarwal:
@moollaza Sure, the instructions made it quite easy! #2816

PR #2515, opened 2016-02-23
Wp-cli cheat sheet: Added one theme command

mayank7:
Added one theme command to wp-cli cheat sheet.

IA Page: https://duck.co/ia/view/wp_cli_cheat_sheet


mayank7:
Reference: http://wp-cli.org/commands/theme/status/

PR #2001, opened 2016-01-03
New WhatIs GoodieRole

GuiltyDolphin:
**Note! There's POD Documentation!**
In the cloned branch, just go to the `zeroclickinfo-goodies` directory, then type `perldoc lib/DDG/GoodieRole/WhatIs.pm` and enjoy!

## _WhatIs_ it?

A GoodieRole that aims to abstract matching certain forms of query so developers can focus on writing the functionality.

### What sort of Goodies might this benefit?

~~Any Goodies that want to be able to match queries that can be boiled down to "What is X in Y?".~~

~~I may wish to add to this, for queries such as `What is ACTION X?`, `What is X of Y?` etc...~~

**_Most if not all!_**

Some Goodies that would directly benefit from this:
 - `PigLatin` - Translations to and from 'Pig Latin'.
 - `LeetSpeak` - Translations to, ehem, 'Leet Speak'.
 - `EmToPx` - Conversions between `px` and `em`.
 - `Ascii` - Conversions between binary and ASCII.
 - `Lowercase` - Produces a lower-case version of its input.
 - `PrimeFactors` - Returns the prime factors of a number.
 - Many others!

### Usage

Importing is just like with any other GoodieRole:

``` perl
use DDG::Goodie;
with 'DDG::GoodieRole::WhatIs';
```

Then you use the `wi_translation` routine to instantiate a matcher:

``` perl
my $bin_conv = wi_translation(
    groups => ['written'],
    options => {
        to => 'Binary',
        primary => qr/\d+/,
    },
);
```

As you see, the `wi_translation` routine takes a hash as an argument, there are several attributes that can be specified which are covered in the module.

After having generated `$bin_conv`, you can use it to match simple queries:

``` perl
my $result = $bin_conv->match('What is 71 in Binary?');
$result->{primary};
# 71

my $result2 = $bin_conv->match('How do I write 42 in Binary?');
$result2->{primary};
# 42
```

Easy peasy!

## To-do

*There's always a to-do.*

### Features

- [x] Support `translation` forms, e.g. `translate X to Y`.
- [x] Support `conversion` forms, e.g., `convert X to Y`.
- [ ] Support `from` forms - e.g., `translate X from Y`.
- [x] Provide a `full_match` option, for cases when the entire query should be a `what is...`.
- [x] Include user named-captures in the result (So you could match directly on value without needing to extract it).
- [ ] Support verb specification - `How do I VERB X in Y`.

### Ideas

[...]

PR #2644, opened 2016-02-27
New WhatIs GoodieRole

GuiltyDolphin:
**Note! There's POD Documentation!**
In the cloned branch, just go to the `zeroclickinfo-goodies` directory, then type `perldoc lib/DDG/GoodieRole/WhatIs.pm` and enjoy!

**Another Note!** This is a continuation of #2001 - but working off a more public branch so collaboration and/or trying it out is easier.

## _WhatIs_ it?

A GoodieRole that aims to abstract matching certain forms of query so developers can focus on writing the functionality.

### What sort of Goodies might this benefit?

**_Most if not all!_** 

Check out https://github.com/duckduckgo/zeroclickinfo-goodies/tree/gd/whatis-conversions for examples.

Some Goodies that would directly benefit from this:
 - `PigLatin` - Translations to and from 'Pig Latin'.
 - `LeetSpeak` - Translations to, ehem, 'Leet Speak'.
 - `EmToPx` - Conversions between `px` and `em`.
 - `Ascii` - Conversions between binary and ASCII.
 - `Lowercase` - Produces a lower-case version of its input.
 - `PrimeFactors` - Returns the prime factors of a number.
 - Many others!

### Usage

Importing is just like with any other GoodieRole:

``` perl
use DDG::Goodie;
with 'DDG::GoodieRole::WhatIs';
```

Then you use the `wi_translation` routine to instantiate a matcher:

``` perl
my $bin_conv = wi_translation(
    groups => ['written'],
    options => {
        to => 'Binary',
        primary => qr/\d+/,
    },
);
```

As you see, the `wi_translation` routine takes a hash as an argument, there are several attributes that can be specified which are covered in the module.

After having generated `$bin_conv`, you can use it to match simple queries:

``` perl
my $result = $bin_conv->match('What is 71 in Binary?');
$result->{primary};
# 71

my $result2 = $bin_conv->match('How do I write 42 in Binary?');
$result2->{primary};
# 42
```

Easy peasy!

## To-do

*There's always a to-do.*

### Issues

- [x] Kindah a biggie... It does slow down the tests as there is no caching - at the moment it adds in the order of 0.2~0.4 seconds (on my machine - which is quite slow anyway) to any Goodie using it, which is probably a bit *too* much over hundreds of Goodies. ***Priority!***
- [x] Improve handling of `translation` groups - currently `what is conversion` is always triggered, but requires the `to` option to be set.

#### Modifiers

- [x] Support `translation` forms, e.g. `translate X to Y`.
- [x] Support `conversion` forms, e.g., `convert X to Y`.
- [x] Support `from` forms - e.g., `translate X from Y`.
- [x] Support verb specification - `How do I VERB X in Y`.
- [x] Perhaps split the `property` group up in terms of singular and/or plural. I don't think we can assume that there is always a _plural_ version for the property - but it might still make sense to say "What is the X of Y?".
- [ ] Support stand-alone primaries (i.e., where `primary` can be used on its own).
- [x] I think the `imperative` group should be renamed to `command` - much more _obvious_ what it means then.
- [ ] We could have a `basic math` modifier. This would support very basic 'mathy' expressions such as `X plus Y`, `X - Y` etc... Obviously it would not support a wide range of 'operators', but could be useful for Goodies like `DateMath` (if I see more examples I'll look into this further).
- [ ] Better modifier/entry customization. I think there's a couple of routes for this: 
    1. Build more optional-options into the modifiers. This would allow greater customization through just specifying more options.
    2. Build an inheritance system for the modifiers - this would perhaps allow a greater number of 'smaller' modifiers which could each be a variant of an existing modifier. There are already some modifiers that are 'sub-modifiers': for example `spoken translation` and `written translation`. This might also make modifiers easier to create in general as the majority of the implementation would already be defined.

#### Matcher

- [x] Provide a `full_match` option, for cases when the entire query should be a `what is...`.
- [x] Include user named-captures in the result (So you could match directly on value without needing to extract it).
- [ ] Have helper regex functions for use with the entry definitions - e.g., for quoting primary it might be `primary => wi_re_quotes` or similar; or alternatively we could maybe expose some of the `Expression` interface? An alternative would be to fully embed these in groups (e.g., `wi_re_quotes` could become the group `quoted`), this would be a bit less powerful (probably) as you'd have to guess _what_ should be quoted.

#### Misc

- [ ] Provide integration with other GoodieRoles - this may include things like using the Dates GoodieRole for date parsing, NumberStyle for parsing numbers etc...

#### POD

- [ ] Give a more robust definition of 'results' and ensure there is high/full coverage.
    - [ ] Perhaps mention that user matches can be embedded in the results?

### Ideas

- [x] Could make it more general by providing access-points through the Role which use presets to generate good defaults for various query-styles.
- [ ] I'm thinking I need to rethink the way modifiers (not expressions) are handled. Currently it does not feel *super* extensible; you have to define each modifier from scratch. I think the main options would be either modifier inheritance or changing the meaning of options/groups such that options control the minutiae of parsing, but the groups can change the way in which options are *interpreted*. Of course I could try a mix of both :wink:


daxtheduck:
Metadata placeholder post
            
---
            
**How to find your IA page**
- New Instant Answer
              
- Existing Instant Answer

GuiltyDolphin:
@duckduckgo/community-leaders Can anyone confirm the (testing) speed decrease/get some decent metrics? It doesn't seem to be doing it so much atm...

lib/DDG/GoodieRole/WhatIs.pm
```diff
@@ -0,0 +1,433 @@
...
+};
+
+# Matcher with no defaults.
+_build_wi wi_custom => sub { @_ };
+
+1;
+
+__END__
+
+=head1 NAME
```
mintsoft: Hmm, we don't usually do POD; usually the role docs go in: http://docs.duckduckhack.com/backend-reference/goodie-helpers.html although I'm wondering about their visibility there

GuiltyDolphin: @mintsoft Yeah, I've discussed this a bit over on `duckduckhack-docs`, I think the consensus was that it's fine to have the POD, but the docs should probably have something too.

TBH I constantly use POD/`perldoc`, I'm not 100% sure what the situation is for others but if it's anything like for me then this is *way* more convenient.

mintsoft: Yeah I like POD in general; I think the actual problem is visibility/discoverability of roles

GuiltyDolphin: @mintsoft Yeah, I think we need to promote them a bit more - they are *very* useful!

The docs should definitely at least *introduce* each of the Roles - but I think promotion in the Goodie development parts would go a long way.

mintsoft:
Run on master: 
```
Files=288, Tests=4048, 26 wallclock secs ( 4.26 usr  0.69 sys + 175.89 cusr  7.44 csys = 188.28 CPU)
Result: PASS

real    0m26.491s
user    3m0.320s
sys     0m8.136s
```

Run on gd/whatis-conversions at `e217472dda45ce0e71baa315402af9f1af79fe62`:
```
Files=288, Tests=4046, 29 wallclock secs ( 4.34 usr  0.66 sys + 184.13 cusr  7.43 csys = 196.56 CPU)
Result: PASS

real    0m29.073s
user    3m8.640s
sys     0m8.100s
```

GuiltyDolphin:
@mintsoft Thanks! I'll keep the speed in mind as I'm working on improving the general design.

t/00-roles.t
```diff
@@ -613,4 +613,346 @@ subtest 'ImageLoader' => sub {
...
+    add_option_queries 'language translation' =>
+        { direction => 'to' }, (
+        'translate hello to Goatee' => 'hello',
+    );
+    add_option_queries 'language translation from' =>
+        { direction => 'from' }, (
+        'translate hello from Gribble' => 'hello',
+    );
+
+    sub hash_tester {
```
mintsoft: Can we move helpers etc out of the way of the main body of tests?

t/00-roles.t
```diff
@@ -613,4 +613,346 @@ subtest 'ImageLoader' => sub {
...
+                    };
+                    %invalid_queries = (%invalid_queries, %to_add);
+                } else {
+                    %invalid_queries = (%invalid_queries, %{$wi_valid_queries{$invalid}});
+                };
+            };
+            subtest 'Invalid Queries' => build_value_test($wi, 0, %invalid_queries);
+        };
+    }
+    sub test_custom { modifier_test(\&wi_with_test)->(@_) };
```
mintsoft: ?

mintsoft:
I'm also getting 
`push on reference is experimental at /media/md2/scripts/ddg/ddh/goodies/master/lib/DDG/GoodieRole/WhatIs.pm line 25.` on my perl(v5.20.2)

GuiltyDolphin:
@mintsoft Aghhh! I forgot about that! Yeah, thanks. Need to remove the experimental stuff.

mintsoft:
at `81112f46f4fe27aa9855f99d54a04087ba6a8f1e`:
```
All tests successful.
Files=288, Tests=4049, 27 wallclock secs ( 4.40 usr  0.70 sys + 177.42 cusr  7.54 csys = 190.06 CPU)
Result: PASS

real    0m27.850s
user    3m1.980s
sys     0m8.252s
```

GuiltyDolphin:
@mintsoft Sweet! I checked it out on mine and it turned out to be a couple of seconds faster :wink: I think that's the speed issue sorted (at least for now).

GuiltyDolphin:
@duckduckgo/community-leaders I am super happy to say - this is finally done!

Of course there's a few things that might want tweaking in the future, but as far as users and the interface are concerned, I am pretty happy for it to stay static!

Here's the documentation in case you wanted to read it: [wi.pdf](https://github.com/duckduckgo/zeroclickinfo-goodies/files/264535/wi.pdf)

So a quick overview of the interface:

* The main interaction is through 'Matchers', where you define what you want to match and perform matches.
* You customize your Matcher through the following:
  * Groups, which in turn enable Modifiers - these determine the general forms the Matcher will match (e.g., `how do I DO X in Y`).
  * Options - allow fine-tuning of the forms matched by the Modifiers - for example `How do I write 7 in binary` (`to => 'binary'`).
  * Option Modifiers - allow *even* more fine-tuning of the results by specifying, for example, that the match should be numeric.

Then you use `$modifier->full_match($query)` to grab a hash of results!

---

I've put a lot of work into making this robust with full documentation and test coverage; of course we can tweak later, it is only the interface that mustn't change!

@zekiel You spoke about expanding Instant Answer coverage of the query space - now we can cover *loads* more queries with *very* little effort!

If anyone has any last comments/suggestions please share! But from my POV this is ready to ship and I'm looking forward to seeing what people make of it!

@talsraviv Documentation shouldn't be an issue - I've written the docs for this in POD and `Pod::Markdown` does an *awesome* job of converting to markdown (see the attached PDF above).

@mintsoft Thanks for giving me feedback through the development process - it's been really useful!

---

@duckduckgo/community-leaders I'm not asking for a review of the internals - though feel free, it is the interface that is important here!

Also, if you want to check out some usage of this, check out [gd/whatis-conversions](https://github.com/duckduckgo/zeroclickinfo-goodies/tree/gd/whatis-conversions) where I've updated many Goodies to use the Role (though I think allowing people to convert in their own time would be great too).

---

I have some pretty big plans for this in the future, such as integration with the Dates and NumberStyle Roles, allowing people to specify matches should be a date or number (they can do this already, but it doesn't have localization) and so forth.

---

:dolphin: ***Thank you!*** :dolphin:

gautamkrishnar:
@GuiltyDolphin  Good job :cake:  :+1:

moollaza:
@GuiltyDolphin thanks a lot for this.

I've discussed this with @zachthompson and we agreed that this is a low priority for us right now. Merging this requires a lot more than a code review and performance test. Internally the team has been working to understand our triggering system (and its problems) better and we know that in the long term, word-triggers are unlikely to the ideal way of triggering Instant Answers. They're easy for developers to understand, but typically trigger on irrelevant queries and miss many relevant queries that should produce an Instant Answer.

Until we know the direction we're heading, we should avoid adding any new features to the triggering system. If there are problems that need to be fixed, we need to discuss them first, plan a solution and agree on the implementation.

GuiltyDolphin:
@moollaza Yo, sure! Let me know if, and when, you want to go somewhere with this - I'll be happy to discuss internals; it was intended more as a supplementary tool than an actual change to the system, but I guess it would have those impacts ^.^

moollaza:
> Yo, sure! Let me know if, and when, you want to go somewhere with this

Thanks a lot for understanding! For now I'm going to close this because I don't expect us to merge it. But keep the code branches around so we can revive this if necessary

PR #2652, opened 2016-02-28
support "dices" spelling for dice trigger

### LINKED ISSUE #1255: Dice: Support for "Roll 2 dices" or "Roll two dice"
On the instant answer "dice":

Keyword: "roll 2 dice" works, although I'm seeing blank rectangles but that's supposedly my browser unicode support that is lacking (latest chrome).

image

I would have supported ***"roll 2 dices"*** too. Dices is an accepted word of the [dictionary](http://www.thefreedictionary.com/dices), albeit less common than dice.

***"Roll two dice"*** doesn't work either.

-----
IA Page: https://duck.co/ia/view/dice

marianosimone:
fixes #1255

I didn't include the part about converting "two" into 2, as I wouldn't know where to stop (one, two, three, ten, twenty???)

### DIFF
--- a/lib/DDG/Goodie/Dice.pm
+++ b/lib/DDG/Goodie/Dice.pm
@@ -4,6 +4,7 @@ package DDG::Goodie::Dice;
 
 use strict;
 use DDG::Goodie;
+use Lingua::EN::Numericalize;
 
 triggers start => "roll", "throw";
 
@@ -34,6 +35,9 @@ sub set_num_dice {
     my $num_dice = $_[0];
     my $num_dice_default = $_[1];
     if(defined($num_dice)){
+        if ($num_dice =~ /^[a-zA-Z]+$/) {
+            return str2nbr($num_dice);
+        }
         if ($num_dice ne ''){
             return $num_dice;
         }else{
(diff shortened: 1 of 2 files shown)


daxtheduck:
Metadata placeholder post
            
---
            
**How to find your IA page**
- New Instant Answer
              
- Existing Instant Answer

GuiltyDolphin:
@marianosimone `dices`? I don't even think that's a word. There's `die` and `dice`.

mintsoft:
@GuiltyDolphin True, but I imagine that people will search for it

GuiltyDolphin:
@mintsoft Fair dos. I'm not sure how well it will play with Numericalize; @marianosimone you might want to isolate the number-words from the rest of the query so Numericalize doesn't mess it up. It might be fine though seeing as it will just strip the extra 's'.

marianosimone:
@GuiltyDolphin @mintsoft thanks for pointing out `Numericalize`. Already integrated!

BTW, as #1255 indicates, `dices` is an accepted form (according to http://www.thefreedictionary.com/dices)

GuiltyDolphin:
@marianosimone Yeah, it's fine.

Can we add a test or two showing the combination of word-numbers with `dices`?

GuiltyDolphin:
@marianosimone Just a couple using the words (like `three`, `seven` etc...) along with `dices`, to make sure they play nicely.

t/Dice.t
```diff
@@ -91,6 +105,25 @@ ddg_goodie_test(
             }
        }
     ),
+
+    # Query with numbers as words
+    "roll five dice" => test_zci(
```
marianosimone: @GuiltyDolphin you mean this?

GuiltyDolphin: No, I mean combining both the things you are adding - `dices` and the number-word triggers. Like `roll seven dices`.

marianosimone:
@GuiltyDolphin added!

lib/DDG/Goodie/Dice.pm
```diff
@@ -76,12 +80,16 @@ handle remainder_lc => sub {
     my $heading = "Random Dice Roll";
     my $total; # total of all dice rolls
     foreach (@values) {
-        if ($_ =~ /^(?:a? ?die|(\d{0,2})\s*dic?e)$/) {
-            # ex. 'a die', '2 dice', '5dice'
+        if ($_ =~ /^(?:a? ?die|(\d{0,2}|[a-zA-Z]+)\s*dic?es?)$/) {
```
GuiltyDolphin: This won't catch all numbers - there are also phrases like `twenty-seven` and `three hundred and fifty-four`. We can probably just pass the stripped-down version off to `Numericalize`.

marianosimone: Mhhh... I'm only using this to catch simple cases (there's a limit on how many dice we accept anyway, which is 99).

I haven't thought about the '-' case though, so maybe supporting `[\s\-]` is a good idea. Adding now!

GuiltyDolphin:
@marianosimone Thanks, this looks great! :dolphin:

PR #2667, opened 2016-02-29
New  GoLang Tools Cheat Sheet

vikashvverma:
IA Page: https://duck.co/ia/view/golang_tools_cheat_sheet

PR #2681, opened 2016-03-01
CheatSheets: Improve file name ID check

### LINKED ISSUE #2685: Add tests for custom cheat sheet triggers
@GuiltyDolphin The cheat sheet tests should check for the following (at least) in custom_triggers:

1. Two cheat sheets using the same trigger.  Currently just silently ignores all but the first.
2. That categories already included by template_type aren't re-added via additional_categories

These should be fatal.  Any others you can think of.  We don't have any going out the door so I didn't think it should hold up the initial PR.

GuiltyDolphin:
@moollaza I've added in the reverse check to make sure weird cases don't make it through.

/cc @zachthompson 

Fixes #2685.

Adds some new tests:

* ***Critical*** if a trigger is specified that is already in use.
* ***Critical*** if a category is specified that is already assigned via the `template_type`.
* ***Critical*** if a `template_type` is used that is not present in the `template_map`.

---

### DIFF
--- a/t/CheatSheets/CheatSheetsJSON.t
+++ b/t/CheatSheets/CheatSheetsJSON.t
@@ -11,9 +11,28 @@ use Term::ANSIColor;
 use JSON;
 use IO::All;
 use List::Util qw(first);
+use YAML::XS qw(LoadFile);
 
 my $json_dir = "share/goodie/cheat_sheets/json";
 my $json;
+
+my $triggers_yaml = LoadFile('share/goodie/cheat_sheets/triggers.yaml');
+
+my %triggers;
+
+sub flat_triggers {
+    my $data = shift;
+    if (my $triggers = $data->{triggers}) {
+        return map { @{$_} } (values $triggers);
+    }
+}
+
+sub check_trigger_existing {
+    my $trigger = shift;
+    return 0 if $triggers{$trigger};
+    $triggers{$trigger} = 1;
+    return 1;
+}
 
 sub file_name_id_match {
     my ($file_name, $id) = @_;
@@ -58,6 +77,17 @@ foreach my $path (glob("$json_dir/*.json")){
     if (my $cheat_id = $json->{id}) {
         $temp_pass = file_name_id_match($file_name, $cheat_id);
         push(@tests, {msg => "Invalid file name ($file_name) for ID ($cheat_id)", critical => 1, pass => $temp_pass});
+    }
+
+    ### Trigger tests ###
+    if (my $cheat_id = $json->{id}) {
+        if (my $custom = $triggers_yaml->{custom_triggers}->{$cheat_id}) {
+            # Duplicate triggers
+            foreach my $trigger (flat_triggers($custom)) {
+                $temp_pass = check_trigger_existing($trigger);
+                push(@tests, {msg => "trigger '$trigger' already in use", critical => 1, pass => $temp_pass});
+            }
+        }
     }
 
     ### Template Type tests ###
--- a/t/CheatSheets/CheatSheetsJSON.t
+++ b/t/CheatSheets/CheatSheetsJSON.t
@@ -15,11 +15,17 @@ use List::Util qw(first);
 my $json_dir = "share/goodie/cheat_sheets/json";
 my $json;
 
-sub file_name_to_id {
-    my $file_name = shift;
-    $file_name =~ s/\.json//;
-    $file_name =~ s/-/_/g;
-    return $file_name . "_cheat_sheet";
+sub file_name_id_match {
+    my ($file_name, $id) = @_;
+    my $check_id = $file_name =~ s/\.json//r;
+    $check_id =~ s/-/_/g;
+    $check_id .= '_cheat_sheet';
+    return 0 unless $check_id eq $id;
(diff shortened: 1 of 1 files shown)


daxtheduck:
Metadata placeholder post
            
---
            
**How to find your IA page**
- New Instant Answer
              
- Existing Instant Answer

GuiltyDolphin:
@zachthompson Do you think this is the best method for this test? (Performing both conversions).

Having just the file name to ID wasn't enough, because some characters (like `_`) could make it through. I think ID -> file name would be a bit more robust, but I'm not sure it would catch everything?

I'm thinking that each ID would have a unique file name associated with it - so ideally we should only have to check in one direction.

GuiltyDolphin:
@zachthompson I've added those tests you suggested. Tested them out and they work pretty well.

t/CheatSheets/CheatSheetsJSON.t
```diff
@@ -50,10 +76,27 @@ foreach my $path (glob("$json_dir/*.json")){
...
+        $temp_pass = file_name_id_match($file_name, $cheat_id);
         push(@tests, {msg => "Invalid file name ($file_name) for ID ($cheat_id)", critical => 1, pass => $temp_pass});
     }
 
+    ### Trigger tests ###
+    if (my $cheat_id = $json->{id}) {
+        if (my $custom = $triggers_yaml->{custom_triggers}->{$cheat_id}) {
+            # Duplicate triggers
+            foreach my $trigger (flat_triggers($custom)) {
+                $temp_pass = check_trigger_existing($trigger);
```
zachthompson: Since `%triggers` is global:
```Perl
$temp_pass = $triggers{$trigger} ? 0 : 1;
```

t/CheatSheets/CheatSheetsJSON.t
```diff
@@ -10,16 +10,42 @@ use Test::More;
...
+    return 1;
+}
+
+sub file_name_id_match {
+    my ($file_name, $id) = @_;
+    my $check_id = $file_name =~ s/\.json//r;
+    $check_id =~ s/-/_/g;
+    $check_id .= '_cheat_sheet';
+    return 0 unless $check_id eq $id;
+    # Check the inverse to make sure no weird cases get through.
```
zachthompson: This is the same case we discussed in the other pr where these are always both true or false.   I think we only need one.

t/CheatSheets/CheatSheetsJSON.t
```diff
@@ -10,16 +10,44 @@ use Test::More;
...
+my $triggers_yaml = LoadFile('share/goodie/cheat_sheets/triggers.yaml');
+
+my %triggers;
+
+my $template_map = $triggers_yaml->{template_map};
+
+sub flat_triggers {
+    my $data = shift;
+    if (my $triggers = $data->{triggers}) {
+        return map { @{$_} } (values $triggers);
```
zachthompson: `@$_` and no need for parens around `values ...`

t/CheatSheets/CheatSheetsJSON.t
GuiltyDolphin: @zachthompson Hmm... I ran into issues with #2667 where the file name wasn't being caught with the one-way conversion. Any ideas on overcoming this without checking both ways?

t/CheatSheets/CheatSheetsJSON.t
zachthompson: Actually we need the trigger type here.  An end rule for one cheat sheet should be ok as a start for another, no?  But it does make the check more complex.  The above won't work.

GuiltyDolphin: @zachthompson Humm.. The only ones that wouldn't clash are `end` and `startend` so... I guess we could, but it would make the code more complex and I think it might be likely that someone who has a `start` trigger for a cheat sheet may wish to change that to `end` or `startend` later on - so I think we should 'reserve' that trigger for them.

t/CheatSheets/CheatSheetsJSON.t
zachthompson: What issue were you seeing?  If I pull this code out it returns 1 for either of them with the file name and id from 2667.

GuiltyDolphin: It was originally `go_tools.json`.

zachthompson: The problem is underscores in the file name.  We have no good way to translate that to an ID.  The second test catches this properly, and the others from what I can tell.

GuiltyDolphin: So you reckon just the ID->file name test would be sufficient?

zachthompson: I tried quite a few combos and it worked ok.

GuiltyDolphin: Okay, sure, I've tried some out too and it seems fine.

GuiltyDolphin:
@zachthompson Righty. Any other tests you want? I can't really think of any more that would be super-beneficial at the moment. Hopefully these three will stop some of the more common mistakes.

zachthompson:
Great, thank you.

PR #2692, opened 2016-03-03
CheatSheets: Add alias trigger tests

GuiltyDolphin:
@zachthompson These ensure the aliases don't contain triggers (prevents aliases like `X cheat sheet`, `X shortcuts` etc...).

* ***Critical*** if an alias contains a trigger defined by a category that is assigned.
* ***Warning*** if an alias contains a trigger defined by a category that is not assigned.
* ***Critical*** if an alias contains a custom trigger.

These check trigger positions (`start`, `startend` etc...).

---

### DIFF
--- a/share/goodie/cheat_sheets/json/duckduckgo-syntax.json
+++ b/share/goodie/cheat_sheets/json/duckduckgo-syntax.json
@@ -7,6 +7,9 @@
     "sourceName": "Duck.co",
     "sourceUrl": "https://duck.co/help/results/syntax"
   },
+  "aliases": [
+    "ddg syntax","duck duck go syntax"
+  ],
   "section_order": [
     "Triggers",
     "Go directly to other sites",
--- a/share/goodie/cheat_sheets/json/duckduckgo-syntax.json
+++ b/share/goodie/cheat_sheets/json/duckduckgo-syntax.json
@@ -7,9 +7,6 @@
     "sourceName": "Duck.co",
     "sourceUrl": "https://duck.co/help/results/syntax"
   },
-  "aliases": [
-    "ddg syntax","duck duck go syntax"
-  ],
   "section_order": [
     "Triggers",
     "Go directly to other sites",
--- a/share/goodie/cheat_sheets/json/sql.json
+++ b/share/goodie/cheat_sheets/json/sql.json
@@ -15,9 +15,8 @@
       "SQL Constraints",
       "Scalar Functions"
    ],
-   "aliases":[  
-      "sql statements",
-       "sql commands"
+   "aliases":[
+      "sql statements"
    ],
    "sections":{  
       "Data Definition Language (DDL)":[  
--- a/share/goodie/cheat_sheets/json/sql.json
+++ b/share/goodie/cheat_sheets/json/sql.json
@@ -16,7 +16,8 @@
       "Scalar Functions"
    ],
    "aliases":[
-      "sql statements"
+      "sql statements",
+       "sql commands"
    ],
    "sections":{  
       "Data Definition Language (DDL)":[  
--- a/t/CheatSheets/CheatSheetsJSON.t
+++ b/t/CheatSheets/CheatSheetsJSON.t
@@ -22,6 +22,10 @@ my %triggers;
 
 my $template_map = $triggers_yaml->{template_map};
 
+my %category_test_whitelist = (
+    duckduckgo_syntax_cheat_sheet => 1,
+);
+
 sub flat_triggers {
     my $data = shift;
     if (my $triggers = $data->{triggers}) {
@@ -141,6 +145,7 @@ foreach my $path (glob("$json_dir/*.json")){
         }
         # Make sure aliases don't contain any category triggers.
         while (my ($category, $trigger_types) = each %{$triggers_yaml->{categories}}) {
+            last if $category_test_whitelist{$cheat_id};
             if (my ($alias, $trigger) = check_aliases_for_triggers(\@aliases, $trigger_types)) {
                 push(@tests, {msg => "alias ($alias) contains a trigger ($trigger) defined in the '$category' category", critical => 1});
             }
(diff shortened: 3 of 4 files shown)


daxtheduck:
Metadata placeholder post
            
---
            
**How to find your IA page**
- New Instant Answer
              
- Existing Instant Answer

share/goodie/cheat_sheets/json/duckduckgo-syntax.json
```diff
@@ -7,9 +7,6 @@
     "sourceName": "Duck.co",
     "sourceUrl": "https://duck.co/help/results/syntax"
   },
-  "aliases": [
-    "ddg syntax","duck duck go syntax"
```
GuiltyDolphin: This is a bit of a weird case - we can't just say "ddg" and "duck duck go" as aliases and add the `syntax` trigger, because there is another cheat sheet that uses those aliases. Adding `ddg syntax` to the triggers would only help it to trigger on things like `duckduckgo syntax ddg syntax` etc...

@zachthompson The way I see it we've got two choices:

1. Only check for the categories that are assigned to the given cheat sheet - but this is likely to miss many erroneous triggers in aliases.
2. Have some way of ignoring individual cheat sheets (or maybe we could have a white-list in the tests?) - cases such as this should be extremely rare and edgy!

Can you think of any better options?

/cc @moollaza

GuiltyDolphin:
@moollaza @zachthompson I've chosen to white-list the `duckduckgo_syntax_cheat_sheet` seeing as it is the only edge-case out of several hundred cheat sheets. I think it would be a shame to remove half the power of the tests for one cheat sheet.

share/goodie/cheat_sheets/json/sql.json
```diff
@@ -15,9 +15,8 @@
       "SQL Constraints",
       "Scalar Functions"
    ],
-   "aliases":[  
-      "sql statements",
-       "sql commands"
```
zachthompson: Don't think we want to remove this as they are pretty commonly referred to as "sql commands".

t/CheatSheets/CheatSheetsJSON.t
```diff
@@ -37,6 +41,28 @@ sub id_to_file_name {
...
+    while (my ($trigger_type, $triggers) = each %{$trigger_types}) {
+        my @triggers = @$triggers;
+        foreach my $alias (@aliases) {
+            my $trigger;
+            if (
+                ($trigger_type =~ /^start/
+                    && ($trigger = first { $alias =~ /^$_/ } @triggers))
+            ||  ($trigger_type =~ /end$/
+                    && ($trigger = first { $alias =~ /$_$/ } @triggers))
+            ||  ($trigger_type eq 'any'
```
zachthompson: Let's remove `any` since it breaks everything and make a note of it in triggers.yaml.  I think @moollaza raised this issue.

share/goodie/cheat_sheets/json/sql.json
GuiltyDolphin: @zachthompson We can add `commands` to the additional triggers for `sql`, would that be better?

zachthompson: I think you might get searches like "sql commands cheatsheet" which wouldn't work, rather than just "sql commands" which I don't think was intended.

GuiltyDolphin: Hmm... Think I should move to the alternative approach then? Only check conflicts with assigned categories?

zachthompson: I'm not sure I follow how that is different.

GuiltyDolphin: Currently it checks *all* category triggers - so even though `sql` doesn't have the `terminal` category assigned it will still check for the `commands` trigger.

The alternative would be to *just* have it check against triggers that are assigned based on the categories - so this would be things like `cheat sheet` and `cheatsheet` for all cheat sheets, but varying others based on the `template_type`.

zachthompson: Hmm, it's probably worthwhile to do the first as I can see people trying to put aliases for triggers they aren't aware of, that come later, etc.  In this case it would help because terminal isn't specified? I'm a little hesitant to have all of these exceptions building up.

GuiltyDolphin: @zachthompson What would your suggestion for `sql` be then? White-list it? It will fail unless we have an exception for it (if we want the `sql commands` alias).

zachthompson: The alternative is fine.  It should catch most of cases.  You could also just warn for categories outside of those specified.

GuiltyDolphin: Good idea!

Done :smile: - It will warn if it is a non-assigned category, and fail otherwise.

GuiltyDolphin:
@moollaza @zachthompson Okay! I've opted to do critical if the triggers are specified as custom for the cheat sheet, or are in an assigned category, and only a warning (good suggestion @zachthompson !) if the category is *not* assigned to the cheat sheet.

t/CheatSheets/CheatSheetsJSON.t
```diff
@@ -63,8 +83,9 @@ foreach my $path (glob("$json_dir/*.json")){
     $temp_pass = (!exists $json->{description} && !$json->{description})? 0 : 1;
     push(@tests, {msg => "has description (optional but suggested)", critical => 0, pass => $temp_pass});
 
+    my $cheat_id;
     ### ID tests ###
-    if (my $cheat_id = $json->{id}) {
+    if ($cheat_id = $json->{id}) {
```
zachthompson: Can't we just combine these two `if` blocks that test for `$cheat_id`?  If not, do the assignment on line 86.  It looks like you're trying to set it dynamically, rather than just reuse it.

GuiltyDolphin: @zachthompson Yup! I've combined the tests that require the ID to be set.

zachthompson:
Thank you.

PR #2705, opened 2016-03-05
SubnetCalc: Minor tidying up of subs

mintsoft:
I did this as a test, might as well get it in master

----------
IA Page : https://duck.co/ia/view/subnet_calc

### DIFF
--- a/lib/DDG/Goodie/SubnetCalc.pm
+++ b/lib/DDG/Goodie/SubnetCalc.pm
@@ -7,7 +7,9 @@ use warnings;
 
 use DDG::Goodie;
 
-triggers query => qr#^(?:[0-9]{1,3}\.){3}(?:[0-9]{1,3})[\s/](?:(?:[1-3]?[0-9])|(?:(?:[0-9]{1,3}\.){3}(?:[0-9]{1,3})))$#x;
+my $ipv4 = qr#(?:[0-9]{1,3}\.){3}(?:[0-9]{1,3})#;
+
+triggers query => qr#^$ipv4[\s/](?:(?:[1-3]?[0-9])|$ipv4)$#x;
 
 zci answer_type => "subnet_calc";
 zci is_cached => 1;
@@ -32,7 +34,7 @@ handle query => sub {
 
         # Flip the bits.
         $mask = 0xffffffff ^ $wildcard_mask;
-    } elsif ($cidr =~ /([0-9]{1,3}\.){3}([0-9]{1,3})/) {
+    } elsif ($cidr =~ /$ipv4/) {
         my @cidr_octlets = split /\./, $cidr;
 
         # An octlet cannot be over 255.
--- a/lib/DDG/Goodie/SubnetCalc.pm
+++ b/lib/DDG/Goodie/SubnetCalc.pm
@@ -21,7 +21,7 @@ handle query => sub {
         return if int($_) > 255;
     }
 
-    my $address_raw = ip_to_int(@octlets);
+    my $address_raw = ip_bytes_to_int(@octlets);
     my $wildcard_mask = 0;
     my $mask = 0;
 
@@ -40,7 +40,7 @@ handle query => sub {
             return if int($_) > 255;
         }
 
-        $mask = ip_to_int(@cidr_octlets);
+        $mask = ip_bytes_to_int(@cidr_octlets);
         # Convert the integer into its binary representation.
         my $mask_binary = sprintf("%b", $mask);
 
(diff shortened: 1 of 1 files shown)


daxtheduck:
Metadata placeholder post
            
---
            
**How to find your IA page**
- New Instant Answer
              
- Existing Instant Answer

lib/DDG/Goodie/SubnetCalc.pm
```diff
@@ -128,4 +111,21 @@ handle query => sub {
         };
 };
 
-1;
+# Convert an integer into an IP address.
+sub int_to_str {
```
GuiltyDolphin: Hmm... Personally I prefer the handler to go at the bottom (just makes more 'flow' sense to me), what do you think?

lib/DDG/Goodie/SubnetCalc.pm
```diff
@@ -13,17 +13,6 @@ zci answer_type => "subnet_calc";
 zci is_cached => 1;
 
 handle query => sub {
```
GuiltyDolphin: Hmm... Personally I prefer the handler to go at the bottom (just makes more 'flow' sense to me), what do you think?

mintsoft: Hmm, I always wax and wane on that, recently I've been finding it easier to debug/review/understand if IA's follow a pattern of:
* usings
* statics
* triggers
* handler
* misc subs
as I usually like to start with the handler, so I want that near the top, as close to the triggers as possible. Technically it makes no difference obviously, Perl will let you define things wherever you like. @MrChrisW what are you preferences here?

GuiltyDolphin: @mintsoft Yeah - same for me but with `handler` and `misc subs` switched over ;) Though usually I'm not too fussed about where the `triggers` are depending on the code - usually near the top or near the handle though.

MrChrisW: @mintsoft My preference is to have the subs below the handler. If I'm debugging I'll look at the handler first then follow the program flow from there (moving to other subs) - in this case it makes sense to have handler first. It really is just personal preference though! :+1:

GuiltyDolphin: @MrChrisW @mintsoft I'm outnumbered!

Very well, this isn't a *big* issue for me - as long as it is split into reasonable chunks!

lib/DDG/Goodie/SubnetCalc.pm
GuiltyDolphin: Why `int_to_str` and not `int_to_ip_str`?

lib/DDG/Goodie/SubnetCalc.pm
MrChrisW: > as long as it is split into reasonable chunks!
@GuiltyDolphin Agreed! :+1:

lib/DDG/Goodie/SubnetCalc.pm
mintsoft: :+1:

lib/DDG/Goodie/SubnetCalc.pm
```diff
@@ -128,4 +111,21 @@ handle query => sub {
...
-1;
+# Convert an integer into an IP address.
+sub int_to_str {
+    my ($ip) = @_;
+    sprintf "%u.%u.%u.%u", $ip >> 24 & 0xff, $ip >> 16 & 0xff, $ip >> 8 & 0xff, $ip & 0xff;
+}
+
+# Convert an IP address into an integer.
+sub ip_to_int {
+    (int($_[0]) << 24) + (int($_[1]) << 16) + (int($_[2]) << 8) + int($_[3]);
```
GuiltyDolphin: Lot of bit shifting going on in this file :wink:

GuiltyDolphin:
@mintsoft The trigger regex seems to have a lot of repetition in it - could we condense it a bit to make it easier to read?

lib/DDG/Goodie/SubnetCalc.pm
```diff
@@ -128,4 +111,21 @@ handle query => sub {
         };
 };
 
-1;
+# Convert an integer into an IP address.
+sub int_to_ip_str {
+    my ($ip) = @_;
+    sprintf "%u.%u.%u.%u", $ip >> 24 & 0xff, $ip >> 16 & 0xff, $ip >> 8 & 0xff, $ip & 0xff;
```
GuiltyDolphin: Seeing as the second half of this is the inverse of `ip_bytes_to_int`, perhaps we could pull it out into `int_to_ip_bytes` to make it clearer?

mintsoft: Hmm, I think personally I'd rather not, if it was used in multiple places, then I'd say yes. This is relatively standard for anything dealing with IP addresses, the ipv4 format is well documented etc. There's no value that I can see to provide an explicit function that returns an array other than maintaining symmetry of the subs; which seems kinda pointless to me

GuiltyDolphin: @mintsoft Fair enough.

GuiltyDolphin:
@mintsoft This is looking pretty good!

Much clearer than before - let me know when you're happy.

mintsoft:
btw @GuiltyDolphin I'm happy with this if you are

GuiltyDolphin:
@mintsoft Sweet! I've gone through it again and it looks much better! :dolphin:

PR #2713, opened 2016-03-07
Update a bunch of tab-answer names

GuiltyDolphin:
I for one ***love*** the answer tabs!

Who doesn't want to have a choice of Instant Answers?:

screenshot from 2016-03-07 16 03 51

Unfortunately many of the Goodies were just using the `Answer` tab name, and some even had weird non-standard names!

@moollaza I've updated the tab names as best I could, but I have a couple of requests:

1. We add a `Date & Time` category: We have *many* Instant Answers that would benefit from this (`Zodiac`, `CalendarToday`, `LeapYear` etc...)

2. We add a `Web` or `Net` category: Goodies such as `URLEncode`, `URLDecode`, and others would benefit from this!

3. We add a `Fun & Games` category - it doesn't feel right putting things into `Games` a lot of the time - for instance `MagicEightBall` - but I think `Fun & Games` is a pretty good catch-all. (There is a `Words & games` category but I'm not sure what that is *supposed* to apply to? Board games? - I much prefer `Fun & Games`).

4. We add an `Easter Eggs` category - we have quite a few of these! (Also, this would tell the person that yes, they have discovered an Easter Egg!)

5. We add a `Random` category for all the Instant Answers that are generating random stuff (there's quite a few)!

6. Either we add an `Engineering` category, or rename `Science` to `Science & Engineering` - there's a few (such as `ResistorColors`) for which this would make *more* sense.

7. We add a `Language` category for anything that deals with language and/or translations (such as `PigLatin`).

8. Sort of a design preference - rather than having lowercase as in `Food & drinks`, we always use uppercase `Food & Drinks` - I think it looks better.

Note: I'm basing all the *allowed* categories on what is stated in the docs, so the suggestion is we *add* these to the docs (as well as IAs).

/cc @duckduckgo/community-leaders @talsraviv


daxtheduck:
Metadata placeholder post
            
---
            
**How to find your IA page**
- New Instant Answer
              
- Existing Instant Answer

GuiltyDolphin:
As a side note on this: While going through the Goodies I noticed *many* were still not using templates! @duckduckgo/instant-answer-maintainers

PR #2740, opened 2016-03-13
CheatSheets: Improve test phrasing and reports

GuiltyDolphin:
@zachthompson As promised! This improves the general phrasing of the tests as well as improve the way the tests are reported. There's a bit more info in my commit messages, but you can easily see the new structure through the Travis log.

* Improves phrasing of existing tests

* Improves structure of test reports

* Fixes an issue wherein cheat sheets within directories under `json/` were not being tested

* Allows multiple cheat sheets to be tested via arguments

---
https://duck.co/ia/view/cheat_sheets

PR #2289, opened 2016-02-14
Sils/spacing

sils1297:
Hi, I saw https://github.com/duckduckgo/zeroclickinfo-goodies/pull/2288#discussion_r52844722 and thought I might help since we're currently developing a tool that can do all kinds of static code analysis.

This PR corrects all the spacing youve got globally on all unicode files. It also adds a config (if you don't want it I can of course remove the commit) for the tool I used. With the config you can use gitmate.io which is the free software service I'm currently building, it's a prototype and not very reliable but it'll automatically tell you if new issues arise as PR comments. (You *can* use it in production but you really shouldn't rely on it and for sure first test if you like all the features on a test repo since you can't turn anything off yet.) If that sounds too shaky (I bet it does) you can just invoke `coala-ci` in your CI builds and get some pretty output if something is not according to the defined style.

While I was at it I spotted some JSON and unified the spacing there too (if you want it can be sorted as well :)). If you want deeper analysis there's lots of it. CSS, JS, Python, you name it. (If it's missing we can add it.)

PR #2741, opened 2016-03-13
CheatSheets: Fix whitespace (attributes)

GuiltyDolphin:
@zachthompson This fixes a bunch of whitespace issues picked up by #2740 (I also thought I'd kill of some other whitespace issues at the same time) - I'm including it separately so there isn't as much clutter on that PR (and you get to see the new reporting).

*Note:* The primary aim of this PR is *not* to clean up *general* whitespace, but rather *specific* forms of whitespace (within attributes) picked up by #2740. Other whitespace corrections are included as an... *'automatic'* bonus.

### DIFF
--- a/share/goodie/cheat_sheets/json/age-of-mythology.json
+++ b/share/goodie/cheat_sheets/json/age-of-mythology.json
@@ -1,8 +1,8 @@
 {
-    "id": "age_of_mythology_cheat_sheet", 
+    "id": "age_of_mythology_cheat_sheet",
     "name": "Age of Mythology Cheat Codes",
-    "description": "Full list of Age of Mythology Codes", 
-    "metadata": { 
+    "description": "Full list of Age of Mythology Codes",
+    "metadata": {
         "sourceName": "HeavenGames",
         "sourceUrl": "http://aom.heavengames.com/gameinfo/cheats"
     },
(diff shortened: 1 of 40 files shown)


daxtheduck:
Metadata placeholder post
            
---
            
**How to find your IA page**
- New Instant Answer
              
- Existing Instant Answer

altern8tif:
@GuiltyDolphin I think this might have been brought up before (probably by @moollaza)... 

Using [HoundCI](http://www.houndci.com) would allow us to directly fix trailing whitespace in code (by implementing a style guide) rather than throwing warnings which still requires a PR to fix... They use JSHint which should work for fixing JSON files...

GuiltyDolphin:
@altern8tif This isn't for trailing whitespace in general - just in certain attributes within the cheat sheets. Vim just automatically sorts out trailing whitespace for me so I thought I'd include it (I use `git add --patch` a lot - a pain with whitespace corrections).

We should note that not *all* fields have to be corrected - only certain `val`, `trn` etc...

If you want to know more about what is *actually* being checked, just have a look at #2740.

altern8tif:
@GuiltyDolphin Ah ok I see... Yea I've got Sublime Text set up to sort out trailing whitespace for me too.

Also realised that style guides are being talked about in #2289 already... Hope that we can implement some kind of style guide to keep code more consistent.  Maybe it's just my OCD, but I sometimes feel a bit uneasy when I see files using different spacing/tabs :grimacing:

GuiltyDolphin:
@altern8tif Yeah - we're automating more and more though, so if people follow the *existing* style introduced by the tools we should all live happily - if they don't... Then :cry:

moollaza:

PR #2777, opened 2016-03-19
CheatSheets: Add 'ignore' attribute

GuiltyDolphin:
@zachthompson As discussed previously on #2170, this re-introduces the 'ignore' attribute.

I've gone over the purpose in some detail in he message for my first commit, but the TL;DR is they allow 'fluff' words that are ignored when triggering.

These are specified in the `triggers.yaml` file.

## TO-DO:

- [ ] Add 'global' ignores: These would be phrases that are automatically ignored by any cheat sheet / template using the category - but would apply to *additional triggers*, not just those defined by the category.
- [x] Add tests. The aim would be to have these tests fail if an alias contains a phrase that would be entirely ignored by the triggering. Would like to rebase onto #2740 first.

---
Maintainer: @zachthompson

### DIFF
--- a/lib/DDG/Goodie/CheatSheets.pm
+++ b/lib/DDG/Goodie/CheatSheets.pm
@@ -45,12 +45,23 @@ sub generate_triggers {
     my %triggers;
     # This will contain a lookup from triggers to categories and/or files.
     my %trigger_lookup;
+    # This will contain all the ignored phrases and their associated categories.
+    my %ignore_phrases;
 
     while (my ($name, $trigger_setsh) = each %spec_triggers) {
+        my $ignore_phrases = delete $trigger_setsh->{ignore};
         while (my ($trigger_type, $triggersh) = each %$trigger_setsh) {
             foreach my $trigger (@{$triggersh}) {
                 # Add trigger to global triggers.
                 $triggers{$trigger_type}{$trigger} = 1;
+                # Handle ignored components - these will be stripped
+                # from query and not be included in final trigger.
+                if (defined $ignore_phrases) {
+                    my %new_ignore_phrases = map { $_ => 1 }
+                        (keys %{$ignore_phrases{$trigger} || {}},
+                        @$ignore_phrases);
+                    $ignore_phrases{$trigger} = \%new_ignore_phrases;
+                }
                 my %new_triggers = map { $_ => 1}
                     (keys %{$trigger_lookup{$trigger}});
                 if ($name !~ /cheat_sheet$/) {
@@ -65,7 +76,7 @@ sub generate_triggers {
     while (my ($trigger_type, $triggers) = each %triggers) {
         triggers $trigger_type => (keys %{$triggers});
     }
-    return %trigger_lookup;
+    return (\%ignore_phrases, %trigger_lookup);
 }
 
 # Initialize aliases.
@@ -100,13 +111,19 @@ sub get_aliases {
 
 my $aliases = get_aliases();
 
-my %trigger_lookup = generate_triggers($aliases);
+my ($trigger_ignore, %trigger_lookup) = generate_triggers($aliases);
 
 handle remainder => sub {
     my $remainder = shift;
 
     my $trigger = join(' ', split /\s+/o, lc($req->matched_trigger));
     my $lookup = $trigger_lookup{$trigger};
+    if (exists $trigger_ignore->{$trigger}) {
+        foreach my $ignore (keys %{$trigger_ignore->{$trigger}}) {
+            $remainder =~ s/\b$ignore\b//;
+        }
+        $remainder =~ s/^\s*(.+?)\s*$/$1/;
+    }
 
     my $file = $aliases->{join(' ', split /\s+/o, lc($remainder))} or return;
     open my $fh, $file or return;
--- a/lib/DDG/Goodie/CheatSheets.pm
+++ b/lib/DDG/Goodie/CheatSheets.pm
@@ -114,18 +114,14 @@ my $aliases = get_aliases();
 my ($trigger_ignore, %trigger_lookup) = generate_triggers($aliases);
 
 handle remainder_lc => sub {
-    my $remainder = shift;
+    my @query_parts = split /\s+/o, shift;
 
     my $trigger = join(' ', split /\s+/o, lc($req->matched_trigger));
     my $lookup = $trigger_lookup{$trigger};
-    if (exists $trigger_ignore->{$trigger}) {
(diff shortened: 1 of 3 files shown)


daxtheduck:
Metadata placeholder post
            
---
            
**How to find your IA page**
- New Instant Answer
              
- Existing Instant Answer

zachthompson:
Do we have a use case for global ignores?  We might want to chalk that up in an issue.  Probably bigger priority things to address.  When does the 'english to' trigger exactly?

GuiltyDolphin:
@zachthompson If a cheat sheet triggers on 'phrases' (or any of the triggers defined by 'language'), then the whole trigger is filtered to remove 'english to', to hopefully leave only the cheat sheet name (or an alias) remaining, which is then used to get the JSON data.

zachthompson:
Could be added as an alias though.  Some of theme are doing it, e.g. 'spanish to english'.

GuiltyDolphin:
@zachthompson Indeed. Unfortunately this does ***not*** allow combination of multiple phrases, and does not scale well with additional phrases (having to add them to every cheat sheet).

Having ignore phrases in the triggers file means cheat sheets do not have to make up a bunch of additional aliases (that do not serve as true aliases anyway), allows applying the additional forms to multiple cheat sheets at once, and enforces more consistent triggering.

Ignoring just the `basic` and `english to` phrases (at the start) alone would require another 3 aliases for every language cheat sheet (`basic X`, `basic english to X`, `english to X`).

As you've mentioned before, it is good to take as much of the triggering out of the hands of individual cheat sheets as possible - so we can rely on a similar behavior across cheat sheets as a whole.

zachthompson:
Yep, I agree with that.  Just trying to balance complexity with real need.  The global ignores I'm not so sure about.

GuiltyDolphin:
@zachthompson I don't think this has increased the complexity by much - but yes, I agree, we don't want to over complicate.

As mentioned on #2781, I think this could play quite well with using topics for categories - for example, we could ignore 'game' for game cheat sheets - automatically enabling triggers like `foo game cheats`.

zachthompson:
@GuiltyDolphin Where are we at with this?  I see you still have to-dos open, etc.

GuiltyDolphin:
@zachthompson And some merge conflicts by the look of it!

It looks like we might not be going for the global ignores?

I'll have a look over the week and sort out anything that needs doing.

zachthompson:
@GuiltyDolphin Let's hold off on the global ignores for now and add to an issue for later review.  I don't think I understand the use case there.

GuiltyDolphin:
@zachthompson Thinking about it, I don't think the tests are required (for checking if 'ignore' phrase is included in alias); the trigger isn't affected when removing the ignored phrases so there should never be a collision.

GuiltyDolphin:
@zachthompson Right - I've added *warning* tests for if an alias contains an ignored phrase, letting the user know it may be omitted.

Do you think we should make this critical?

zachthompson:
@GuiltyDolphin Yeah, let's make that critical.  At best it just creates overhead.

GuiltyDolphin:
@zachthompson Done - I've sent out a maintainer issue for all the cheat sheets that need updating; apart from that you happy with this?

zachthompson:
@GuiltyDolphin I'll try to take a look at it Thu/Fri.

GuiltyDolphin:
@zachthompson All the external conflicts have been resolved, this should be ready unless you have any comments.

GuiltyDolphin:
@zachthompson Any updates?

zachthompson:
@GuiltyDolphin I'll take a look at it today.

lib/DDG/Goodie/CheatSheets.pm
```diff
@@ -100,13 +111,19 @@ sub get_aliases {
...
+my ($trigger_ignore, %trigger_lookup) = generate_triggers($aliases);
 
 handle remainder => sub {
     my $remainder = shift;
 
     my $trigger = join(' ', split /\s+/o, lc($req->matched_trigger));
     my $lookup = $trigger_lookup{$trigger};
+    if (exists $trigger_ignore->{$trigger}) {
+        foreach my $ignore (keys %{$trigger_ignore->{$trigger}}) {
+            $remainder =~ s/\b$ignore\b//;
```
zachthompson: These regexes are a bit slow: compare [this PR](https://zach.duckduckgo.com/nytprof/2777/CheatSheets-pm-8-line.html#116), [master](https://zach.duckduckgo.com/nytprof/csmaster/CheatSheets-pm-8-line.html#105), and another [possible implementation](https://zach.duckduckgo.com/nytprof/2777j/CheatSheets-pm-8-line.html#116).

lib/DDG/Goodie/CheatSheets.pm
```diff
@@ -100,13 +111,19 @@ sub get_aliases {
 
 my $aliases = get_aliases();
 
-my %trigger_lookup = generate_triggers($aliases);
+my ($trigger_ignore, %trigger_lookup) = generate_triggers($aliases);
 
 handle remainder => sub {
```
zachthompson: Just realized we can probably use `remainder_lc` and skip doing it below.

lib/DDG/Goodie/CheatSheets.pm
GuiltyDolphin: @zachthompson Hmm, looks like the other implementation wins? I'll test it out locally and update it if everything is okay :+1:

lib/DDG/Goodie/CheatSheets.pm
GuiltyDolphin: @zachthompson That one ain't even in the docs!

zachthompson: whoops! ;)

lib/DDG/Goodie/CheatSheets.pm
zachthompson: Yeah, not sure if we can tweak the REs or not.  Either way.

GuiltyDolphin: @zachthompson Your implementation seems pretty good (though you'd want to join with ' ', not ''); I'm pushing now.

lib/DDG/Goodie/CheatSheets.pm
GuiltyDolphin: @zachthompson I've added it - hopefully we haven't missed any others ^.^

GuiltyDolphin:
@zachthompson Right, anything else?

zachthompson:
Nice, thank you.

PR #2802, opened 2016-03-26
Remove ID and name from templates

GuiltyDolphin:
@zachthompson @moollaza I've removed the ID and name attributes from the structured answers in the templates.

I have also (you may disagree with this) added a template for building the structured answer response in tests. Seeing as we are removing 'simple' structured answers I feel this is appropriate as it will reduce the visual load of having many structured answer definitions. Additionally it should make it easier to move to a new testing system if desirable in the future.

### DIFF
--- a/template/share/goodie/example/example.js
+++ b/template/share/goodie/example/example.js
@@ -17,8 +17,6 @@ DDH.<: $ia_id :> = DDH.<: $ia_id :> || {};
 
         return {
 
-            id: '<: $ia_id :>',
-
             meta: {
                 sourceName: "Source Domain",
                 sourceUrl: "https://source.website.com"
--- a/template/lib/DDG/Goodie/Example.pm
+++ b/template/lib/DDG/Goodie/Example.pm
@@ -35,15 +35,6 @@ handle <: $ia_handler :> => sub {
     return "plain text response",
         structured_answer => {
 
-            # ID - Must be unique and match Instant Answer page
-            # E.g. https://duck.co/ia/view/calculator has `id => 'calculator'``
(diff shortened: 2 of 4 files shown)


daxtheduck:
Metadata placeholder post
            
---
            
**How to find your IA page**
- New Instant Answer
              
- Existing Instant Answer

moollaza:
Assigning to @zachthompson. Would like your thoughts on the test "helper". Should that just be ported to  duckduckgo/duckduckgo?

GuiltyDolphin:
@moollaza Herumm... Maybe, but I'm not sure if it'd be able to handle the varying forms of IAs? For example, conversions has a custom spin on this, so does interactive-calculator.

The general idea is it is ***much*** easier to update and read if we only have to add say, one line rather than a whole structured answer - then people can customize the function to suit their own testing needs.

zachthompson:
@moollaza Doesn't the [Test::Deep change](https://github.com/duckduckgo/duckduckgo/pull/172) at least partially alleviate this?  What's the hold up there?

template/share/goodie/example/example.js
```diff
@@ -17,8 +17,6 @@ DDH.<: $ia_id :> = DDH.<: $ia_id :> || {};
 
         return {
 
-            id: '<: $ia_id :>',
```
zachthompson: @moollaza we still need this id?

GuiltyDolphin: @zachthompson In the JS? I thought this just *extended* the response from the Perl?

zachthompson:
Thanks.

PR #2808, opened 2016-03-27
Countdown Goodie

rasikapohankar:
Basic version that gives the countdown for time specified as am/pm and days of the week, still want to add handling for countdown to a specific date, or day.

IA Page: https://duck.co/ia/view/countdown


daxtheduck:
Metadata placeholder post
            
---
            
**How to find your IA page**
- New Instant Answer
              
- Existing Instant Answer

lib/DDG/Goodie/Countdown.pm
```diff
@@ -0,0 +1,34 @@
...
+
+triggers any => 'countdown to';
+
+# Handle statement
+handle remainder => sub {
+
+    return "",
+        structured_answer => {
+
+            id => 'countdown',
```
moollaza: Please remove the `name` and `id` -- we very recently removed them from all IAs and I will be updating the templates so they are not included there either.

From now on, the IA Page will house this information

share/goodie/countdown/countdown.handlebars
```diff
@@ -0,0 +1,32 @@
+<div class="timer_container tile status_running">
+    <!--<div class="tx-clr--slate play_pause">
```
moollaza: Are these commented blocks supposed to be there?

Let's remove them if they're not being used

rasikapohankar: @moollaza : I copied the whole of that file from the timer code and am now making changes to it. Will remove the commented blocks.

GuiltyDolphin: @rasikapohankar I don't think we want much from the timer - I doubt it'd be useful to have an `X days to Y` countdown in that format, especially considering the multiple timers.

Lets focus on one display for the result.

lib/DDG/Goodie/Countdown.pm
```diff
@@ -0,0 +1,34 @@
...
+
+            name => 'Answer',
+
+            data => {
+                remainder => $_
+            },
+
+            templates => {
+                group => "text",
```
GuiltyDolphin: We probably want to use a base template if this is going to be interactive/live.

lib/DDG/Goodie/Countdown.pm
```diff
@@ -0,0 +1,34 @@
...
+zci answer_type => 'countdown';
+
+zci is_cached => 1;
+
+triggers any => 'countdown to';
+
+# Handle statement
+handle remainder => sub {
+
+    return "",
```
GuiltyDolphin: It would be useful if we gave the text version here. So `days until sunday` -> `3 days` or similar.

lib/DDG/Goodie/Countdown.pm
```diff
@@ -0,0 +1,34 @@
...
+# ABSTRACT: Provides a countdown to a particular date or time
+
+use DDG::Goodie;
+use strict;
+
+zci answer_type => 'countdown';
+
+zci is_cached => 1;
+
+triggers any => 'countdown to';
```
GuiltyDolphin: Lets expand. Supporting queries of the form `days until`, `how long until` etc... (many others) would be mighty fine.

share/goodie/countdown/countdown.js
```diff
@@ -0,0 +1,147 @@
...
+            } else {
+                break;
+            }
+        }
+    }
+    
+    function displayCountdown() {                
+        var parts = countdown.split(":");
+        if(parts.length > 1) {
+            $(".time_display .hours_minutes").html(parts[0]+":"+parts[1]+":"+parts[2]);
```
GuiltyDolphin: I think we'll want to support days as well. Maybe even weeks and/or years.

We could dynamically change which form is displayed based on the initial time difference. `How long until 20th march 2057` would probably want to go up to years, but `Time until 5 o'clock` would probably only want to go up to hours.

share/goodie/countdown/countdown.css
```diff
@@ -0,0 +1,363 @@
...
+}
+
+.zci--countdown .timer_container.status_stopped .play_pause a {
+    color: #4ab857;
+}
+
+/* button look for different timer statuses */
+
+.zci--countdown .play_pause a:after {
+    content: "►";
```
GuiltyDolphin: We ain't gonna want to be able to play/pause :wink:

share/goodie/countdown/countdown.js
rasikapohankar: @GuiltyDolphin : I have been looking at the timer's code for reference. Yes, we should remove play/pause, adding of multiple timers, adding a minute, and also naming the timer. Copy the query as the countdown timer's name? For now I have done it only as days:hrs:mins:secs. Sorry about the names of everything, I haven't yet changed the css selectors' names.

rasikapohankar:
@GuiltyDolphin : I didn't understand the point about the base template.

share/goodie/countdown/countdown.js
GuiltyDolphin: Seeing as we'd have one display we could be more elaborate with wording - `X days, Y hours, Z minutes, A seconds` or maybe `D days, HH:MM:SS`.

GuiltyDolphin:
@rasikapohankar Because we're building up an interface that won't rely on any templates for display, we can just use the `base` template so we don't have any unnecessary defaults.

lib/DDG/Goodie/Countdown.pm
rasikapohankar: @GuiltyDolphin : Then we will need to calculate the difference here as well? Right now the difference is only getting calculated in the javascript.

GuiltyDolphin: @rasikapohankar *Only* calculate the difference here - pass it off to the JS. My understanding is that after that it will be interactive right? Then we won't need to perform any text operations in the JS.

rasikapohankar: Hey @GuiltyDolphin! This is what I've done so far: using Moment.js finding the difference by constructing a moment object for the current time and a moment object for whatever is specified in the query by parsing the query, and this difference is updated every second. So to find the difference here we will need to shift the parsing to the perl, find the difference in seconds and pass that difference to the JS?

GuiltyDolphin: @rasikapohankar Whoops! I thought I was on a different PR - ignore that, doing the difference in the JS is fine!

What we'll want to do is parse the date/initial countdown in the Perl and pass that to the JS. For example `countdown to Jan 1st 2016` -> parse the date (we get `2016-01-01`), then pass that to the JS (probably in nanosecond form).

rasikapohankar: Or could we convert 1st Jan 2016 to the date also in the JS? In the perl only verify that the query is of the form 1st Jan 2016 otherwise don't send to the javascript?

GuiltyDolphin: @rasikapohankar It should be pretty easy to convert the date in the Perl - we have the Dates GoodieRole which should be able to do pretty much all the work.

share/goodie/countdown/countdown.js
```diff
@@ -0,0 +1,190 @@
...
+            regex = new RegExp(/[\s]+([\d]{1,2})[\s]*(?:st|th|nd)?[\s]*(?:[\s]+|[\/|-]?)[\s]*(?:(Jan(?:uary)?|Feb(?:ruary)?|Mar(?:ch)?|Apr(?:il)?|May|Jun(?:e)?|Jul(?:y)?|Aug(?:ust)?|Sept(?:ember)?|Oct(?:ober)?|Nov(?:ember)?|Dec(?:ember)?|[\d]{1,2}))[\s]*(?:[\s]+|[\/|-]?)[\s]*([\d]{4})[\s]*/i);
+            match = regex.exec(query);
+            console.log(match);
+            if(match) {                
+                if(isNaN(match[2])) {                    
+                    date = match[3]+"/"+getMonth(match[2].toLowerCase())+"/"+match[1];                    
+                } else {
+                    date = match[3]+"/"+match[2]+"/"+match[1];    
+                }                                              
+            }
```
rasikapohankar: @GuiltyDolphin : I had done it like so. Then we should move this part to the perl? And in the case where the query is like countdown to x am/pm?

GuiltyDolphin: @rasikapohankar Currently the Dates Role provides the `time_12h` and `time_24h` routines, that produce regular expressions for matching 12 and 24 hour time respectively. This should allow matching against `5 am` or `5 pm`.

Yes, but we don't need to do all this extracting - we can just use `parse_datestring_to_date` and then pull out relevant information.

rasikapohankar: @GuiltyDolphin Okay..I will try doing it using the Dates Goodie Role and let you know. Thanks.

rasikapohankar:
@GuiltyDolphin: I have tried writing it to find the initial difference using the Dates GoodieRole. It should work for queries of type x am/pm datestring or datestring x am/pm. I had some doubts : 1. The calculation is using the timezone UTC or if we give it as local then it takes the time from America/New_York. How can we get the user's timezone to perform the calculation? 2. When the entered date is after the current date, I am not returning a valid response. And if the user says countdown to 10 pm but it is past 10 pm on that day, then count down till next day's 10 pm. Is this okay?

GuiltyDolphin:
@rasikapohankar You should just be able to do `$my_date_object->time_zone` to get the time zone. The Dates role should automatically grab the correct time zone.

I think it is fine to count down to the *next* occurrence, seeing as we can't count backwards! We would want to make the exact date/time it is counting down to clear though (`date_output_string($date, 1)` will give full clock format).

rasikapohankar:
Hey @GuiltyDolphin! I tried by giving a date of the form 26/07/2016, so that the Dates Role returns a date object and then did `$then->time_zone()` , it still returns America/New_York. Did I miss something? The time zone is found by looking for tz in /etc on Unix for example. Is that right? So I thought that maybe this will always return the value found on the server, which might not be equal to the user's timezone?

GuiltyDolphin:
@rasikapohankar Locally the TZ will always be in America unless you use `DDG::Test::test_location` in testing or change it manually. 

Try doing `DDG_LOCALE='ms_MY' duckpan server` and check the results.

moollaza:
@rasikapohankar thanks again for creating the Countdown Goodie! You’ve been doing a really good job here and we'd like you to help our community tackle the next big challenge - our mission to become the best search engine for programmers: https://duck.co/blog/post/297/help-for-programmers.

When you're ready to move on, please: 

- Create a new Instant Answer to increase our Programming IA coverage

OR

- Test live programming Instant Answers (or those in development) for bugs and suggest improvements 

Detailed instructions explaining how you can help are available at the top of this [page](https://github.com/duckduckgo/duckduckgo/wiki/Programming-IA-Coverage).

Thanks again!

rasikapohankar:
@moollaza Sure!

rasikapohankar:
@GuiltyDolphin: Can you please review?

lib/DDG/Goodie/Countdown.pm
```diff
@@ -0,0 +1,85 @@
...
+zci answer_type => 'countdown';
+
+zci is_cached => 1;
+
+triggers any => 'countdown to','time until','how long until';
+
+my $datestring_regex = datestring_regex();
+my $relative_dates_regex = relative_dates_regex();
+my $time_regex = time_12h_regex();
+my $output_string;
```
GuiltyDolphin: This shouldn't be global. Move it into a sub and pass it as an argument.

t/Countdown.t
```diff
@@ -0,0 +1,51 @@
...
+
+ddg_goodie_test(
+    [qw( DDG::Goodie::Countdown )],
+    
+    'countdown to 10:00:00 am 26th July' => build_test("10:00:00 am 26th July", 8792938000000000, "26 Jul 2016 10:00:00 EDT"),
+    'countdown to 10:00:00 am' => build_test("10:00:00 am", 66538000000000, "16 Apr 2016 10:00:00 EDT"),
+    'time until 1st June 2016' => build_test("1st June 2016", 4004938000000000, "01 Jun 2016 00:00:00 EDT"),
+    'how long until 31st December 2016' => build_test("31st December 2016", 22411738000000000, "31 Dec 2016 00:00:00 EST"),
+    'how long until 1st May 12:00:00 pm' => build_test("1st May 12:00:00 pm", 1369738000000000, "01 May 2016 12:00:00 EDT"),
+    'how long until 01:00:00 pm tomorrow' => build_test("01:00:00 pm tomorrow", 163738000000000, "17 Apr 2016 13:00:00 EDT"),
```
GuiltyDolphin: Tomorrow would be the 16th, not the 17th.

rasikapohankar: Hey @GuiltyDolphin, what I've done is if the entered time is greater than the current time then it moves to the next day. I'll add a condition for when the query is tomorrow so it won't add one more day. In the case of today, if it is 2 pm and the query is countdown to 1 pm today then also it will move to the next day. Will this seem incorrect to the user?

GuiltyDolphin: Cases:

- `countdown to 1 pm` (current time `11 am`): Count down to `1 pm` today.
- `countdown to 1 pm` (current time `2 pm`): Count down to `1 pm` tomorrow.
- `countdown to DATE` (date is in the past): ***Do not trigger***.
- `countdown to DATE` (date is in the future): Countdown to that date.

rasikapohankar: I was just wondering if the 2nd case output will confuse the user?

GuiltyDolphin: @rasikapohankar I would say having it roll over to the next day is fine (but only when no day is specified).

You could always reject the case for now, and open it up later if people request it.

rasikapohankar:
@moollaza @GuiltyDolphin This is not a giving a result on the beta server and the IA page says 'not yet installed on beta'. What might be wrong?

moollaza:
@rasikapohankar it's now on Beta: https://beta.duckduckgo.com/?q=countdown+to+26th+July

Looks like it's broken though....

rasikapohankar:
@moollaza It is working when I try it using duckpan server..

moollaza:
@rasikapohankar there's no guarantee that `getReferences()` is run after your `DDG.require()` callback function. We need to make a request to load Moment.js first which takes time, but execution continues, so I suspect your IA hasn't been built yet by the time `getReferences()` is called, so it's returning empty selectors.

Move that function call inside `DDG.require()` and you should be good.

Does that make sense?

share/goodie/countdown/countdown.js
```diff
@@ -0,0 +1,181 @@
...
+                }                                                
+                hasShown = true;                          
+                DDG.require('moment.js', function() {                
+                    duration = getCountdown(moment.duration(initialDifference));                    
+                    setInterval(function() { 
+                        duration = getCountdown(duration);                         
+                    }, 1000);
+                });                
+                $(".name_input").html("Counting down to "+countdown_to+",");
+                getReferences();
```
moollaza: Move this up into your anonymous callback fn, this will ensure your IA has been built before we try to grab the various DOM elements

share/goodie/countdown/countdown.js
```diff
@@ -0,0 +1,181 @@
...
+            difference = d.subtract(1, 's');            
+        } else {
+            stopped = true;
+            endCountdown();     
+        }
+        return difference;
+    }
+    
+    function getReferences() {
+        $countdownContainer = $(".zci__body").find(".countdown_container");
```
moollaza: Let's make this more specific, use `.zci--countdown` instead of `zci__body`

rasikapohankar:
@moollaza Thanks a lot for pointing that out. I guess when I was trying it locally, `require` always entered the callback after `getReferences` had finished executing and so it found the css selectors, and I never realised that `getReferences` needed to execute first. But on the beta server, `require` was finishing first and `getReferences` was never called upto that point! I moved the call to `getReferences` before `DDG.require` instead of inside the callback, to see what happens. Is that okay?

rasikapohankar:
Does this get redeployed automatically like the cheat sheets?

moollaza:
> Does this get redeployed automatically like the cheat sheets?

@rasikapohankar not yet, we need to manually verify the code is safe to go on Beta and doesn't contain malicious stuff

moollaza:
Deploying it now...

rasikapohankar:
@moollaza @GuiltyDolphin I have made some more changes and tried to add css rules for mobile. Can you please review and deploy it again?

rasikapohankar:
Normal view,
d3d941bf-4150-4faa-b17c-d0b8100b869d

and on mobile, 
rsz_screenshot_from_2016-05-26_023318

Hid the progress circle for small screen, because it was crowded. Is this okay?
@moollaza

GuiltyDolphin:
@rasikapohankar I'm assuming the clock is static? If so, I'm not sure we should keep it - it might give the wrong impression.

I'm not sure it fits in with the general DDG style either.

rasikapohankar:
@GuiltyDolphin Yes the clock is static, the hands don't move. I've given it fade toggle, so it keeps toggling to show that the countdown is progressing. I added it because if the duration is longer the progress circle fills very slowly and it looks empty. I was looking for an hourglass icon, but haven't found a good one yet. I think an hourglass will look better?
It fits if the year digits grow, I've tried with a date upto the year 4999, after which it breaks on duckpan server saying the date is far off in the future! :smile:

rasikapohankar:
Hourglass,
1b4b592e-282a-402e-bb68-56aab7344ee2

1c7a521b-0716-4e73-8f01-56c5df867f62
@moollaza @GuiltyDolphin What do you think? Will this be alright?

rasikapohankar:
@moollaza @GuiltyDolphin Could you please check?

GuiltyDolphin:
@rasikapohankar As I said before - I don't really think it fits in with the DDG style; maybe we could have a small graphic that spins round (similar to buffering on various sites) or something?

Thoughts @moollaza ?

moollaza:
I agree with @GuiltyDolphin about the hourglass and progress meter. Not sure it's necessary. I suspect our design team will agree. I think we should remove it and leave the focus on the numbers. I would also switch this to a detail templates, instead of tile template, so that the answer isn't in it's own box.

Let me ping the designers for their opinions though...

moollaza:
@rasikapohankar I've asked the design team for their input (/cc @abeyang ) and told them my proposal would be something much simpler:

countdown_to_10_00_am_26th_july_at_duckduckgo


The HTML and CSS for this approach is a lot simpler too:

```css
.zci--countdown .time-unit,
.zci--countdown .separator {
    text-align: center;
    display: inline-block;
}

.zci--countdown .time-unit {
    min-width: 3em;
}

.zci--countdown .separator {
    padding: 0 0.4em 0;
    vertical-align: top;
}

.zci--countdown .number,
.zci--countdown .separator {
    font-size: 2.5em;
}
```

```html
<div class="countdown_container seventy--screen-m whole--screen-s">
	<span class="tx-clr--slate time-unit is-hidden">
		<div class="number years">00</div>
		<div class="unit">Years</div>
	</span>
	<span class="tx-clr--slate separator y_separator is-hidden">:</span>
	<span class="tx-clr--slate time-unit is-hidden">
		<div class="number months">00</div>
		<div class="unit">Months</div>
	</span>
	<span class="tx-clr--slate separator m_separator is-hidden">:</span>
	<span class="tx-clr--slate time-unit">
		<div class="number days">00</div>
		<div class="unit">Days</div>
	</span>
	<span class="tx-clr--slate separator">:</span>
	<span class="tx-clr--slate time-unit">
		<div class="number hours">00</div>
		<div class="unit">Hours</div>
	</span>
	<span class="tx-clr--slate separator">:</span>
	<span class="tx-clr--slate time-unit">
		<div class="number minutes">00</div>
		<div class="unit">Minutes</div>
	</span>
	<span class="tx-clr--slate separator">:</span>
	<span class="tx-clr--slate time-unit">
		<div class="number seconds">00</div>
		<div class="unit">Seconds</div>
	</span>
</div>
```

I've also set `item: false` to force a detail template and I've set the title:

```js
data: {
    title: "Counting down to " + countdown_to
},
```

rasikapohankar:
@moollaza Thanks! So we wait for input on the design? I could throw away the progress circle, what you have done looks so much better :+1: Just curious, are `seventy--screen-m` and `whole--screen-s` classes from the DDG style sheet?

moollaza:
@rasikapohankar yes please implement the design I have here and we can get this merged. The design can provide suggestions and you can implement them for V2 of this IA. What I've proposed is good enough to get this merged 👍 

> Just curious, are seventy--screen-m and whole--screen-s classes from the DDG style sheet?

Yes! We have a small CSS grid framework we use internally. All our CSS classes can be seen here: https://duckduckgo.com/styleguide

rasikapohankar:
@moollaza I have made the changes you suggested. Can you please check?

moollaza:
Deploying to beta...

lib/DDG/Goodie/Countdown.pm
```diff
@@ -0,0 +1,139 @@
...
+    #add 0s if input is not of the form hr:min:sec
+    for(my $len = @time; $len < 3; $len++) {
+        push (@time, 00);
+    }
+    push (@time, $meridiem);
+    return @time;
+}
+
+sub get_initial_difference {
+    my $then,my $date,my $time,my $day_of_week, my $days_to_add = 1;
```
GuiltyDolphin: You can do `my ($then, $date, $time, $day_of_week, $days_to_add)`.

lib/DDG/Goodie/Countdown.pm
```diff
@@ -0,0 +1,139 @@
...
+    wed => 3,
+    thu => 4,
+    fri => 5,
+    sat => 6,
+    sun => 7    
+);
+
+sub get_regex() {
+    my $time_12h = qr#(?:(?:0?[1-9])|(?:1[012]))(:[0-5][0-9])?(:[0-5][0-9])?\s?(?:am|pm)#i;
+    return (datestring_regex(), relative_dates_regex(), $time_12h, full_day_of_week_regex()."|".short_day_of_week_regex());
```
GuiltyDolphin: What is this trying to achieve? We should be doing all the parsing via the Role (some forms may not yet be supported - but that should be raised as an issue against the Role rather than trying to _roll_ your own).

moollaza:
@rasikapohankar nicely done! This looks great

GuiltyDolphin:
@rasikapohankar It looks like you're doing a lot of unnecesarry work in the Perl - why not just use the `Dates` GoodieRole to parse a date, then subtract the current epoch from that? You can then just return if the value is negative (meaning the date has already passed).

I know we've spoken about certain ambiguities with date parsing - but this should really _all_ be handled by the Role; we've got a bunch of new parsing capabilities coming in with https://github.com/duckduckgo/zeroclickinfo-goodies/pull/2810 - which should hopefully also allow better control over parsing the past/present/future for the user (Goodie).

lib/DDG/Goodie/Countdown.pm
```diff
@@ -0,0 +1,139 @@
...
+
+
+# Handle statement
+handle remainder => sub {    
+
+    my @output = get_initial_difference($_);            
+
+    my $initialDifference = $output[0];    
+
+    return unless $initialDifference;
```
GuiltyDolphin: Basically I think this should be...

```Perl
my $remainder = $_;

my $date = parse_datestring_to_date($remainder) or return;

my $current = parse_datestring_to_date('now');

my $diff = $date->epoch - $current->epoch;

return ...;
```

That should really be all you need.

lib/DDG/Goodie/Countdown.pm
```diff
@@ -0,0 +1,139 @@
...
+
+use DateTime;
+
+use strict;
+
+zci answer_type => 'countdown';
+
+zci is_cached => 1;
+
+triggers any => 'countdown to','time until','how long until';
```
moollaza: should be `start` unless we expect these to appear at the end or middle of queries?

lib/DDG/Goodie/Countdown.pm
```diff
@@ -0,0 +1,139 @@
...
+sub get_regex() {
+    my $time_12h = qr#(?:(?:0?[1-9])|(?:1[012]))(:[0-5][0-9])?(:[0-5][0-9])?\s?(?:am|pm)#i;
+    return (datestring_regex(), relative_dates_regex(), $time_12h, full_day_of_week_regex()."|".short_day_of_week_regex());
+}
+
+sub split_time {
+    my $time_input = shift(@_);
+
+    #remove leading and trailing whitespaces
+    $time_input =~ s/^\s+|\s+$//g;
```
moollaza: I would `use Text::Trim;` then `trim($time_input)`

lib/DDG/Goodie/Countdown.pm
mintsoft: maybe `startend` so it'd handle Yoda queries; `15:00 how long until?`

lib/DDG/Goodie/Countdown.pm
rasikapohankar: Hey @GuiltyDolphin, `time_12h_regex` is `qr#(?:(?:0[1-9])|(?:1[012])):[0-5][0-9]:[0-5][0-9]\s?(?:am|pm)#i` currently so it matches time only of the form, for example, 10:00:00 am/pm and not 10 am or 10:00 am. I thought maybe this is a little tedious to type and won't match if I only want to say `countdown to 10 am`. So I changed it a little. Should I make an issue for this?

rasikapohankar:
Hey @moollaza @GuiltyDolphin Thanks for reviewing!
@GuiltyDolphin Wow, #2810 has a lot of changes! I commented the call to everything I've written in the perl file and used only this from your comment above to try,

   `my $remainder = $_;
    my $date = parse_datestring_to_date($remainder) or return;
    my $current = parse_datestring_to_date('now');
    my $diff = $date->epoch - $current->epoch;
    return if $diff <= 0;
    return $diff,
        structured_answer => {
            data => {
                remainder => $_,
                difference => $diff,
                countdown_to => date_output_string($date,1)
            },
            templates => {
                group => "text",
            }
        };`       

It gives an answer for these type of queries directly: 
1. countdown to 26th July
2. countdown to 26th July 2016
3. countdown to next day, next week, etc. :+1: 

These type of queries don't get an answer at the moment:
1. countdown to 3 pm
2. countdown to 03:00:00 pm
3. countdown to 3pm 
4. countdown to 26th July 3 pm
5. countdown to 26th July 02:00 pm
6. countdown to 26th July 02:00:00 pm and vice versa with time before the date
7. countdown to Monday, countdown to next Monday; and other days of the week
8. countdown to Monday 10 pm, countdown to next Monday 10 am

I had written everything to handle these cases(except 3pm, this doesn't work because I had assumed a space between 3 and pm), as I didn't find anything for calculation based on time along with date in the current Role. So all these will be handled after #2810? I'm sorry if I'm only repeating some conversation from that PR, I didn't go through everything there properly :sweat_smile:

lib/DDG/Goodie/Countdown.pm
rasikapohankar: I'll use that :+1:

lib/DDG/Goodie/Countdown.pm
GuiltyDolphin: @rasikapohankar Yeah - this is to do with Role functionality so we should just raise it as an issue against the Role rather than try and 'modify' the functionality here.

GuiltyDolphin:
@rasikapohankar I would recommend creating an issue with a list of query forms you think should match, then we can see which ones should/shouldn't match and update the role accordingly.

It looks like #2810 isn't handling generic time queries at the moment, this is something we can add.

Lets do as I suggested before, remove all the Perl apart from the suggested handler and update tests appropriately - we can talk parsing in an issue :wink:

rasikapohankar:
@GuiltyDolphin I have created an issue here, #3188.

GuiltyDolphin:
@rasikapohankar Cool, thanks.

The changes will be dealt with on #2810, but in the meantime would you mind updating the Perl? When the changes to the Role are complete they should automatically take effect here too.

rasikapohankar:
Hey @GuiltyDolphin, I have updated the perl and test files. Can you check once? I guess the example queries on the IA page will have to be changed too?

t/Countdown.t
```diff
@@ -0,0 +1,50 @@
...
+    'how long until 31st December 2016'   => build_test("31st December 2016", 22426138, "31 Dec 2016 00:00:00 EST"),
+    'countdown to tomorrow'               => build_test("tomorrow", 86400, "16 Apr 2016 11:31:02 EDT"),
+    
+    #invalid
+    'how long until 01:00:00 am yesterday' => undef,
+    'countdown to 10:00:00 am 26th July'   => undef,
+    'countdown to 10:00:00 am'             => undef,
+    'time until 1st May 12:00:00 pm'       => undef,
+    'how long until 01:00:00 pm tomorrow'  => undef,
+    'how long until 01:00:00 am today'     => undef,
```
GuiltyDolphin: Lets just move the 2nd, 3rd, 4th, 5th tests and comment them out, with a note saying they *should* trigger in the future and should be uncommented when the role is updated (i.e., add the original results back and comment out).

GuiltyDolphin:
@rasikapohankar Looking good! Just left a comment about the tests, but apart from that :+1:

rasikapohankar:
@GuiltyDolphin Modified the test file..
I am not able to update the example queries as I had not created the IA page. I think @moollaza had updated them before. Is it possible for me to update the queries?

rasikapohankar:
@moollaza @GuiltyDolphin One other small thing (not related to this), I was going through the DDG Style Guide https://duckduckgo.com/styleguide and noticed a _really_ small typo in the 'Responsiveness' section (the spelling for usually is incorrect - usueally), its not very noticeable but I thought I should let you know. :smile:

GuiltyDolphin:
@rasikapohankar Thanks for pointing that out - feel free to make an issue and/or pull request on https://github.com/duckduckgo/duckduckgo-publisher

And you might have to wait until the PR is merged before you can update the page - it requires contributer privileges or higher.

rasikapohankar:
Thanks @GuiltyDolphin, I didn't know about the https://github.com/duckduckgo/duckduckgo-publisher repository. Found the file that has the typo, I'll make a pull request.

lib/DDG/Goodie/Countdown.pm
```diff
@@ -0,0 +1,43 @@
...
+zci is_cached => 1;
+
+triggers startend => 'countdown to','time until','how long until';
+
+# Handle statement
+handle remainder => sub {    
+
+    my $remainder = $_;
+
+    my $date = parse_datestring_to_date($remainder) or return;
```
moollaza: nit: would write as `return unless my $date = ...`

moollaza:
@rasikapohankar Thanks! This LGTM :+1:

Excellent Work! 💯

daxtheduck:
@rasikapohankar this is now merged! It should be live on DuckDuckGo within one week (or during the next release) and we'll be sure to notify you when it happens.

lib/DDG/Goodie/Countdown.pm
GuiltyDolphin: Prefs ;)

PR #2810, opened 2016-03-28
Dates GoodieRole: Use more YAML

GuiltyDolphin:
@mintsoft Here's a PR for the changes.


daxtheduck:
Metadata placeholder post
            
---
            
**How to find your IA page**
- New Instant Answer
              
- Existing Instant Answer

t/00-roles.t
```diff
@@ -198,11 +199,13 @@ subtest 'Dates' => sub {
             },
         );
 
+        set_fixed_time('2015-01-01T00:00:00Z');
```
mintsoft: Hmm none of these times should be temporal; why is the `set_fixed_time` needed?

GuiltyDolphin: Uhm... I don't remember adding that... Works fine without it though so I'll remove it.

GuiltyDolphin:
@mintsoft I've added in location support (I think).

@moollaza Could we get this deployed to https://moollaza.duckduckgo.com or similar? I'd like to test out the location preferences without it affecting all the date IAs on beta.

GuiltyDolphin:
@moollaza Which packages are the new dependencies? `Module::Data`?

GuiltyDolphin:
@mintsoft Looks like location settings are working:

`DDG_TEST_LOCATION=de duckpan server DateMath`:

screenshot from 2016-04-02 10 07 10

(Only valid in DMY format)

screenshot from 2016-04-02 10 10 46


`DDG_TEST_LOCATION=us duckpan server DateMath`:

screenshot from 2016-04-02 10 07 36

(Only valid in DMY format)

screenshot from 2016-04-02 10 10 31

---

So it works fine with location. Unfortunately it seems this is based on IP, so they can't easily change their preferences...

mintsoft:
@GuiltyDolphin some things I just noticed when making Timediff; "today" is including the current time (so `2016-04-09T17:04:00` as I'm writing this, however it should be defined as `2016-04-09T00:00:00`); this is true of `tomorrow` and `yesterday` too, presumably it adds days from `today` to get to them (I can't actually remember how that's implemented).

Also I think we should add `now` (shamelessly stolen from PostgreSQL) which would be the current time

GuiltyDolphin:
@mintsoft It treats `now` and `today` as the same. I could split it into two groups: the 'day', and the 'time': So we could have `today`, `this day` etc. represent the date w/o time, and `now`, `time now` etc. As the date w/ the time.

Sound good?

mintsoft:
@GuiltyDolphin Yeah that's exactly the behaviour I want!

GuiltyDolphin:
@mintsoft No, 5:37 pm (yes, midnight).

I'm currently working on some more locale stuff, but I'll stick that on my to-do.

GuiltyDolphin:
@mintsoft I'm pretty happy with this - there are more changes that could be made, but I'd prefer to get this sorted so we can make any further changes in separate PRs.

I noticed some issues with some of the IA tests (notably, testing the Role rather than the IA as you've mentioned); so I might be making some issues (after this PR - too many merge conflicts otherwise)!

@moollaza Can we get this deployed?

lib/DDG/Goodie/CalendarConversion.pm
```diff
@@ -27,7 +25,7 @@ sub format_date {
 
 handle query_lc => sub {
     return unless my ($datestring, $input_calendar, $output_calendar) = $_ =~ /^
-            ($datestring_regex)\s+
+            (?<date>.+?)\s+
```
mintsoft: @GuiltyDolphin Is there a particular reason you've removed the "looks like a date" regex from here. Is it part of a attempt to remove the functionality or just because the parser will handle invalid formats anyway so why not?

GuiltyDolphin: @mintsoft We no longer provide regexes - it is not required. All matching should be done through the interface provided by the `Parser` package.

We shouldn't run into any issues doing this because the query *must* be well-formed for any of these 'my query should look like this' regexes to parse.

Another reason is that we no longer *use* the descriptive regexes - another package deals with that (and does a pretty good job).

mintsoft: Hmm what if we want to trigger on dates? That was essentially why the DatesRole was made in the first place :)

GuiltyDolphin: @mintsoft Oh God - I think we'd have some issues if we used date regexes for triggers; just imagine having to check through a thousand-line regex on every query for several IAs!

GuiltyDolphin: @mintsoft We also have the power of `extract_dates_from_string`, which extracts dates where possible, and returns the *remainder* - allowing you to check additional structure of the query and/or whether or not there *is* a remainder.

mintsoft: so what if the string is of form "there was one at 01-jan 2015 and then another at 03-feb-2016; I was unsure"; what's the 'remainder' there?

GuiltyDolphin: "there was one at  and then another at  ; I was unsure" I'd imagine.

mintsoft: The use case I'm thinking is when you know that there are dates within a string; but you're interested in ones at certain positions; having the regex public gives the consumer the ability to specify their exact position rather than depending on the output of a generic parsinator (tm) which may return an unknown number of instances. However this is probably academic as I'm not sure anyone other than me has ever actually used the regex functionality!

GuiltyDolphin: @mintsoft Then `.+?` should be fine. It will be bounded by the expected structure of the regex.

GuiltyDolphin: Also, in updating the Goodies I didn't notice any cases where position was required in away not manageable by the new system; hence being able to update everything relatively easily.

mintsoft: Hmm that may be true; I personally wouldn't lean to using `.+?` for it. I prefer to match the format of things that I know if I know it; hence my fetish for the aforementioned regexes! However if they're useless then they're useless :) If a situation arises that would use them then we can always make them public later on

GuiltyDolphin: @mintsoft I was thinking of integrating this with WhatIs (#2644) in the future - so you'd be able to specify that say, the 'primary' value should be a date; which I think is much more powerful than directly using the regexes yourself.

GuiltyDolphin: @mintsoft I'm not certain, but I noticed the parsing was going pretty slowly at times and profiling indicates that it *may* be something to do with DateTime::Format::Natural; I'll look into it further but we may end up needing our own regexes anyway!

GuiltyDolphin:
@mintsoft I've opted to go back to using custom regexes due to speed and flexibility concerns (DateTime::Format::Natural had no support for anything other than `en`).

I've made a couple of other changes too, want to review?

lib/DDG/Goodie/CalendarConversion.pm
GuiltyDolphin: @mintsoft Enjoy the regexes.

I've switched to using a Regexp::Common-ish method for regexes, so you can use `$RE{date}{...}` or `$RE{time}{...}` to access regexes. You can also access matches/captures through `{-keep}` (standard) and `{-names}` (custom).

Hopefully I'll be able to convert the descriptive and recursive regexes over too.

mintsoft:
@GuiltyDolphin some nice conflicts here

nalinbhardwaj:
I think it's a good idea if common dates/holidays are also added.

For example, parsing queries such as `Christmas 2018` or `new years` should also have get parsed to `25th December, 2018`, `1st January {current_year+1}` etc.

moollaza:
What's the status here? I see there are some conflicts now. Is it otherwise finished? Can we break out the IA changes into separate PRs?

Issue #2771, opened 2016-03-18
Braille: bad triggers


zachthompson:
This was taken offline I believe because it triggers on stuff like "braille alphabet", "how long to learn braille", etc.

Could we modify the triggers to allow something like "braille: ", "to braille", "in braille", etc., to get it back online?

------
IA Page:  http://duck.co/ia/view/braille 
[Maintainer](http://docs.duckduckhack.com/maintaining/guidelines.html): @mintsoft

Issue #1171, opened 2015-06-02
Convert simple structured_answer Goodies to full templates


moollaza:
There are many Goodies which use the simpler form of Goodie templating ("auto templating") which works when the Goodie returns a 3-property `structured_anwer` which contains the `input`, `operation` and `result`.

Now that we have fully fledged templating support, many of these Goodies, especially the ones that return HTML should be converted to use the "full" template system. In many cases switching to the Text template group will likely suffice, but in others we may benefit from using other built-in templates, which will allow for more complex results.

BTW, converting many of these should be fairly simple and easy for newcomers!

mintsoft:
@moollaza do you think it'd be worth having an example of a before and after from a simple one?

moollaza:
Closing. Nearly all of them have been converted now. There are in-progress PR's to replace the remaining Goodies.

PR #2811, opened 2016-03-29
Braille: fix for bad triggers #2771

lamanh:
Fixes: #2771. Use specified triggers to avoid bad triggers. Also add types to show the result is in Braille or ASCII/Unicode. Removed `html_enc()` for `structured_answer->input` since it will show raw braille code instead of braille string.

https://duck.co/ia/view/braille

### DIFF
--- a/lib/DDG/Goodie/Braille.pm
+++ b/lib/DDG/Goodie/Braille.pm
@@ -34,7 +34,7 @@ handle query_raw => sub {
       structured_answer => {
           data => {
               title    => $response,
-              subtitle => 'Translate "' . $query . '" to ' . $type,
+              subtitle => 'Braille translation: ' . html_enc($query),
           },
           templates => {
               group => 'text',
--- a/lib/DDG/Goodie/Braille.pm
+++ b/lib/DDG/Goodie/Braille.pm
@@ -7,7 +7,7 @@ use DDG::Goodie;
 use Convert::Braille;
 use utf8;
 
-triggers query_raw => qr/\p{Braille}|braille/i;
+triggers query_raw => qr/\p{Braille}|( in|to){1} braille$|^braille:/i;
 
 zci is_cached => 1;
 
@@ -16,20 +16,24 @@ my $braille_space = '⠀';    # the braille unicode space (U+2800)
 handle query_raw => sub {
 
     my $query = $_;
-    $query =~ s/translate to braille |( in)? braille$|^braille //;
+    $query =~ s/( in|to){1} braille$|^braille: //;
     return unless $query;
 
     my $result;
+    my $type;
 
     if ($query =~ /\p{Braille}/) {
         $result = join(" ", map { lc(brailleDotsToAscii($_)) } split(/$braille_space/, $query));
+        $type = "Ascii/Unicode";
     } else {
         $result = join($braille_space, map { brailleAsciiToUnicode(uc $_) } split(/\s/, $query));
+        $type = "Braille";
     }
 
-    return $result . ' (Braille)',
+    return answer => $result,
+      answer_type => $type,
       structured_answer => {
-        input     => [html_enc($query)],
+        input     => [$query],
         operation => 'Braille translation',
         result    => html_enc($result),
       };
--- a/lib/DDG/Goodie/Braille.pm
+++ b/lib/DDG/Goodie/Braille.pm
@@ -7,7 +7,7 @@ use DDG::Goodie;
 use Convert::Braille;
 use utf8;
 
-triggers query_raw => qr/\p{Braille}|( in|to){1} braille$|^braille:/i;
+triggers query_raw => qr/\p{Braille}|( in| to){1} braille$|^braille:/i;
 
 zci is_cached => 1;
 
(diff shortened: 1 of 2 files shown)


daxtheduck:
@lamanh Hey!
        
Thanks!

daxtheduck:
Metadata placeholder post
            
---
            
**How to find your IA page**
- New Instant Answer
              
- Existing Instant Answer

lamanh:
I look into the build fail log, seems like the test is based on previous design. Should I keep the same or we could update the test as well?

Thanks!
Eric

moollaza:
> Should I keep the same or we could update the test as well?

@lamanh thanks for tackling this!

We definitely need to update the tests as well. However, we are in the process of upgrading all our IAs to use the full templates as outlined in #1171 

Would you be able to update the `structured_answer` to the new format and also update the tests?

lib/DDG/Goodie/Braille.pm
```diff
@@ -16,20 +16,24 @@ my $braille_space = '⠀';    # the braille unicode space (U+2800)
 handle query_raw => sub {
 
     my $query = $_;
-    $query =~ s/translate to braille |( in)? braille$|^braille //;
+    $query =~ s/( in|to){1} braille$|^braille: //;
```
GuiltyDolphin: Why the `{1}`? A single occurrence is implicit.

lamanh: It's based on #2771, original `?` will be triggered when query like "how long to learn braille" or "history of braille" etc. So change it to `{1}` to be less sensitive. However, I'm not sure if this is the best solution since I'm not a native English speaker. Please let me know if there's a better way.

lamanh:
I'm getting error from DuckPAN server on Codio with this new template. The same error happen with https://duck.co/ia/view/anagram. Do I need to update my Codio environment?

Error:
**Use of uninitialized value in concatenation (.) or string at /home/codio/perl5/perls/perl-5.18.2/lib/site_perl/5.18.2/App/DuckPAN/Web.pm line 442**

zachthompson:
@lamanh try running `duckpan update`

lib/DDG/Goodie/Braille.pm
GuiltyDolphin: Huh? I'm just wondering why you've done `( in|to){1}` rather than `( in|to)` - `{1}` just means 'once', which is the same as not having it there at all.

lamanh:
@zachthompson Thanks! it's working now.

I just follow the docs to setup my Codio environment few days ago, so maybe the doc or the default Codio environment could be update to latest to avoid this error for a new user like me?

moollaza:
Deploying this to beta...

moollaza:
> Codio environment could be update to latest to avoid this error for a new user like me?

Sorry about that, I've now updated the DuckDuckGo Codio box so new forkers won't have issues.

moollaza:
I just realized this won't work on Beta right now because it needs to be updated to be able to handle Goodies that don't have a `name` and `id`. Will test locally..

lib/DDG/Goodie/Braille.pm
```diff
@@ -16,22 +16,29 @@ my $braille_space = '⠀';    # the braille unicode space (U+2800)
...
 
-    return $result . ' (Braille)',
+    return $response,
       structured_answer => {
-        input     => [html_enc($query)],
-        operation => 'Braille translation',
-        result    => html_enc($result),
+          data => {
+              title    => $response,
+              subtitle => 'Translate "' . $query . '" to ' . $type,
```
moollaza: Typically this should be of the form `<operation>: input`

e.g `Braille Translation: $query`

moollaza: We also still need to `html_enc()` the `$query` for now until we have a better fix in place

lamanh: @moollaza, with `html_enc()`, the output would be like:
image

Which I personally don't think it's correct answer, answer should show in braille format instead of raw code.

Without `html_enc()`, the output make more sense:
image
 
But I understand that `html_enc()` is needed for now. So, maybe this function should keep offline until there's a better fix.

Please let me know the next step for this PR. Thanks!

zachthompson: It just needs to be done on the input, not the output.

moollaza: @lamanh Whoops, my mistake, just needed for the `subtitle`

jdorweiler:
@moollaza what needs to be updated on beta to get this working?

moollaza:
@jdorweiler it should be fine now -- we needed to update DDG but it's since been updated 👍

jdorweiler:
It's working on beta now.  
https://beta.duckduckgo.com/?q=hello+in+braille&ia=answer

jdorweiler:
@mintsoft want to check this out?

mintsoft:
@jdorweiler @lamanh Apologies; I missed this, checking it out now

lib/DDG/Goodie/Braille.pm
```diff
@@ -16,22 +16,29 @@ my $braille_space = '⠀';    # the braille unicode space (U+2800)
...
 
-    return $result . ' (Braille)',
+    return $response,
       structured_answer => {
-        input     => [html_enc($query)],
-        operation => 'Braille translation',
-        result    => html_enc($result),
+          data => {
+              title    => $response,
+              subtitle => 'Braille translation: ' . html_enc($query),
```
mintsoft: This should not be required as it's using the template? @moollaza can you confirm?

mintsoft:
@lamanh we are getting this in the response when asked to go from braille -> english

beta_ddg

I think it's due to the `html_enc` in the output; however I don't think this is required as long as it's using a full template but I'd like some staff to confirm

lib/DDG/Goodie/Braille.pm
moollaza: It's still required for IA's with templates. Not because the template doesn't escape the content, but because it can break the HTML because we pass the data along via JSON

mintsoft: Ah right; I guess we'll fix that in the IA system later on?

mintsoft:
@moollaza this LGTM then

moollaza:
LGTM 👍 

@lamanh I've updated the example query because it no longer triggers the Instant Answer -- we should probably still support queries of the format "translate to braille X" and  "translate X to braille" though.

PR #2577, opened 2016-02-23
New 2048 Game Goodie

moollaza:
Continuing PR #1885 

Partially addresses #1637 -- _still some work to be done!_

-----
IA Page: https://duck.co/ia/view/game_2048

PR #2825, opened 2016-04-03
2048, continuing from #2577

Roysten:
My work on #2577:
Done:
- [X] Code cleanup
- [X] Added animations for moving tiles
- [X] Disable on mobile
- [x] Animations for merging of tiles (simple crossfade)
- [ ]  Improve performance? I don't think this is necessary, could someone verify?
- [x] More code cleanup
- [x] Test in all browsers
- [x] Fix IE :(

---

IA Page: https://duck.co/ia/view/game2048


daxtheduck:
Metadata placeholder post
            
---
            
**How to find your IA page**
- New Instant Answer
              
- Existing Instant Answer

Roysten:
@moollaza I have a hard time trying to figure out how to disable this Goodie if it is requested on mobile. Can I disable it from the backend (in Perl)? I found that I can use the `is_mobile` javascript function, but that doesn't seem ideal...
Is there any progress on a way to process input events from mobile devices?

GuiltyDolphin:
@roysten I'm pretty sure we can't do that from the Perl (it doesn't know any front end information) - using `is_mobile` should be fine - as long as you do the check early.

moollaza:
@Roysten thanks for the awesome work!

> Can I disable it from the backend (in Perl)?

Unfortunately not yet, user agent information isn't exposed to Instant Answers so you'll have to block it in the JS for now.

Roysten:
@moollaza Merge animations have been added. I tested it in Chromium, Firefox 45, Internet Explorer 11, 9 and Edge. Internet Explorer 11 and Edge had some problems, but I was able to come up with two workarounds which solve the problems.

The animations don't work in IE9, since this version does not support transitions yet. I'm assuming this isn't a big deal (since you were ok with releasing it without any animations anyway :smiley:)?

moollaza:
> The animations don't work in IE9, since this version does not support transitions yet. I'm assuming this isn't a big deal (since you were ok with releasing it without any animations anyway )?

It would be nice to support IE 9 at some point because I can see this IA being popular among users of all browsers, and we still have many IE users. If it's not possible, then there's not much we can do as long as it gracefully degrades.

@Roysten It looks likes there's some conflicts on the PR -- can you please fix that and then I'll gladly get this onto Beta for final testing!

share/goodie/game2048/game2048.js
```diff
@@ -1,249 +1,345 @@
...
-    function print_area() {
-        cells.each(function(index) {
-            var val = area[index].val;
-            if(val === 0) {
-                $(this).html("").attr("class", "boxtile val-");
-            } else {
-                $(this).html(val).attr("class", "boxtile val-" + val);
+        var addition = "<div class='score-addition'>+" + points + "</div>";
+        $('.game2048__points_addition').html(addition);
+        $('.game2048__points').text(score);
```
moollaza: @Roysten we should store selectors in variables so we don't repeat these DOM lookups:

`var $game2048_points = $('.game2048_points');`

share/goodie/game2048/game2048.js
```diff
@@ -1,249 +1,345 @@
...
+
+    //Update the board
+    function update_tiles() {
+        for(var i = 0; i < TILE_COUNT; ++i) {
+            var pos = index_to_rc(i);
+            var tile = tiles[i];
+            
+            if("undefined" !== typeof(tile.div) && tile.val > 0) {
+                var translate_string = gen_translate_string(pos.row, pos.col);
+                tile.div .html(tile.val)
```
moollaza: looks like there's a rogue space here? `tile.div.html`?

share/goodie/game2048/game2048.js
```diff
@@ -1,249 +1,345 @@
...
+                    .attr("class", "boxtile val-" + tile.val)
+                    .css({
```
moollaza: Why do we need to apply this same CSS class throughout the code in different places? Can we instead add/remove a class?

Or simply set it once?

share/goodie/game2048/game2048.js
```diff
@@ -1,249 +1,345 @@
...
+        var rand_val = Math.floor(Math.random() * 11) < 2 ? 4 : 2;
+
+        tiles[rand_tile].div = create_tile_div();
+        tiles[rand_tile].val = rand_val;
+    }
+
+    function create_tile_div() {
+        //Display:none is an IE workaround
+        //Otherwise it will always move from tile (0,0) which looks silly
+        var div = $("<div style=\"display: none\"></div>");
```
moollaza: I would argue this should all be done the jQuery way: `var div = $("<div>").hide()`

share/goodie/game2048/game2048.js
```diff
@@ -1,249 +1,345 @@
...
+    }
 
-        return false;
+    function is_game_over() {
+        return has_won() || has_lost();
+    }
+
+    // This function shows game over message
+    function game_over_message(game_won) {
+        var result_msg = $('#game2048__area_container .game2048__message p');
```
moollaza: Again, we should be storing these once in a var and using them in the function.

share/goodie/game2048/game2048.js
GuiltyDolphin: Yeah, anywhere you can use jQuery, do so - it'll make life easier.

share/goodie/game2048/game2048.js
Roysten: The translate attribute is generated based on the place where the tile has to move to. We could add a separate class for every tile location (or generate it using tools such as Sass). However, what would we gain by doing so?

share/goodie/game2048/game2048.js
Roysten: Why? This works, gets the job done and `display: none` will work in every browser, and it is faster because no jQuery caching is done.

share/goodie/game2048/game2048.js
Roysten: I do not agree on this occasion, storing the game over message won't gain us much performance wise. I rather avoid having to many global variables laying around, as it clutters the code.

share/goodie/game2048/game2048.js
Roysten: I agree, as this element gets updated frequently.

share/goodie/game2048/game2048.js
Roysten: Yep, nice catch!

share/goodie/game2048/game2048.js
mintsoft: Consistency?

share/goodie/game2048/game2048.js
moollaza: It seems silly that we're doing these lookups every time the game finishes, even if there aren't considerable performance gains, in my opinion in simplifies the code and makes it a little more flexible because we have the selector in 1 place and updating it only requires changing 1 line instead of 2 or more.

We're also already storing the `#game2048__area_container` selector in the function below, but not using it here which again seems silly.

We can store the games area and the container once when the DOM is done and then use `find()` elsewhere to lookup other selectors within the game area.

GuiltyDolphin: Yar - good coding practice would encourage the use of variables when we have repeated values/look-ups. Also, giving a name to some of the look-ups could help improve the readability of the code.

Roysten:
@moollaza @GuiltyDolphin I wanted to fix this PR and get it finished as soon as possible, but for some reason the css is not used anymore and the build function in the javascript code is not called? I am not sure what goes wrong, I even reverted to some old commits (before the merge) but those also don't work. Has something changed internally?

lib/DDG/Goodie/Game2048.pm
```diff
@@ -1,35 +1,21 @@
...
+    return '',
     structured_answer => {
-        data => {
-            title => $text,
-            subtitle => $dimension . 'x' . $dimension,
-            inputNum => $inputNum,
-            dimension => $dimension
-        },
+        id => 'game2048',
+        name => '2048',
```
GuiltyDolphin: We should remove the name and ID.

Roysten: Yes, I tried that... unfortunately it does not fix the problem. When I call the function DDH.game2048.build.onShow() manually it generates tiles, so the Javascript is valid. However, when the IA is triggered this function is not called automatically. Also the css is not applied. Is the name still `game2048` or should it be something different?

GuiltyDolphin: @Roysten I mean - we don't support ID and name anymore (it is dealt with by the metadata on duck.co); so we don't need it here.

Roysten:
I know, I am sorry that was not clear. However I don't know how one can determine what the css prefix is and what the javascript object should be named for ddg to be able to find it?

lib/DDG/Goodie/Game2048.pm
```diff
@@ -1,35 +1,21 @@
...
 
 use DDG::Goodie;
 
 zci answer_type => "2048";
 zci is_cached   => 1;
 
-my @game_modes = qw (128 256 512 1024 2048 4096 8192);
-my @triggers = map { "play $_" } @game_modes;
-triggers start => @triggers;
+triggers any => "play 2048", "game 2048";
```
GuiltyDolphin: We should probably switch this to `start`, or at least use a regex to guard against queries like `how do I play 2048`.

share/goodie/game2048/game2048.js
moollaza: In general we try to keep HTML strings out of the JS. We also avoid the use of inline styles, even in our template files, when we can.

Granted, we're not dealing with complex or lengthy HTML here, but IMO my suggestion is a little cleaner.

moollaza:
> Has something changed internally?

@Roysten yes -- we're now using the `id` and `name` defined on the IA Page for Goodies and Spice.

The ID on the IA Page is `game_2048` -- which means the expected files folders are `share/goodie/game_2048/game_2048.js`.

I'll change the ID so that it is `game2048` which matches the current file/folder structure.

moollaza:
@Roysten you'll also need to fix the first two lines of `game2048.js`:

```js
DDH.game2048 = DDH.game2048 || {};

DDH.game2048.build = function(ops) {
```

And remove the `id` and `name` from the Perl:

```perl
structured_answer => {
        data => { },
        templates => {
            group => 'text',
            item => 0,
            options => {
                content => 'DDH.game2048.content'
            }
        }
    };
```

moollaza:
With that, it will be working again 👍

Roysten:
@moollaza Thanks, it works again! :) I think I fixed most of the remarks, the only one that remains:

> Why do we need to apply this same CSS class throughout the code in different places? Can we instead add/remove a class?

> Or simply set it once?

What do we want to do here? I could create a separate class for every possible position, but this will still require some sort of string concatenation to build the class name.

moollaza:
@Roysten so sorry about the delay here! This code is definitely good enough for merge. We can always make adjustments to the CSS/JS later if necessary.

It looks like there's a small conflict with your changes to the test file. Once that's fixed I can deploy this to Beta to make sure everything is OK and Merge it 👍

Roysten:
@moollaza Done!

moollaza:
@Roysten Thanks! -- Let me get this on Beta to make sure it's all good...

moollaza:
@Roysten its now live on Beta! https://beta.duckduckgo.com/?q=play+2048&ia=answer

moollaza:
@Roysten hmm the only thing that's odd is the mobile experience. We show an answer tab, and only once clicked does it hide the tab, which seems odd.

Let me check with the team if we have a better way of handling that...

moollaza:
@Roysten I'm going to merge this for now and talk to team about how we want to handle mobile for now.

Would you be interested in adding support for touch events so it can be played on mobile screens?

moollaza:
@Roysten by the way, I meant to say you've done a splendid job getting it this far! Thanks a **_lot_** 👍 💯 🍰 🎉

edgesince84:
hey @Roysten thank you for your contribution to DuckDuckHack, we really appreciate it! I'm reaching out to let you know that we recently launched a new forum with the initiative of improving programming-related searches, and I hope you'll [join us](https://forum.duckduckhack.com/)!

There are a lot of important tasks up for grabs that we'd like your help with, so I hope you'll [jump in]](https://forum.duckduckhack.com/t/duckduckhack-programming-mission-overview/53)!

Roysten:
So... have we figured out how to do touch events on mobile devices yet? :smile:

PR #2833, opened 2016-04-05
NumberStyler GoodieRole: Improvements

GuiltyDolphin:
Working on the NumberStyler Role!
## TO-DO
- [ ] Incorporate #2766 
- [x] Update existing IAs to new format(s)
- [ ] ~~Move number formats into YAML~~
- [x] Make formats locale aware (per #2797)

PR #2898, opened 2016-04-17
Improvements to AltCalendars IA

gaulrobe:
Triggers and calendar metadata now come from a JSON file so that additional calendar conversions can be added easily and more reliably. I also added some new calendars I found while searching through wikipedia.

https://duck.co/ia/view/alt_calendars
Maintainer: **_NONE_**

### DIFF
--- a/lib/DDG/Goodie/AltCalendars.pm
+++ b/lib/DDG/Goodie/AltCalendars.pm
@@ -10,6 +10,8 @@ use JSON;
 with 'DDG::GoodieRole::NumberStyler';
 my $number_regex = number_style_regex();
 
+my $base_wiki_link = "https://en.wikipedia.org/wiki/";
+
 my $definitions_json = share('definitions.json')->slurp();
 my $year_definitions = decode_json($definitions_json);
 
@@ -35,7 +37,7 @@ handle query_parts => sub {
         my $era_hash = $year_definitions->{$parent_era};
         
         my $gregorian_year_started = $era_hash->{'gregorian_year_started'};
-        my $wikipedia_link = $era_hash->{'wikipedia_link'};
+        my $wiki_page = $era_hash->{'wikipedia_page'};
         my $year = $gregorian_year_started + $era_year;
         my $era = DateTime->now->set_year($year)->era;
         
@@ -52,7 +54,7 @@ handle query_parts => sub {
                 },
                 meta => {
                     sourceName => "Wikipedia",
-                    sourceUrl => "$wikipedia_link"
+                    sourceUrl => "$base_wiki_link$wiki_page"
                 },
                 templates => {
                     group => 'info',
--- a/lib/DDG/Goodie/AltCalendars.pm
+++ b/lib/DDG/Goodie/AltCalendars.pm
@@ -29,7 +29,7 @@ handle query_parts => sub {
     # Ignore single word queries
     return unless scalar(@_) > 1;
 
-    if ($_ =~ /^(.*\b)([A-Za-z]+)\s+($number_regex)(.*\b)$/i) {
+    if ($_ =~ /^(.*\b)([A-Za-z]+)\s+(\d*[1-9]\d*)(.*\b)$/) {
         my $era_name = lc($2);
         my $era_year = $3;
         
--- a/lib/DDG/Goodie/AltCalendars.pm
+++ b/lib/DDG/Goodie/AltCalendars.pm
@@ -7,10 +7,18 @@ use DateTime;
 use DDG::Goodie;
 use JSON;
 
+with 'DDG::GoodieRole::NumberStyler';
+my $number_regex = number_style_regex();
+
 my $definitions_json = share('definitions.json')->slurp();
 my $year_definitions = decode_json($definitions_json);
 
-triggers any => keys $year_definitions;
+my %year_map_with_aliases = map { 
+  my $name = $_; 
+  map { $_ => $name } ($name, @{$year_definitions->{$name}->{'aliases'} // []}) 
+} (keys $year_definitions);
+
+triggers any => keys %year_map_with_aliases;
 
 zci answer_type => 'date_conversion';
 zci is_cached => 1;
@@ -19,11 +27,15 @@ handle query_parts => sub {
     # Ignore single word queries
     return unless scalar(@_) > 1;
 
-    if ($_ =~ /^(.*\b)([A-Za-z]+)\s+(\d*[1-9]\d*)(.*\b)$/i) {
+    if ($_ =~ /^(.*\b)([A-Za-z]+)\s+($number_regex)(.*\b)$/i) {
         my $era_name = lc($2);
         my $era_year = $3;
-        my $gregorian_year_started = $year_definitions->{$era_name}{'gregorian_year_started'};
-        my $wikipedia_link = $year_definitions->{$era_name}{'wikipedia_link'};
+        
+        my $parent_era = $year_map_with_aliases{$era_name};
+        my $era_hash = $year_definitions->{$parent_era};
+        
+        my $gregorian_year_started = $era_hash->{'gregorian_year_started'};
+        my $wikipedia_link = $era_hash->{'wikipedia_link'};
         my $year = $gregorian_year_started + $era_year;
         my $era = DateTime->now->set_year($year)->era;
         
--- a/lib/DDG/Goodie/AltCalendars.pm
+++ b/lib/DDG/Goodie/AltCalendars.pm
@@ -7,9 +7,6 @@ use DateTime;
 use DDG::Goodie;
 use JSON;
 
-with 'DDG::GoodieRole::NumberStyler';
-my $number_regex = number_style_regex();
-
 my $base_wiki_link = "https://en.wikipedia.org/wiki/";
 
 my $definitions_json = share('definitions.json')->slurp();
--- a/lib/DDG/Goodie/AltCalendars.pm
+++ b/lib/DDG/Goodie/AltCalendars.pm
@@ -2,6 +2,8 @@ package DDG::Goodie::AltCalendars;
 # ABSTRACT: Convert non-Gregorian years to the Gregorian calendar
 
 use strict;
+
+use DateTime;
 use DDG::Goodie;
 use JSON;
 
@@ -23,15 +25,17 @@ handle query_parts => sub {
         my $gregorian_year_started = $year_definitions->{$era_name}->{'gregorian_year_started'};
         my $wikipedia_link = $year_definitions->{$era_name}->{'wikipedia_link'};
         my $year = $gregorian_year_started + $era_year;
+        my $era = DateTime->now->set_year($year)->era;
         
         $era_name = ucfirst($era_name);
+        $year = abs($year);
 
-        my $answer = "$era_name $era_year is equivalent to $year in the Gregorian Calendar";
+        my $answer = "$era_name $era_year is equivalent to $year $era in the Gregorian Calendar";
 
         return $answer,
             structured_answer => {
                 data => {
-                    title => $year,
+                    title => "$year $era",
                     subtitle => "$era_name Year $era_year"
                 },
                 meta => {
--- a/lib/DDG/Goodie/AltCalendars.pm
+++ b/lib/DDG/Goodie/AltCalendars.pm
@@ -17,12 +17,14 @@ handle query_parts => sub {
     # Ignore single word queries
     return unless scalar(@_) > 1;
 
-    if ($_ =~ /^([A-Za-z]+)\s+(\d*[1-9]\d*)$/i) {
-        my $era_name = $1;
-        my $era_year = $2;
+    if ($_ =~ /^(.*\b)([A-Za-z]+)\s+(\d*[1-9]\d*)(.*\b)$/i) {
+        my $era_name = lc($2);
+        my $era_year = $3;
         my $gregorian_year_started = $year_definitions->{$era_name}->{'gregorian_year_started'};
         my $wikipedia_link = $year_definitions->{$era_name}->{'wikipedia_link'};
         my $year = $gregorian_year_started + $era_year;
+        
+        $era_name = ucfirst($era_name);
 
         my $answer = "$era_name $era_year is equivalent to $year in the Gregorian Calendar";
 
--- a/lib/DDG/Goodie/AltCalendars.pm
+++ b/lib/DDG/Goodie/AltCalendars.pm
@@ -3,33 +3,26 @@ package DDG::Goodie::AltCalendars;
 
 use strict;
 use DDG::Goodie;
+use JSON;
 
-triggers any => 'juche', 'minguo', 'meiji', 'taisho', 'taishou', 'showa', 'shouwa', 'heisei';
+my $definitions_json = share('definitions.json')->slurp();
+my $year_definitions = decode_json($definitions_json);
+
+triggers any => keys $year_definitions;
 
 zci answer_type => 'date_conversion';
 zci is_cached => 1;
-
-my %eras = (
-    'Meiji' => [1867, 'Meiji_period'], # Japanese Meiji era
-    'Taisho' => [1911, 'Taisho_period'], # Japanese Taisho era
-    'Taishou' => [1911, 'Taisho_period'], # Alternative spelling of "Taisho"
-    'Showa' => [1925, 'Showa_period'], # Japanese Showa era
-    'Shouwa' => [1925, 'Showa_period'], # Alternative spelling of "Showa"
-    'Heisei' => [1988, 'Heisei_period'], # Japanese Heisei era
-    'Juche' => [1911, 'North_Korean_calendar'], # North Korean Juche era
-    'Minguo' => [1911, 'Minguo_calendar'], # ROC (Taiwanese) Minguo era
-);
 
 handle query_parts => sub {
     # Ignore single word queries
     return unless scalar(@_) > 1;
 
-    if ($_ =~ /^(.*\b)(meiji|taisho|taishou|showa|shouwa|heisei|juche|minguo)\s+(\d*[1-9]\d*)(\b.*)$/i) {
-        my $era_name = ucfirst($2);
-        my $era_year = $3;
-        my $year = $eras{$era_name}[0] + $era_year;
-        my $result = $1.$year.$4;
-        my $wiki = 'https://en.wikipedia.org/wiki/';
+    if ($_ =~ /^([A-Za-z]+)\s+(\d*[1-9]\d*)$/i) {
+        my $era_name = $1;
+        my $era_year = $2;
+        my $gregorian_year_started = $year_definitions->{$era_name}->{'gregorian_year_started'};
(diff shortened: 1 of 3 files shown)


daxtheduck:
@gaulrobe Hey!
        
If you have any questions along the way, feel free to ask them here. Our staff and community are also available on Slack to answer any questions you may have. If you'd like to join us there please head to https://quackslack.herokuapp.com/ to get an invite.

>More Info: http://docs.duckduckhack.com/submitting/submitting-overview.html

Thanks!

daxtheduck:
Metadata placeholder post
            
---
            
**How to find your IA page**
- New Instant Answer
              
- Existing Instant Answer

lib/DDG/Goodie/AltCalendars.pm
```diff
@@ -3,45 +3,38 @@ package DDG::Goodie::AltCalendars;
...
+        my $year = $gregorian_year_started + $era_year;
 
         my $answer = "$era_name $era_year is equivalent to $year in the Gregorian Calendar";
 
         return $answer,
             structured_answer => {
                 data => {
                     title => $year,
-                    subtitle => "$era_name $era_year - Equivalent Gregorian Year"
+                    subtitle => "$era_name Year $era_year"
```
moollaza: @gaulrobe All the tests need to be updated to match this format, right now they're failing

gaulrobe:
@moollaza sorry about that... I think I've patched things up now! 

It also occurred to me that having the dates handle AD/BC years would be really nice here. For another PR though. :smile:

GuiltyDolphin:
@gaulrobe You can use `DateTime` (`DateTime::Locale`) for eras. 

```perl
perl -MDateTime -E 'say DateTime->now->set_year(-5000)->era;'
BC
```
```perl
perl -MDateTime -E 'say DateTime->now->era;'
AD
```

If you look up about the Common Locale Data Repository there is a wealth of information on date attributes/formatting.

gaulrobe:
@GuiltyDolphin oh wow that is really helpful! I will definitely look into this. Thanks!

gaulrobe:
@GuiltyDolphin I've updated this PR with DateTime era handling. Having AD/BC info here is really nice!

lib/DDG/Goodie/AltCalendars.pm
```diff
@@ -2,46 +2,45 @@ package DDG::Goodie::AltCalendars;
...
-    if ($_ =~ /^(.*\b)(meiji|taisho|taishou|showa|shouwa|heisei|juche|minguo)\s+(\d*[1-9]\d*)(\b.*)$/i) {
-        my $era_name = ucfirst($2);
+    if ($_ =~ /^(.*\b)([A-Za-z]+)\s+(\d*[1-9]\d*)(.*\b)$/i) {
+        my $era_name = lc($2);
         my $era_year = $3;
-        my $year = $eras{$era_name}[0] + $era_year;
-        my $result = $1.$year.$4;
-        my $wiki = 'https://en.wikipedia.org/wiki/';
+        my $gregorian_year_started = $year_definitions->{$era_name}->{'gregorian_year_started'};
+        my $wikipedia_link = $year_definitions->{$era_name}->{'wikipedia_link'};
```
GuiltyDolphin: You can do `$year_definitions->{$era_name}{'wikipedia_link'};` seeing as it is a multi lookup. (Same elsewhere).

gaulrobe: @GuiltyDolphin thanks for the tip! Done.

gaulrobe:
@GuiltyDolphin @moollaza how is this PR looking? Thanks!

lib/DDG/Goodie/AltCalendars.pm
```diff
@@ -2,46 +2,45 @@ package DDG::Goodie::AltCalendars;
...
-    'Minguo' => [1911, 'Minguo_calendar'], # ROC (Taiwanese) Minguo era
-);
-
 handle query_parts => sub {
     # Ignore single word queries
     return unless scalar(@_) > 1;
 
-    if ($_ =~ /^(.*\b)(meiji|taisho|taishou|showa|shouwa|heisei|juche|minguo)\s+(\d*[1-9]\d*)(\b.*)$/i) {
-        my $era_name = ucfirst($2);
+    if ($_ =~ /^(.*\b)([A-Za-z]+)\s+(\d*[1-9]\d*)(.*\b)$/i) {
```
GuiltyDolphin: Why do we have both `[A-Za-z]` and the `i` modifier? We only need one or the other.

share/goodie/alt_calendars/definitions.json
```diff
@@ -0,0 +1,70 @@
...
+        "gregorian_year_started": -1166,
+        "wikipedia_link": "https://en.wikipedia.org/wiki/Discordian_calendar"
+    },
+    "erisian": {
+        "gregorian_year_started": -1166,
+        "wikipedia_link": "https://en.wikipedia.org/wiki/Discordian_calendar"
+    },
+    "nanakshahi": {
+        "gregorian_year_started": 1468,
+        "wikipedia_link": "https://en.wikipedia.org/wiki/Nanakshahi_calendar"
```
GuiltyDolphin: I think we should do two things here:

1. Have just the main name with aliases,

2. Only use the base-part of the link,

For example:

```json
"discordian": {
  "gregorian_year_started": -1166,
  "aliases": ["erisian"],
  "wikipedia_page": "Discordian_calendar"
},
```

Then just infer the first part of the link, and use aliases to point to the main calendar.

lib/DDG/Goodie/AltCalendars.pm
GuiltyDolphin: Also, looks like you are trying to match numbers here, so we should be using the [NumberStyle GoodieRole](http://docs.duckduckhack.com/backend-reference/goodie-helpers.html#numbers).

gaulrobe: Nice. I hadn't seen those GoodieRoles yet. Super handy! :+1:

share/goodie/alt_calendars/definitions.json
gaulrobe: > Have just the main name with aliases

Having each name/alias as a key here allows for easier triggering (https://github.com/gaulrobe/zeroclickinfo-goodies/blob/altcalendars-cleanup/lib/DDG/Goodie/AltCalendars.pm#L13). 

What I could do though is something like:
```
"discordian": {
  "gregorian_year_started": -1166,
  "wikipedia_page": "Discordian_calendar"
},
"erisian": {
  "alias_to": "discordian"
}
```

So that we retain easier triggering and can then use the presence of `alias_to` to get the relevant data.

> Only use the base-part of the link

I don't feel strongly either way, so I will make this change.

Thanks!

GuiltyDolphin: @gaulrobe Hmm? Just create a reverse HASH lookup from names+aliases to the original names. 

```perl
my %alias_map = map { 
  my $name = $_; 
  map { $_ => $name } ($name, @{$eras{$name}->aliases // []}) 
} (keys %eras);

# Then you can grab what you want

my @all_eras = keys %alias_map;

my $actual_era = $alias_map{$era}
```

gaulrobe: Maybe there is a gap in my understanding of the IA infrastructure, but wouldn't this be more expensive of a process? To be fair, the JSON right now is pretty small so this extra processing is probably not a huge factor. But what if more calendars are added?

Having a list of `aliases` in the JSON certainly makes for cleaner JSON, but the code change needed for my proposal would be (in my opinion of course) easier to understand.

```
 my $era_hash = $year_definitions->{$era_name};                                                                                                                                                                                                                 
 if ($era_hash->{'alias_to'}) {                                                                                                                                                                                                                                 
   $era_hash = $year_definitions->{$era_hash->{'alias_to'}}                                                                                                                                                                                                   
}                                                                                                                                                                                                                                                              
my $gregorian_year_started = $era_hash->{'gregorian_year_started'};                                                                                                                                                                                            
my $wikipedia_link = $era_hash->{'wikipedia_link'};  
```

After typing this out, I'm not sure there is a clear winner here for an implementation. If you see a flaw in my thinking please let me know! I appreciate the discussion :smile:

GuiltyDolphin: @gaulrobe TBH you could have hundreds of lookups without any real speed impact, they are pretty cheap.

There's two situs here:

- Inside the handler; anything declared in here gets initialized *every* time the IA triggers.
- Outside the handler; this only gets initialized once per module load (doesn't happen very often).

So you'd be fine sticking it outside the handler.

If you do get to a point where you need thousands of keys, let me know... But that ain't gonna happen ^.^

Remember, the CheatSheets Instant Answer has *hundreds* of full JSON files to process and generate triggers for; but because pretty much all the processing is done at 'compile' time it isn't really an issue; only the work of detecting which cheat sheet triggered is done in the handler.

Another note; the implementation I've suggested is more extensible as we require new aliases, it requires only the addition of an alias (a string), not an entire hash.

Of course the *actual* implementation I gave you is just an example and feel free to roll your own; but I think we should go for having aliases as an attribute rather than a top-level structure.

gaulrobe: @GuiltyDolphin really appreciate the explanation. Understanding the architecture a bit better will help drive future implementations! Leveraging aliases like you recommended sounds like the best way forward here 👍 Thanks!

gaulrobe: Also, looks like I have to find 1000+ alternate calendars 😜

gaulrobe:
@GuiltyDolphin refactored per your suggestion. It turned out nicely, thanks again!

One problem I had was trying to use the NumberStyle GoodieRole... Substituting `(\d*[1-9]\d*)` with that NumberStyle regex broke several of the unit tests. Particularly the ones like https://github.com/gaulrobe/zeroclickinfo-goodies/blob/0a05b287e7e1b7456f367e1f3daf9bed59597317/t/AltCalendars.t#L50 which have additional words or numbers after the `[calendar] [year]`.

Therefore, I've left `(\d*[1-9]\d*)` for handling the `[year]`.

Please take another look!

GuiltyDolphin:
@gaulrobe TBH for the year it might actually be better to use the Dates Role ^.^

Hmm... The role is undergoing some refurbishment at the moment, so hopefully any issues will get cleared up in that.

It might be that you need to wrap the regex with parens to ensure it doesn't consume the rest of the regex with ORs, can you give me an example of how you used it?

lib/DDG/Goodie/AltCalendars.pm
```diff
@@ -29,7 +29,7 @@ handle query_parts => sub {
     # Ignore single word queries
     return unless scalar(@_) > 1;
 
-    if ($_ =~ /^(.*\b)([A-Za-z]+)\s+($number_regex)(.*\b)$/i) {
```
GuiltyDolphin: Ah, you probably wanted `(.*?)$`

gaulrobe:
@GuiltyDolphin 

Some investigation notes..

## Number regex
The `number_style_regex()` matches with a space after a number, so this regex matches `Heisei 12 was a leap year` with "Heisei 12 " and `Heisei 12 4444` with "Heisei 12 4444". 

This matches the expectation set here: https://github.com/duckduckgo/zeroclickinfo-goodies/blob/master/t/00-roles.t#L29

And sure enough, adding "34 " to the valid test cases passes the test.

Is a space a valid character a number can have? Should this be fixed, or is there use cases where this is applicable?

## Date regex
I noticed the full year regex here: https://github.com/duckduckgo/zeroclickinfo-goodies/blob/master/lib/DDG/GoodieRole/Dates.pm#L44 

But that assumes a 4 digit year (which I guess is a "full" year?). Nothing else jumped out at me as something I could use here. I may have missed something though...

You mention the Date role is getting refactored, maybe this use case can be taken into consideration? Maybe something like `qr#[0-9]{1,4}#` to represent a potential year value (until the year 10,000 :wink:) or maybe even something more lenient?

## Moving forward
I'm leaning towards sticking with what I have for now in this PR. When there is new Date/Number roles to be tested, I would be glad to help!

GuiltyDolphin:
@gaulrobe Yeah, the NumberStyle role completely ignores spaces (but allows them in numbers).

#2833 is the NumberStyle GoodieRole Improvements.

#2810 is the Dates GoodieRole Improvements.

gaulrobe:
@GuiltyDolphin thanks for letting me know about these improvements! Will the weekly newsletter mention these improvements once they are completed?

Does this PR look good to go?

lib/DDG/Goodie/AltCalendars.pm
```diff
@@ -2,46 +2,56 @@ package DDG::Goodie::AltCalendars;
...
-        my $answer = "$era_name $era_year is equivalent to $year in the Gregorian Calendar";
+        my $answer = "$era_name $era_year is equivalent to $year $era in the Gregorian Calendar";
 
         return $answer,
             structured_answer => {
                 data => {
-                    title => $year,
-                    subtitle => "$era_name $era_year - Equivalent Gregorian Year"
+                    title => "$year $era",
+                    subtitle => "$era_name Year $era_year"
```
moollaza: I think it's still important to indicate what exactly the answer we're showing is, being the "Equivalent Gregorian Year"

lib/DDG/Goodie/AltCalendars.pm
```diff
@@ -2,46 +2,56 @@ package DDG::Goodie::AltCalendars;
...
-    'Taisho' => [1911, 'Taisho_period'], # Japanese Taisho era
-    'Taishou' => [1911, 'Taisho_period'], # Alternative spelling of "Taisho"
-    'Showa' => [1925, 'Showa_period'], # Japanese Showa era
-    'Shouwa' => [1925, 'Showa_period'], # Alternative spelling of "Showa"
-    'Heisei' => [1988, 'Heisei_period'], # Japanese Heisei era
-    'Juche' => [1911, 'North_Korean_calendar'], # North Korean Juche era
-    'Minguo' => [1911, 'Minguo_calendar'], # ROC (Taiwanese) Minguo era
-);
-
 handle query_parts => sub {
```
moollaza: We likely need to do some more query validation in here. Right now anything containing a valid year, regardless of the rest of the query seems to trigger this IA:

heisei 20 + 45

gaulrobe: In some of these cases having the IA be triggered might actually help. Seeing `2001 AD` when you search for `rare books jp heisei 13` (or something similar) seems like we are providing a more immersive experience and provides some good context about what/when you are searching for.

share/goodie/alt_calendars/definitions.json
```diff
@@ -0,0 +1,58 @@
...
+    },
+    "minguo": {
+        "gregorian_year_started": 1911,
+        "wikipedia_page": "Minguo_calendar"
+    },
+    "pisan": {
+        "gregorian_year_started": -1,
+        "wikipedia_page": "Pisan_calendar"
+    },
+    "bengali": {
```
moollaza: Seems to be having issues -- returns the same year number in AD, is that correct?

image 2016-05-24 at 12 17 32 pm

share/goodie/alt_calendars/definitions.json
```diff
@@ -0,0 +1,58 @@
+{
+    "meiji": {
```
moollaza: Again, some autocomplete queries that shouldn't trigger this do:

meiji_26_revolver_at_duckduckgo

lib/DDG/Goodie/AltCalendars.pm
gaulrobe: I have no objections to making this clear to users, but I do not like the `Equivalent Gregorian Year` verbiage here much as seeing `Heisei 13 - Equivalent Gregorian Year` makes it seem like Heisei 13 is a Gregorian date. I was hoping adding AD/BC to the calculated year would be enough to show that this is a Gregorian date.

moollaza:
@gaulrobe the code looks pretty good to me!

However, it looks like there are relevancy issues here. We may need to rely on a regex to ensure the query matches our expected pattern(s). Alternatively we could strip away the useful words/numbers and check if there's anything leftover then bail

lib/DDG/Goodie/AltCalendars.pm
GuiltyDolphin: @gaulrobe Usually we err on the side of *don't trigger* - unless we *know* the user is searching for this result (or it is safe to assume so), we probably shouldn't trigger.

These are Instant *Answers* after all, we should be *answering* the query.

gaulrobe: @GuiltyDolphin @moollaza 

Knowing when a user is expecting a result is a tough question to answer. There are some words we could use here (like `when` being in the query) and say yeah, that is when this should trigger. You could also argue that queries like `heisei 13` aren't necessarily something users expect a Gregorian calendar translation to.

Is there a feedback loop for users to say "Good/Bad" given an instant answer that appears on the search results? That could be a powerful thing for us so that we can understand how these IAs are being used (and make better triggers, better IAs, etc).

Either way, I'm fine with coming up with some better query validation, but I would love to have more insight on how this IA is being used and what kind of triggers would make this more impactful/helpful to users.

GuiltyDolphin: @gaulrobe I've discussed exactly this (better IA feedback opportunities) with some of the other Community Leaders - currently feedback doesn't *really* provide any *direct* application to IAs - which is a bit of a pain!

gaulrobe: @GuiltyDolphin very happy to hear about this! 👍 I hope this gains some traction!

edgesince84:
@GuiltyDolphin @gaulrobe when you guys mention "direct feedback," do you mean you're looking to hear when someone leaves feedback in regards to the IA?

We're currently doing that -- any feedback reports in regards to an open-source IA gets relayed as an issue in GitHub, when appropriate.  I'm heading it up, so let me know if you have any questions or suggestions about the process.

gaulrobe:
@edgesince84 It would be nice to have even more direct feedback for just the IA. I see the `Send Feedback` button, but that is more in line with the search results. It'd be really cool, to have a similar thing inside the IA...

I'd love to know raw numbers (good vs bad) on what people think given a search query and the IA that may have appeared to them. Being able to see what queries are "good" vs what queries are "bad" would be helpful! Having an additional text field for additional explanations would be nice too in case users want to provide more feedback.

share/goodie/alt_calendars/definitions.json
gaulrobe: @moollaza good find, fixed this in 4774611756531923a7331350d5b292c1be84f60d

gaulrobe:
@moollaza @GuiltyDolphin I have toned down the regex so that we only match `[CALENDAR] [YEAR]` so that this can get out the door. I still think there is value for broadening when this IA is triggered, so hopefully we can iterate on this at a later time.

gaulrobe:
@moollaza can this RP get merged? I'd love to see these improvements get out the door :smiley:

moollaza:
@gaulrobe Thanks! This LGTM :+1:

PR #3019, opened 2016-05-09
Conversions: Harmonise unit definitions

mintsoft:
- [x] Improvement
    - [x] Enhancement: Moves definition of plurals into YAML

###### Description of changes

We've had a few issues crop up where the plural name of the unit is incorrect recently; also a few PR's adding units haven't seen the magical `%plural_exceptions` which means when they're adding definitions that they get missed. I'm moving the definition into the YAML with the unit name, that way the user is encouraged to define the word for the plural when they define the unit itself. This is part of a much larger rework but I want this in ASAP really.

As part of this I've changed a few query results for which the questions were ambiguous; neither way is "right" however I think the new way is (for better or worse) what *I* would expect.

---


daxtheduck:
Metadata placeholder post
            
---
            
**How to find your IA page**
- New Instant Answer
              
- Existing Instant Answer

t/Conversions.t
```diff
@@ -588,14 +588,14 @@ ddg_goodie_test(
...
-        '2,500 large calories = 0.003 tons of TNT',
+        '2,500 large calories = 0.003 tons of tnt',
         structured_answer => make_answer({
             markup_input => '2,500',
             raw_input => '2500',
             from_unit => 'large calories',
             styled_output => '0.003',
             raw_answer => '0.003',
-            to_unit => 'tons of TNT',
+            to_unit => 'tons of tnt',
```
GuiltyDolphin: Why have we gone lowercase? I think `TNT` is more correct.

t/Conversions.t
```diff
@@ -984,14 +984,14 @@ ddg_goodie_test(
         })
     ),
     'ml in gallons' => test_zci(
-        '1 millilitre = 0.000264 us gallons',
+        '1 us gallon = 3,785.412 millilitres',,
```
GuiltyDolphin: So, we've swapped the order? First arg goes second? Or is there another ordering?

t/Conversions.t
mintsoft: It currently (*completely* incorrectly) doesn't handle case properly for example it doesn't distinguish between mV and MV; it'll be fixed later on

t/Conversions.t
mintsoft: Yeah basically

mintsoft:
@GuiltyDolphin I've removed `Work In Progress`; because as far as I'm concerned this is finished as a PR. I want to tackle all the other stuff in separate PRs etc

t/Conversions.t
GuiltyDolphin: @mintsoft Are you planning to fix this in this PR? I'd rather we didn't 'invalidate' the production version :/

mintsoft: It's already behaving like this in production.

mintsoft:
@GuiltyDolphin If you find it misbehaving, I'm only interested in regressions against live. The billions of edge-cases that are handled terribly I already know about!

t/Conversions.t
GuiltyDolphin: @mintsoft Very well - I've created a PR against this branch with a fix - and... Some automatic white space correction.

mintsoft: @GuiltyDolphin I'll pull that into this branch; easier merging then

mintsoft: EDIT: Actually I won't because then we'll have to wait for a third person to merge.

GuiltyDolphin: @mintsoft Eh? A third...

mintsoft: If I pull your changes into here, then this PR contains both mine and your code; I'd wait for somebody who's code isn't in the PR to merge it basically.

GuiltyDolphin: @mintsoft It's (roughly) the same as doing it against master and you approving it - I don't think there's any issues. :suspect:

mintsoft: True; but I wouldn't do it personally :)

GuiltyDolphin: Though this does re-raise the point about having 'community' branches - they do it quite a bit over on the Platform (have a 'dev' branch that they PR against); so it'd be nice if we could open up some dev branches to users.

GuiltyDolphin: And okay - we'll sort this first then we can PR the other branch against master.

mintsoft: Yeah; when mwmiller and I used to work on things together in the same branch we found the feedback/change cycle much easier to manage between us

lib/DDG/Goodie/Conversions.pm
```diff
@@ -194,31 +170,31 @@ handle query_lc => sub {
...
+          }
       };
 };
 
 sub looks_plural {
-    my ($unit) = @_;
-    my @unit_letters = split //, $unit;
-    return exists $singular_exceptions{$unit} || $unit_letters[-1] eq 's';
+    my ($input) = @_;
+    return any {lc($_) eq lc($input)} values(%plurals);
```
GuiltyDolphin: Rather than having to build this every time, why not have a HASH from lowercase plurals to units, then do `return defined $plural_to_unit{$input}`?

t/Conversions.t
mintsoft: @GuiltyDolphin Now I've looked at it again, I've completely misdiagnosed this; it's not the same problem at all, so I've just fixed it

mintsoft:
@GuiltyDolphin All done and dusted

GuiltyDolphin:
@mintsoft Well, to my young eyes this LGTM, thank you :dolphin:

Issue #2767, opened 2016-03-18
Convert Simple Structured Answer Goodies to Full Templates


moollaza:
### Background

Goodies have full support for templates, and we'd like to make sure all Goodies are using our full templates. We provided the "simple structured_answer" as a way to get them using basic templates before they had full template support and now the time has come to upgrade them all.

### What To Do

1. Change the returned `structured_answer` object (in one of the .pm files below) to match the new format, which specifies a template and the data
    
    Nearly all of these Goodies should be able to use our simple `text` template and should return a `structured_answer` that looks like the one, returned by the PigLatin Goodie in [PigLatin.pm](https://github.com/duckduckgo/zeroclickinfo-goodies/blob/master/lib/DDG/Goodie/PigLatin.pm#L40).

2. Update the Instant Answer's test file to match the changes made. The tests verify the expected content of the `structured_answer` for a given query. 

3. Test the IA locally with DuckPAN to make sure it still looks good and the tests pass. 

    You can run the test file with `duckpan test`:
    Example:
    Anagram.pm => `duckpan test Anagram`
    AspectRatio.pm => `duckpan test AspectRatio`

    You can test the visual display using `duckpan server`. You'll want to try some of the queries in the test file and make sure the visuals look good. Ideally identical to how they currently look on the live site.
    
### Example

@GuiltyDolphin has made [a great Pull Request](https://github.com/duckduckgo/zeroclickinfo-goodies/pull/2769/files) that demonstrates the necessary changes for the Anagram Instant Answer:

<img width="999" alt="anagrams__update_to_full_template_by_guiltydolphin_ _pull_request__2769_ _duckduckgo_zeroclickinfo-goodies" src="https://cloud.githubusercontent.com/assets/8598426/14021490/a110363e-f1d3-11e5-938f-faf11bef7016.png">

When updating the test file, it will make things _**much**_ easier if you create a small function that generates a structured answer object. This way you won't need to manually type out the full structured answer we expect for each individual query.

@GuiltyDolphin has done this in the Anagram PR as well:
<img width="918" alt="anagrams__update_to_full_template_by_guiltydolphin_ _pull_request__2769_ _duckduckgo_zeroclickinfo-goodies" src="https://cloud.githubusercontent.com/assets/8598426/14021562/e795ccc2-f1d3-11e5-8817-cd16fe0531d7.png">

[...]

PR #3041, opened 2016-05-12
HTMLEntitiesDecode: Converting to Text Template

mintsoft:
In reference to: #2767 ; I've not tested it in `duckpan server` yet as I can't run it from my current location.

--------------------------------
Instant Answer Page: https://duck.co/ia/view/htmlentities_decode

[Maintainer](http://docs.duckduckhack.com/maintaining/guidelines.html): @nishanths

### DIFF
--- a/t/HTMLEntitiesDecode.t
+++ b/t/HTMLEntitiesDecode.t
@@ -9,11 +9,19 @@ use DDG::Test::Goodie;
 zci answer_type => 'html_entity';
 zci is_cached   => 1;
 
-sub any_structured_answer {
+sub build_structured_answer {
+    my ($title, $subtitle) = @_;
 	return {
-		input => ignore(),
-		result => ignore(),
-		operation => "HTML Entity Decode"
+	    data => {
+	        title => $title,
+	        subtitle => $subtitle,
+	    },
+	    templates => {
+	        group => 'text',
+	        options => {
+	            content => 'DDH.html_entity.content'
+	        }
+	    }
 	};
 }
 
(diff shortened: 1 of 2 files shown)


daxtheduck:
Metadata placeholder post
            
---
            
**How to find your IA page**
- New Instant Answer
              
- Existing Instant Answer

t/HTMLEntitiesDecode.t
```diff
@@ -55,11 +63,12 @@ ddg_goodie_test(
     'html decode apostrophe' => undef,
 
     # natural querying
-    'What is the decoded html entity for &#960;?' => test_zci(qr/.*/, structured_answer => any_structured_answer()),
+    'What is the decoded html entity for &#960;?' => test_zci(qr/.*/, structured_answer => build_structured_answer("&pi;","HTML Entity Decode: &amp;#960;")),
```
GuiltyDolphin: We should probably hard-code the title.

t/HTMLEntitiesDecode.t
```diff
@@ -8,44 +8,52 @@ use DDG::Test::Goodie;
...
+	    }
 	};
 }
 
 ddg_goodie_test(
     [qw(DDG::Goodie::HTMLEntitiesDecode)],
 
     # Simple decimal test
-    'html decode &#33;' => test_zci("Decoded HTML Entity: !", structured_answer => any_structured_answer()),
+    'html decode &#33;' => test_zci("Decoded HTML Entity: !", structured_answer => build_structured_answer("!", "HTML Entity Decode: &amp;#33;")),
```
GuiltyDolphin: It'd be nice if we used a `build_test` sub for calling `test_zci(build_structured_answer(@_))` (or similar).

mintsoft: Hmm; is that the standard we're going with now? Personally I prefer it without the `build_test`

GuiltyDolphin: @mintsoft I believe so - we have it in the templates and it is supposed to be a wrapper to let you do whatever with the test if you need to.

mintsoft: Fair enough I'll switch it out when I get home

GuiltyDolphin:
@mintsoft This doesn't display anything for me on `duckpan server` (triggers, but gives empty response); not sure if it's just me though - can you confirm?

mintsoft:
@GuiltyDolphin Completely correct; the ID is wrong I think. I'm just looking at that now

mintsoft:
@GuiltyDolphin Should be sorted now, I have no idea why I thought I needed to return options content!? 

I've removed all the `html_enc` calls so that it displays "what I want"; I think this should be safe from an XSS perspective; however I'd like signoff from staff. cc @moollaza

moollaza:
Deploying to beta...

moollaza:
@mintsoft Thanks! This LGTM :+1:

PR #3088, opened 2016-05-17
DDG::Goodie::DuckDuckGo: Initial DNS delegation explanation.

nilnilnil:
DDG::Goodie::DuckDuckGo: Initial DNS delegation explanation.

### DIFF
--- a/lib/DDG/Goodie/DuckDuckGo.pm
+++ b/lib/DDG/Goodie/DuckDuckGo.pm
@@ -61,9 +61,10 @@ handle query_raw => sub {
 
     #warn "Query: '$_'\tTrigger: '$trigger'\tMajor Key: '$key'";
 
+    return if $trigger eq 'whois' && $key ne 'duckduckgoownedserveryahoonet'; # Whois locked to valid keys.
+
     my $response = $responses->{$key};
     return unless $response;
-    return if $trigger eq 'whois' && $key ne 'duckduckgoownedserveryahoonet'; # Whois locked to valid keys.
 
     return $response->{text},
     structured_answer => {
--- a/lib/DDG/Goodie/DuckDuckGo.pm
+++ b/lib/DDG/Goodie/DuckDuckGo.pm
@@ -59,7 +59,7 @@ handle query_raw => sub {
     $key =~ s/$skip_words_re//g; # Strip skip words.
     $key =~ s/\W+//g;            # Strip all white space.
 
-    #warn "Query: '$_'\tTrigger: '$trigger'\tMajor Key: '$key'";
+    warn "Query: '$_'\tTrigger: '$trigger'\tMajor Key: '$key'";
 
     # Whois escape hatch for invalid keys.
     if ($trigger eq 'whois') {
--- a/lib/DDG/Goodie/DuckDuckGo.pm
+++ b/lib/DDG/Goodie/DuckDuckGo.pm
@@ -8,7 +8,7 @@ use YAML::XS 'LoadFile';
 
 my @ddg_aliases = map { ($_, $_ . "'s", $_ . "s") } ('duck duck go', 'duckduck go', 'duck duckgo', 'duckduckgo', 'ddg');
 
-triggers any => @ddg_aliases, "zeroclickinfo", "private search";
+triggers any => @ddg_aliases, "zeroclickinfo", "private search", "whois duckduckgo-owned-server.yahoo.net";
 
 zci is_cached => 1;
 
--- a/lib/DDG/Goodie/DuckDuckGo.pm
+++ b/lib/DDG/Goodie/DuckDuckGo.pm
@@ -61,7 +61,11 @@ handle query_raw => sub {
 
     #warn "Query: '$_'\tTrigger: '$trigger'\tMajor Key: '$key'";
 
-    return if $trigger eq 'whois' && $key ne 'duckduckgoownedserveryahoonet'; # Whois locked to valid keys.
+    # Whois escape hatch for invalid keys.
+    if ($trigger eq 'whois') {
+        return if $key ne 'duckduckgoownedserveryahoonet';
+        $key = 'yahoo';
+    }
 
     my $response = $responses->{$key};
     return unless $response;
--- a/lib/DDG/Goodie/DuckDuckGo.pm
+++ b/lib/DDG/Goodie/DuckDuckGo.pm
@@ -6,15 +6,16 @@ use DDG::Goodie;
 
 use YAML::XS 'LoadFile';
 
-my @ddg_aliases = map { ($_, $_ . "'s", $_ . "s") } ('duck duck go', 'duckduck go', 'duck duckgo', 'duckduckgo', 'ddg');
+my @ddg_aliases = map { ($_."'s", $_."s", $_) } ('duck duck go', 'duckduck go', 'duck duckgo', 'duckduckgo', 'ddg');
 my @any_triggers = (@ddg_aliases, "zeroclickinfo", "private search", "whois");
 
 triggers any => @any_triggers;
 
 zci is_cached => 1;
 
-my $trigger_qr = join('|', @any_triggers);
-$trigger_qr = qr/\b$trigger_qr\b/i;
+my $trigger_qr = join('|', map { "\Q$_\E" } @any_triggers);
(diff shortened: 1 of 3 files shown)


daxtheduck:
Metadata placeholder post
            
---
            
**How to find your IA page**
- New Instant Answer
              
- Existing Instant Answer

lib/DDG/Goodie/DuckDuckGo.pm
```diff
@@ -8,7 +8,7 @@ use YAML::XS 'LoadFile';
 
 my @ddg_aliases = map { ($_, $_ . "'s", $_ . "s") } ('duck duck go', 'duckduck go', 'duck duckgo', 'duckduckgo', 'ddg');
 
-triggers any => @ddg_aliases, "zeroclickinfo", "private search";
+triggers any => @ddg_aliases, "zeroclickinfo", "private search", "whois duckduckgo-owned-server.yahoo.net";
```
GuiltyDolphin: @nilnilnil Why do we need this in the triggers explicitly? And why none of the others?

nilnilnil: @GuiltyDolphin no worries! I appreciate the question, because I could be wrong. From my understanding, the word "duckduckgo", or an alias of it, needs to be in the triggers for this to fire.

I was trying to contain this feature all in the DuckDuckGo.pm module for simplicity; but I am now thinking it may be easier as it's own IA. What do you think?

GuiltyDolphin: @nilnilnil Well, it seems to be in line with the rest of what the IA is trying to do - and yes, the existing triggers require "duckduckgo" - seeing as this is a trigger-based IA it might make sense to use the phrases in the YAML to generate the triggers? I.e., build a list of all the 'aliases' and use those as the triggers.

nilnilnil:
@GuiltyDolphin I don't have time to sanity check this at the moment, but it's pretty important. If you have time, and are motivated to look at it -- please do.

I think there was something broken before this change with the alias system. In the handle function there is a big punctuation and whitespace strip. I don't see that being applied to aliases. That seems wrong.

Again, I don't have any more time to investigate this today directly, but would be interested in your thoughts.

/cc @MariagraziaAlastra

nilnilnil:
Also, I'd like to keep this branch the same and discuss proposals before further code changes.

lib/DDG/Goodie/DuckDuckGo.pm
```diff
@@ -43,16 +49,26 @@ foreach my $keyword (keys %$responses) {
...
-    $key =~ s/\?//g;    # Allow for questions, but don't pollute skip words.
-    $key =~ s/$skip_words_re//g;
-    $key =~ s/\W+//g;
+    my ($trigger) = $key =~ /($trigger_qr)/;
 
-    my $response = $responses->{$key};
+    $key =~ s/\b$trigger\b//g;   # Strip trigger on word boundaries.
+    $key =~ s/\?//g;             # Allow for questions, but don't pollute skip words.
+    $key =~ s/$skip_words_re//g; # Strip skip words.
+    $key =~ s/\W+//g;            # Strip all white space.
```
GuiltyDolphin: Whitespace should be `\s`

lib/DDG/Goodie/DuckDuckGo.pm
```diff
@@ -43,16 +49,26 @@ foreach my $keyword (keys %$responses) {
...
+
+    warn "Query: '$_'\tTrigger: '$trigger'\tMajor Key: '$key'";
```
GuiltyDolphin: We don't need this anymore :wink:

lib/DDG/Goodie/DuckDuckGo.pm
GuiltyDolphin: Btw, why are we stripping all the whitespace? Surely the aliases will work fine with the existing normalized whitespace?

nilnilnil: I didn't modify this bit. You'll need to `git log` or `git bisect` to see why it was added.

lib/DDG/Goodie/DuckDuckGo.pm
nilnilnil: I pulled it and fixed the tests.

lib/DDG/Goodie/DuckDuckGo.pm
GuiltyDolphin: @nilnilnil Oh, whoops! It was showing up on the diff as changed - I guess it is just the comment. I think the comment should probably read `Strip all non-word characters` rather than `whitespace`.

lib/DDG/Goodie/DuckDuckGo.pm
```diff
@@ -6,15 +6,16 @@ use DDG::Goodie;
 
 use YAML::XS 'LoadFile';
 
-my @ddg_aliases = map { ($_, $_ . "'s", $_ . "s") } ('duck duck go', 'duckduck go', 'duck duckgo', 'duckduckgo', 'ddg');
+my @ddg_aliases = map { ($_."'s", $_."s", $_) } ('duck duck go', 'duckduck go', 'duck duckgo', 'duckduckgo', 'ddg');
```
GuiltyDolphin: @nilnilnil Btw, you could do `"${_}'s"` and `"${_}s"`.

lib/DDG/Goodie/DuckDuckGo.pm
```diff
@@ -6,15 +6,16 @@ use DDG::Goodie;
...
 my @any_triggers = (@ddg_aliases, "zeroclickinfo", "private search", "whois");
 
 triggers any => @any_triggers;
 
 zci is_cached => 1;
 
-my $trigger_qr = join('|', @any_triggers);
-$trigger_qr = qr/\b$trigger_qr\b/i;
+my $trigger_qr = join('|', map { "\Q$_\E" } @any_triggers);
```
GuiltyDolphin: If we're using a `map` may as well do `map { quotemeta }`

lib/DDG/Goodie/DuckDuckGo.pm
```diff
@@ -6,19 +6,25 @@ use DDG::Goodie;
...
-my @ddg_aliases = map { ($_, $_ . "'s", $_ . "s") } ('duck duck go', 'duckduck go', 'duck duckgo', 'duckduckgo', 'ddg');
+my @ddg_aliases = map { ($_."'s", $_."s", $_) } ('duck duck go', 'duckduck go', 'duck duckgo', 'duckduckgo', 'ddg');
+my @any_triggers = (@ddg_aliases, "zeroclickinfo", "private search", "whois");
 
-triggers any => @ddg_aliases, "zeroclickinfo", "private search";
+triggers any => @any_triggers;
 
 zci is_cached => 1;
 
+my $trigger_qr = join('|', map { "\Q$_\E" } @any_triggers);
```
GuiltyDolphin: If we're using a `map` may as well do `map { quotemeta }`

GuiltyDolphin:
@nilnilnil Regarding your comment about the normalization in the handler not being applied to aliases - surely this is expected? Aliases are defined locally so we know what form they take - or were you referring to something else?

lib/DDG/Goodie/DuckDuckGo.pm
```diff
@@ -43,16 +50,26 @@ foreach my $keyword (keys %$responses) {
...
 my $skip_words_re = qr/\b(?:what|where|is|of|for|the|in|on)\b/;
 
-handle remainder => sub {
+handle query_raw => sub {
     my $key = lc;
 
-    $key =~ s/\?//g;    # Allow for questions, but don't pollute skip words.
-    $key =~ s/$skip_words_re//g;
-    $key =~ s/\W+//g;
+    my ($trigger) = $key =~ /($trigger_qr)/;
```
GuiltyDolphin: We should be able to use `$req->matched_trigger` for this.

nilnilnil: Huh. I didn't know that existed :smile:.

MariagraziaAlastra:
@nilnilnil LGTM :+1:

nilnilnil:
Thank you @MariagraziaAlastra, and @GuiltyDolphin!

PR #3108, opened 2016-05-20
New MilitaryRank Instant Answer

lowellosaurus:
https://duck.co/ia/view/military_rank


daxtheduck:
@lowellosaurus Hey!
        
If you have any questions along the way, feel free to ask them here. Our staff and community are also available on Slack to answer any questions you may have. If you'd like to join us there please head to https://quackslack.herokuapp.com/ to get an invite.

>More Info: http://docs.duckduckhack.com/submitting/submitting-overview.html

Thanks!

daxtheduck:
Metadata placeholder post
            
---
            
**How to find your IA page**
- New Instant Answer
              
- Existing Instant Answer

lib/DDG/Goodie/MilitaryRank.pm
```diff
@@ -0,0 +1,1287 @@
...
+# ABSTRACT: Return rank structure for a given military branch.
+
+use DDG::Goodie;
+use strict;
+
+zci answer_type => 'military_rank';
+zci is_cached   => 1;
+
+sub DATA {
+    return {
```
GuiltyDolphin: Holy sweet... Lets pull this out into a YAML or JSON file in `share`.

t/MilitaryRank.t
```diff
@@ -0,0 +1,163 @@
...
+zci answer_type => "military_rank";
+zci is_cached   => 1;
+
+sub build_structured_answer {
+    my ($country, $branch, $text) = @_;
+
+    # Use DATA() sub from module because we simply want to test that
+    # the query returns the proper data. We do not test each bit of data
+    # for correctness. Rely on the module to be correct in that sense.
+    my $answer = DDG::Goodie::MilitaryRank::DATA()->{$country}->{$branch};
```
GuiltyDolphin: We should just specify this in the tests - otherwise we are duplicating any data bugs; also we then won't need to import `DDG::Goodie::MilitaryRank`.

lib/DDG/Goodie/MilitaryRank.pm
```diff
@@ -0,0 +1,1287 @@
+package DDG::Goodie::MilitaryRank;
+# ABSTRACT: Return rank structure for a given military branch.
+
+use DDG::Goodie;
+use strict;
+
+zci answer_type => 'military_rank';
+zci is_cached   => 1;
+
+sub DATA {
```
GuiltyDolphin: Btw hard coding data in this manner is *very* inefficient as it will be created repeatedly; best thing to do is pull it out into a separate file that we can load once.

lib/DDG/Goodie/MilitaryRank.pm
lowellosaurus: Will do!

t/MilitaryRank.t
lowellosaurus: By "this" you mean specify the actual data that we were previously pulling from the imported DATA() sub (and, in the updated version, will be loading from the yml file)?

GuiltyDolphin: No - each expected result; the tests are only a subset of the possible queries so it should be manageable. Also you could pull one out into a var for use with multiple tests where the answer doesn't have to change in order to test the feature (e.g., query forms).

lowellosaurus:
Moved data out to YAML file. Not sure if I fully understand what the tests need to look like yet, but I made some changes there too.

lib/DDG/Goodie/MilitaryRank.pm
```diff
@@ -0,0 +1,96 @@
...
+};
+
+my $country_pat = join '|', values %{$PATTERNS->{countries}};
+my $branch_pat  = join '|', values %{$PATTERNS->{branches}};
+my $grade_pat   = join '|', values %{$PATTERNS->{grades}};
+my $keywords    = join '|', values %{$PATTERNS->{keywords}};
+
+my $complete_regex = qr/^(?:($country_pat)\s+)?($branch_pat)\s+(?:(?:$grade_pat)(?:\s+))?(?:$keywords)/i;
+
+triggers query_clean => $complete_regex;
```
GuiltyDolphin: Lets just use a `words` trigger here (at least for now); something like `military rank` would be fine.

moollaza:
@lowellosaurus hey there, thanks for this. Just FYI, I changed the ID to `military_rank` which is the canonical form given the package name `MilitaryRank.pm`

I've updated the link on your Pull Request. The code looks good as well

lib/DDG/Goodie/MilitaryRank.pm
```diff
@@ -0,0 +1,93 @@
...
+        group       => 'media',
+        detail      => 'false',
+        item_detail => 'false',
+        variants => { tile => 'narrow' },
+        # Scales oversize images to fit instead of clipping them.
+        elClass  => { tileMedia => 'tile__media--pr' },
+    };
+    # To protect against "Use of uninitialized value in concatenation (.) or
+    # string at .../App/DuckPAN/Web.pm line..."
+    $structured_answer->{id} = 'military_rank';
```
moollaza: This isn't necessary now that I've fixed the ID of the IA Page 👍

moollaza:
@lowellosaurus out of curiosity, did you consider implementing this as a Fathead? It feels like that would really be the ideal implementation. This is essentially a mapping of specific terms -> answers right?

If you're interested, it might be fairly easy for you to do. Especially if you used a script to build your data file

share/goodie/military_rank/military_rank.yml
```diff
@@ -0,0 +1,776 @@
...
+      - altSubtitle: CMSgt
+        image: https://upload.wikimedia.org/wikipedia/commons/e/ef/E9a_USAF_CMSGT.svg
+        subtitle: E-9 | OR-9
+        title: Chief Master Sergeant
+      - altSubtitle: CCM
+        image: https://upload.wikimedia.org/wikipedia/commons/3/36/E9c_USAF_CCMS.svg
+        subtitle: E-9 | OR-9
+        title: Command Chief Master Sergeant
+      - altSubtitle: CMSAF
+        image: "https://upload.wikimedia.org/wikipedia/commons/8/8d/E9d_USAF_CMSAF.svg\n                "
```
moollaza: this looks like it's a little broken? the `\n                      "` shouldn't be there?

lowellosaurus:
I haven't looked into implementing this as a fathead (I think I might've been spooked by the warning against first-timers making fatheads on the Docs page: http://docs.duckduckhack.com/welcome/determining-ia-type.html). I'll give it another look this evening though since I've got all the source files handy (and in a public repo: https://github.com/lowellosaurus/military_rank_helpers).

share/goodie/military_rank/military_rank.yml
```diff
@@ -0,0 +1,776 @@
+---
+ba:
+  army:
+    data:
+      - altSubtitle: ''
+        image: https://upload.wikimedia.org/wikipedia/commons/d/d3/Bosnia_and_Herzegovina_Private_Insignia.svg
```
GuiltyDolphin: @lowellosaurus Are these all `https://upload.wikimedia.org/wikipedia/commons`? If so, we could probably do the URL relative to that and just prepend the rest.

moollaza:
@lowellosaurus this is now live on Beta: https://beta.duckduckgo.com/?q=american+Army+rank+structure&ia=answer

It looks like that first image isn't working?

Overall this is looking pretty good. I think we can reduce the tile height a bit and then we should have a solid V1.

I didn't realize that the IA displaying a list of results. Seeing as that's the case, it won't work as a Fathead, however a Fathead could be built to show the image and description for a specific rank when searched.

share/goodie/military_rank/military_rank.yml
lowellosaurus: @GuiltyDolphin Hmm, sure. All of the URLs start with the same path and we could definitely do something like the following on line 77 of the module.
```
my $WIKI_URL_PREFIX = 'https://upload.wikimedia.org/wikipedia/commons/';
foreach my $entry (@{$structured_answer->{data}}) {
    $entry->{image} = $WIKI_URL_PREFIX . $entry->{image}
        if length $entry->{image} && $entry->{image} !~ /$WIKI_URL_PREFIX/;
}
```

Just a few of impressions that spring into my head about doing it this way. None of these are show-stoppers, but I figure it's worth jotting them down so you can let me know your thoughts on my impressions:
- As this instant article expands the image source may not be wikipedia (though the conditional above currently accounts for this case), so the YAML file would be a mix of full URLs and WIKI_URL_PREFIX suffixes. Even though anybody expanding on the data in this IA really should be reading the module in depth, it would not be immediately obvious why such a discrepancy might (and honestly, might not ever) exist in the data. This introduces additional cognitive load.
- We're adding another loop and minor complexity to the module that could be avoided by having a complete URL in the source YAML data.
- As I build the CSV data files from which the military_rank.yml file is then generated (https://github.com/lowellosaurus/military_rank_helpers), I just copy and paste the URLs, so removing the common WIKI_URL_PREFIX would take a little more time then. That being said, I could simply `s/$WIKI_URL_PREFIX//g` in all of those source files after pasting in the full URLs, so that's of minimal concern. Similarly, I could just the store the URL as the portion of the URL which is unique and prepend the WIKI_URL_PREFIX when the YAML file is generated to save from doing it in the module.

I don't feel strongly either way and am absolutely willing to change if you'd like the common WIKI_URL_PREFIX removed from the YAML file, or if you have a better idea about how to prepend the shared path.

lowellosaurus:
@moollaza The answer is correctly displaying no image for the first result. In the US Army, a Soldier does not wear rank insignia until they reach PV2 (which generally happens by the time they have completed all of their initial training indoctrinating them in the Army).

I definitely see what you're saying with regards to the tile height. I just tried using `description` in place of `subtitle` but that actually increased the size of the chin. I'll see if I can insert some kind of class to reduce the `margin-bottom` that seems to be causing the Leno-esque chin.

lowellosaurus:
Also, I just noticed a bug: the details pane persists and cannot be closed when you click on one of the results (https://beta.duckduckgo.com/?q=Army+rank&ia=answer).

I'm not sure if this is the result of something in the module as written (I had intended for detail view to be disabled for the time being: line 71, `detail      => 'false',`) or if it this might affect other IAs.

lib/DDG/Goodie/MilitaryRank.pm
```diff
@@ -0,0 +1,90 @@
...
+    $country = 'us' unless $country; # Default $country to us. 
+    $country = get_key_from_pattern_hash($PATTERNS->{countries}, $country);
+    $branch  = get_key_from_pattern_hash($PATTERNS->{branches}, $branch);
+
+    my $text_response = join ' ', ($DISPLAY_NAME_FOR->{$country}, $DISPLAY_NAME_FOR->{$branch}, 'Rank');
+
+    my $structured_answer = $DATA->{$country}->{$branch};
+    $structured_answer->{templates} = {
+        group       => 'media',
+        detail      => 'false',
```
moollaza: These should be be `0` instead of `'false'`

moollaza:
> @moollaza The answer is correctly displaying no image for the first result.

Understood, we should disable the image tag in that case. Chrome shows me a grey square because there's an `img` tag, but it has no content. I think passing along `false`, `null`, or `undefined` will prevent the template from trying to add an image.

> I just tried using description in place of subtitle but that actually increased the size of the chin. I'll see if I can insert some kind of class to reduce the margin-bottom that seems to be causing the Leno-esque chin.

If you use the `description` you'll probably want to set the `tileSnippet` variant to `small` (e.g. https://github.com/duckduckgo/zeroclickinfo-spice/blob/236ce11bccb378e08cea1b29d3a80046dcff82ef/share/spice/rfc/rfc.js#L37)

Another option is to use a footer template, instead of a subtitle/description.

> I'm not sure if this is the result of something in the module as written (I had intended for detail view to be disabled for the time being: line 71, detail => 'false',) or if it this might affect other IAs.

Good catch -- left a comment on the code. FWIW the `item_detail` is actually the culprit here.

lowellosaurus:
@moollaza I'm struggling to figure out how to prevent the `img` tag from appearing for items that do not have an image. I've tried passing `null`, `undef`, `0`, and `false` values (in Perl and JS), as well as not passing any image element for items without images (The docs are unclear about whether the `image` property is required for the `media_item` template--http://docs.duckduckhack.com/frontend-reference/templates-reference.html#mediaitem-template--but based on the `media_item_detail` template I think `image` should be optional).

I'd appreciate any help with removing the `img` tag when it is unnecessary if you've got ideas...

moollaza:
> The docs are unclear about whether the image property is required for the media_item template

Sorry about that. Nothing is technically `require` but this template was built to display an image, so the template file expects to get one -- hence the `img` tag is always created: https://github.com/duckduckgo/duckduckgo-answerbar-templates/blob/master/item/media_item.handlebars

I didn't realize the template doesn't check for an existing image. The simplest solution I can think of would be to use a 1px image (e.g. https://placeholdit.imgix.net/~text?txtsize=5&txt=1%C3%971&w=1&h=1&txtpad=1)

GuiltyDolphin:
@moollaza Wouldn't it be best to fix the template? It's probably better than having weird hacks attempting to circumnavigate the requirements; or have I missed something?

moollaza:
> Wouldn't it be best to fix the template?

The template isn't exactly broken though. It's purpose is to display an image tile with a text descriptor. We have a special case here where we're trying to use it without an image.

Now that I think about it a bit more, the 1x pixel isn't great either because it almost looks broken having a blank tile. I feel like we need an image that says "No Emblem" or similar so that it's clear what the empty tile actually means

GuiltyDolphin:
@moollaza Yeah - that'd be better (some indication); having a 1px image just feels like a hacky work-around ^.^ - and fair do's.

zekiel:
@lowellosaurus need any help with this?

lowellosaurus:
@zekiel, I don't think I need any help per se, maybe just some clarification on what next steps need to be taken on this instant answer? I think I've made all the requested changes, but if I missed some, could you let me know?

I am holding off on adding more countries until this instant answer is released into the wild in case I have to make changes to the data format, but I can add more countries if that's holding things up.

zekiel:
That sounds like a good plan to me @lowellosaurus Passing to @moollaza for final call.

moollaza:
@lowellosaurus sorry I hadn't noticed you pushed a commit for the "No insignia"! Deploying to Beta for final testing. 

Otherwise, this LGTM 👍

share/goodie/military_rank/military_rank.js
```diff
@@ -0,0 +1,12 @@
+DDH.military_rank = DDH.military_rank || {};
+var no_insignia_svg = DDG.get_asset_path('military_rank', 'no_insignia.svg');
```
moollaza: We should move this into the `build` function so we're not polluting the global namespace 👍

moollaza: Actually sorry, this approach isn't going to work in Production. The path generated is versioned according to the Spice repo. The Goodie repo has a different version number so it won't work in production.

moollaza: You'll want to grab the Goodie version number in the Perl and either define the path in the Perl, or pass the version using the `data` hash and build the path in the JS.

This is an example of a Goodie grabbing the version number: https://github.com/duckduckgo/zeroclickinfo-goodies/blob/master/lib/DDG/Goodie/SolarSystem.pm#L14

lowellosaurus: Updated!

moollaza:
@lowellosaurus thanks for updating! Deploying to Beta...

moollaza:
@lowellosaurus Thanks! This LGTM :+1:

daxtheduck:
@lowellosaurus this is now merged! It should be live on DuckDuckGo within one week (or during the next release) and we'll be sure to notify you when it happens.

Issue #2643, opened 2016-02-27
Regexp: Improve display for mobile devices


GuiltyDolphin:
Unfortunately in getting `Regexp` to display nicely on full-width screens, it seems the mobile display has been reduced:

![screenshot from 2016-02-27 12 16 30](https://cloud.githubusercontent.com/assets/8598426/13372810/f73af9aa-dd4b-11e5-915e-ead68ec32414.png)

VS

![screenshot from 2016-02-27 12 16 37](https://cloud.githubusercontent.com/assets/8598426/13372812/fd4138b4-dd4b-11e5-9bf8-e0a3be9c327e.png)

What needs to be done:
- Fix the mobile display so that both the pattern names **_and**_ matches consistently display well.
- Ensure the full-screen display is robust and won't be messed up by long queries. 

---

IA Page: https://duck.co/ia/view/regexp

jagtalon:
Oh hey I'm the maintainer!

GuiltyDolphin:
@jagtalon Apparently yes! Any thoughts?

jagtalon:
Uhm, that it should be fixed? Hahaha

I kid. I'll look into it

NaveenKarippai:
@GuiltyDolphin @jagtalon If there has been no update on this issue so far, I'd be interested to try some CSS magic here.

GuiltyDolphin:
@NaveenKarippai Go for it - I'm in need of a CSS magician.

holtjohnson:
Bumping this issue.

Is there anything else that needs to be done for this to be finished? I'm very adept at CSS, should it need any finishing touches.
@GuiltyDolphin, @jagtalon, & @NaveenKarippai

GuiltyDolphin:
@holtjohnson The PR is over on https://github.com/duckduckgo/zeroclickinfo-goodies/pull/3154 if you wanted to take a look.

@moollaza What's the status with the PR? Are there any outstanding issues?

moollaza:
> @moollaza What's the status with the PR? Are there any outstanding issues?

Yes, the PR should not add any media queries. If we simply remove the set width, this should mostly be fixed. There's a separate issue that needs to be fixed internally with our `chomp()` function

holtjohnson:
@moollaza I'll make the changes and create a new PR.

moollaza:
@holtjohnson thans a lot!

holtjohnson:
Made the changes and created a PR.

[...]

PR #3154, opened 2016-05-28
Regexp: Responsive table design

NaveenKarippai:

your Pull Request.
- [ ] New Instant Answer
- [x] Improvement
  - [x] Enhancement: **`regexp: {2643} make table output responsive in design`**
- [ ] Non–Instant Answer
###### Description of changes

The goodie output table was not responsive on small screen
devices like tablet on mobile phones. This could be fixed by
some CSS styling magic.

Provide an overview of the changes this pull request introduces.
- CSS media was used to detect screen width
- responsive on small screen < 760px

See also: #2643

Fixes #NNNN - how specifically does it fix the issue?
###### People to notify (@GuiltyDolphin, @moollaza, @tagawa)

---

Instant Answer Page: https://duck.co/ia/view/regexp

[Maintainer](http://docs.duckduckhack.com/maintaining/guidelines.html): @jagtalon

### DIFF
--- a/share/goodie/regexp/regexp.css
+++ b/share/goodie/regexp/regexp.css
@@ -1,5 +1,4 @@
 .zci--regexp .record .record__cell--key {
-    width: auto;
     text-transform: none;
 }
 
--- a/share/goodie/regexp/regexp.css
+++ b/share/goodie/regexp/regexp.css
@@ -7,7 +7,7 @@
 @media
 only screen and (max-width: 760px),
 (min-device-width: 768px) and (max-device-width: 1024px)  {
- td.record__cell.record__cell--key, td.record__cell.record__cell--value {
+.zci--regexp .record .record__cell {
         display: block;
         border-top: 1px solid #d0d0d0;
 	border-bottom: 1px solid #eee;
--- a/share/goodie/regexp/regexp.css
+++ b/share/goodie/regexp/regexp.css
@@ -1,5 +1,5 @@
 .zci--regexp .record .record__cell--key {
-    width: 20em;
+    width: auto;
     text-transform: none;
 }
 
--- a/share/goodie/regexp/regexp.css
+++ b/share/goodie/regexp/regexp.css
@@ -2,3 +2,17 @@
     width: 20em;
     text-transform: none;
 }
+
+/* responsive table */
+@media
+only screen and (max-width: 760px),
+(min-device-width: 768px) and (max-device-width: 1024px)  {
+ td.record__cell.record__cell--key, td.record__cell.record__cell--value {
+        display: block;
+        border-top: 1px solid #d0d0d0;
+	border-bottom: 1px solid #eee;
+	position: relative;
+	padding-right: 60%;
+        overflow:visible;
+ }
+}
--- a/share/goodie/regexp/regexp.css
+++ b/share/goodie/regexp/regexp.css
@@ -6,12 +6,12 @@
 @media
 only screen and (max-width: 760px),
 (min-device-width: 768px) and (max-device-width: 1024px)  {
-.zci--regexp .record .record__cell {
+    .zci--regexp .record .record__cell {
         display: block;
         border-top: 1px solid #d0d0d0;
-	border-bottom: 1px solid #eee;
-	position: relative;
-	padding-right: 60%;
+        border-bottom: 1px solid #eee;
+        position: relative;
+        padding-right: 60%;
         overflow:visible;
- }
+    }
 }


daxtheduck:
Metadata placeholder post
            
---
            
**How to find your IA page**
- New Instant Answer
              
- Existing Instant Answer

share/goodie/regexp/regexp.css
```diff
@@ -2,3 +2,17 @@
...
 }
+
+/* responsive table */
+@media
+only screen and (max-width: 760px),
+(min-device-width: 768px) and (max-device-width: 1024px)  {
+.zci--regexp .record .record__cell {
+        display: block;
+        border-top: 1px solid #d0d0d0;
+	border-bottom: 1px solid #eee;
```
moollaza: nit: indentation isn't consistent

moollaza:
Deploying to beta...

moollaza:
@NaveenKarippai it seems like this fix may be a little overkill -- We seem to have the width of the first `<td>` set to 20em, ignoring the width of the page.

It may be better to set the width to a max of 35-50%, and right align the 2nd `<td>`

We can also save a lot of horizontal space by rewording `Subpattern Match` to `Group`

GuiltyDolphin:
@moollaza This is the main case I am concerned with https://beta.duckduckgo.com/?q=regexp+%2F%28.*%29%28%3F%3Cfooooooooooooooooooooooooooooo%3E.%29%2F+ddg&ia=answer

(And pretty much everything on mobile).

I'd rather the name stayed as `Subpattern Match` on desktop - though maybe `Group` would be more suitable on mobile (feel free to veto).

I initially hard-coded the width as it seemed a pretty sensible amount, providing a fairly nice display for most sensible-length names; unfortunately it does not do anything good for the eyes on mobile.

moollaza:
> This is the main case I am concerned with https://beta.duckduckgo.com/?q=regexp+%2F%28.*%29%28%3F%3Cfooooooooooooooooooooooooooooo%3E.%29%2F+ddg&ia=answer

I understand where you're going, but I think that's a _bit_ of an edge case? Obviously at some point we have to clip it, but we've got room to expand the desktop UI.

> I'd rather the name stayed as Subpattern Match on desktop - though maybe Group would be more suitable on mobile (feel free to veto).

The reason I suggested `Group` is because most of the Regex testing tools I've used seem to also use that terminology. "Capture group" is more descriptive, but again lengthy. Either way, the terminology should be consistent across designs though.

I can have the design team take a look as well?

GuiltyDolphin:
@moollaza Yeah, some expert opinion wouldn't go amiss :)

moollaza:
@NaveenKarippai @GuiltyDolphin -- design team is a little busy, but there's a simple fix to the immediate problem which is that we have a fixed width of `20em` on mobile. If we remove that content that _should_ fit in the space provided, _will_ fit.

Let's just remove that fixed width for now and then figure out how to handle edge cases where the text is wider than the screen.

moollaza:
regexp________ddg_at_duckduckgo

@NaveenKarippai you can use a rule like:

```css
.is-mobile .zci--regexp .record__cell--key {
    width: auto;
}
```

NaveenKarippai:
@moollaza Could you please review the new changset.

Should we keep the css `media` query?

moollaza:
> Should we keep the css media query?

Yes please. The table _should_ be providing a "show more" button whenever the content is being clipped to allow it to expand which should alleviate the need for the media query

NaveenKarippai:
@moollaza understood. In that case, the current changes  should be sufficient. Could you please make an update if it's ready to merge. Thanks!

moollaza:
@NaveenKarippai whoops just realized I misread your message your question above -- thought you asked if we should **_remove_** the media query -- which we should After that we can merge this.

The search team will need to update the table to make sure we show a "Show more" when necessary

share/goodie/regexp/regexp.css
```diff
@@ -1,4 +1,18 @@
 .zci--regexp .record .record__cell--key {
-    width: 20em;
+    width: auto;
```
moollaza: I think the default is `auto`? We can just remove this rule and it should have the same effect I think?

NaveenKarippai:
@moollaza hm, we have removed unnecessary style properties as you pointed out. I'm not sure about how to proceed further by removing media queries. If there is a similar IA to take inspiration from, it would be awesome!

share/goodie/regexp/regexp.css
```diff
@@ -1,4 +1,17 @@
...
+@media
+only screen and (max-width: 760px),
+(min-device-width: 768px) and (max-device-width: 1024px)  {
+.zci--regexp .record .record__cell {
+        display: block;
+        border-top: 1px solid #d0d0d0;
+	border-bottom: 1px solid #eee;
+	position: relative;
+	padding-right: 60%;
+        overflow:visible;
```
GuiltyDolphin: Nd here.

GuiltyDolphin:
@moollaza @NaveenKarippai This looks good on full size:

screenshot-2016-06-18-19 17 00_745_281

And 'default' view for mobile:

screenshot-2016-06-18-19 17 26_369_514

But if you click 'Show More', it screws up the names:

screenshot-2016-06-18-19 17 36_363_506

I think we should make sure that name stays on the same line regardless of whether they pressed 'Show More' or not.

share/goodie/regexp/regexp.css
NaveenKarippai: done!

moollaza:
@NaveenKarippai all we need to do is remove the media query. On mobile, there should be a "Show More" button when the content is clipped. Clicking it will expand the table vertically and the column text will wrap.

On desktop we're not always doing that when we should though and that needs to be fixed in the template. I think the JS code is internal though.

NaveenKarippai:
@moollaza Thanks for the update!

> I think the JS code is internal though

Do I have an access to internal JS code to make the changes?

moollaza:
> Do I have access on internal JS code to make the changes?

Unfortunately not. By internal I mean closed source. We're still working to open source more of the code, but we haven't gotten around to some parts of the template code yet.

Let's get this finished and merged and I can get the team to adjust the internal code.

mintsoft:
Fixed by : https://github.com/duckduckgo/zeroclickinfo-goodies/issues/2643 https://github.com/duckduckgo/zeroclickinfo-goodies/pull/3586

Issue #3169, opened 2016-06-01
Random Number: Don't trigger if lower and upper bounds are the same


GuiltyDolphin:
For example ['random 7 7'](https://duckduckgo.com/?q=random+7+7&ia=answer) and ['random 1'](https://duckduckgo.com/?q=random+1&ia=answer). We cannot provide a 'random' number when the bounds are the same.

---

IA Page:  http://duck.co/ia/view/random_number 
[Maintainer](http://docs.duckduckhack.com/maintaining/guidelines.html): @jonk1993

jmwoliver:
@GuiltyDolphin I can take this one and https://github.com/duckduckgo/zeroclickinfo-goodies/issues/3168 if you would like. They are pretty much identical to the one you just helped me finish.

pjhampton:
Hey all

[DuckDuckHack is now in Maintenance Mode](https://duckduckhack.com/) and from now on, we are only accepting issues and PRs for essential bugs and bug fixes.

Unfortunately, with the above in mind, this isn't something we can action and will be closed.

We appreciate you taking the time to contribute and apologize for not being able to triage this issue.

PR #3171, opened 2016-06-01
Random Number: Fixed #3168 and #3169, limit triggers to only when you have two different numbers

### LINKED ISSUE #3168: Random Number: Prevent triggering on just 'random'
See ['random'](https://duckduckgo.com/?q=random&ia=answer) - we shouldn't trigger on such a generic query.

---

IA Page:  http://duck.co/ia/view/random_number 
[Maintainer](http://docs.duckduckhack.com/maintaining/guidelines.html): @jonk1993

jmwoliver:

your Pull Request.
- [ ] New Instant Answer
- [X] Improvement
  - [X] Bug fix: **`Random Number: Fix #3168 and #3169`**
- [ ] Non–Instant Answer
###### Description of changes

Limit triggers to only when you have two different numbers. Don't trigger if lower and upper bounds are the same, there is only one number, or no numbers present.

Fixes #3168 and #3169 - Limit triggers to only when you have two different numbers

@GuiltyDolphin 

---

Instant Answer Page: https://duck.co/ia/view/random_number

[Maintainer](http://docs.duckduckhack.com/maintaining/guidelines.html): @jonk1993

### DIFF
--- a/lib/DDG/Goodie/RandomNumber.pm
+++ b/lib/DDG/Goodie/RandomNumber.pm
@@ -17,6 +17,8 @@ handle query_lc => sub {
 
     my $start = $1 || 0;
     my $end   = $2 || 0;
+    
+    return unless ($start != $end) && ($end != undef);
 
     $start = 1000000000 if $start > 1000000000;
     $start = 0          if $start < 0;
@@ -37,6 +39,7 @@ handle query_lc => sub {
         $rand *= ($valDiff + 1);
         $rand = int($rand) + $start;
     }
+    
 
     return "$rand (random number)",
       structured_answer => {
--- a/lib/DDG/Goodie/RandomNumber.pm
+++ b/lib/DDG/Goodie/RandomNumber.pm
@@ -24,7 +24,7 @@ handle query_lc => sub {
     
     #makes sure user inputs two numbers
     return if ($start != $end) && ($end == 0);
-    #  and both numbers are not zero
+    #or both numbers are zero
     return if ($start == $end) && ($end == 0);
 
     $start = 1000000000 if $start > 1000000000;
--- a/lib/DDG/Goodie/RandomNumber.pm
+++ b/lib/DDG/Goodie/RandomNumber.pm
@@ -47,7 +47,7 @@ handle query_lc => sub {
         $rand *= ($valDiff + 1);
         $rand = int($rand) + $start;
     }
-    
+
     return "$rand (random number)",
       structured_answer => {
         data => {
--- a/lib/DDG/Goodie/RandomNumber.pm
+++ b/lib/DDG/Goodie/RandomNumber.pm
@@ -22,10 +22,9 @@ handle query_lc => sub {
     my $start = $1 || 0;
     my $end   = $2 || 0;
     
-    #makes sure user inputs two numbers
+    #makes sure user inputs two numbers, and both numbers are not zero
     return if ($start != $end) && ($end == 0);
-    #or both numbers are zero
-    return if ($start == $end) && ($end == 0);
+    return if ($start == $end) && ($end == 0) && !(($_ =~ /^\!?(?:rand(?:om|)(?: num(?:ber|)|))$/i));
 
     $start = 1000000000 if $start > 1000000000;
     $start = 0          if $start < 0;
--- a/lib/DDG/Goodie/RandomNumber.pm
+++ b/lib/DDG/Goodie/RandomNumber.pm
@@ -17,7 +17,6 @@ handle query_lc => sub {
 
     my $start = $1 || 0;
     my $end   = $2 || 0;
-    
     return unless ($start != $end) && ($end != undef);
 
     $start = 1000000000 if $start > 1000000000;
@@ -40,7 +39,6 @@ handle query_lc => sub {
         $rand = int($rand) + $start;
     }
     
-
     return "$rand (random number)",
       structured_answer => {
         data => {
--- a/lib/DDG/Goodie/RandomNumber.pm
+++ b/lib/DDG/Goodie/RandomNumber.pm
@@ -24,6 +24,8 @@ handle query_lc => sub {
     
     #makes sure user inputs two numbers
     return if ($start != $end) && ($end == 0);
+    #  and both numbers are not zero
+    return if ($start == $end) && ($end == 0);
 
     $start = 1000000000 if $start > 1000000000;
     $start = 0          if $start < 0;
--- a/lib/DDG/Goodie/RandomNumber.pm
+++ b/lib/DDG/Goodie/RandomNumber.pm
@@ -13,11 +13,17 @@ handle query_lc => sub {
     srand();
     # Random number.
     # q_check (as opposed to q_internal) Allows for decimals.
+    
+    # checks if only 'rand' or 'random' is entered
+    return if ($_ =~ /^\!?(?:rand(?:om|))$/i);
+    
     return unless ($_ =~ /^\!?(?:rand(?:om|)(?: num(?:ber|)|)(?: between|))( [\d\.]+|)(?: and|)( [\d\.]+|)$/i);
 
     my $start = $1 || 0;
     my $end   = $2 || 0;
-    return unless ($start != $end) && ($end != undef);
+    
+    #makes sure user inputs two numbers
+    return if ($start != $end) && ($end == 0);
 
     $start = 1000000000 if $start > 1000000000;
     $start = 0          if $start < 0;
@@ -31,6 +37,9 @@ handle query_lc => sub {
     ($end, $start) = ($start, $end) if ($start > $end);
 
     my $valDiff = $end - $start;
+    
+    #returns if both end and start are equal, and end has been changed by the user
+    return if ($valDiff == 0) && ($end != 0);
 
     my $rand = rand;
 
--- a/t/RandomNumber.t
+++ b/t/RandomNumber.t
@@ -41,7 +41,7 @@ ddg_goodie_test(
     'random number between 1 and '    => undef,
     'random 10'                       => undef,
     'random 7 7'                      => undef,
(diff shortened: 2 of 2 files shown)


daxtheduck:
Metadata placeholder post
            
---
            
**How to find your IA page**
- New Instant Answer
              
- Existing Instant Answer

jmwoliver:
@GuiltyDolphin the IA page may need to be changed because it says 'Random Number' is a viable trigger.

lib/DDG/Goodie/RandomNumber.pm
```diff
@@ -17,6 +17,7 @@ handle query_lc => sub {
 
     my $start = $1 || 0;
     my $end   = $2 || 0;
+    return unless ($start != $end) && ($end != undef);
```
jonk1993: So I think this may have to be handled a bit differently. The lines 

```
 my $start = $1 || 0;
 my $end   = $2 || 0;
```

Will guarantee that `$start` and `$end` will be set to `0` if their regex groups are not matched. So `($end != undef)` will never be true, as `$end` will default to `0`.

Since we aren't including negatives as valid input, one possibility is to change `$start` and `$end` initialization to:

```
 my $start = $1 || -1;
 my $end   = $2 || -1;
```

and change your return statement to return unless `$end` does not equal `-1`.

GuiltyDolphin:
@jmwoliver Yes - `random number` *is* a valid trigger; it is `random` on its own that I'm concerned about.

lib/DDG/Goodie/RandomNumber.pm
GuiltyDolphin: I think just doing `return if ($start == $end)` should be fine - we already ensure they aren't undefined (as @jonk1993).

Btw, @jonk1993 why are we using `query_lc` rather than `remainder_lc`? Assuming the triggerword is `random` we can just check the remainder for `number`.

jmwoliver:
@jonk1993 Okay, that makes sense, I'll try it that way. @GuiltyDolphin Thank you for clarifying. I was wondering, would "random 1 10" still be a valid input or should it only trigger on "random number 1 10"?

GuiltyDolphin:
@jmwoliver I think `random 1 10` should still trigger - the cases that should not trigger are:

1. `random` (on its own)
2. Anything that causes the lower and upper bounds to be the same.

jmwoliver:
@GuiltyDolphin It should be good now. It fails on just 'rand'/'random' being passed in, and it fails when only one number is given and when both numbers given are the same. And it still works if you just put 'random number'. It also still works with 'random 1 10'. It looks like some test cases failed, I'll give it a look.

lib/DDG/Goodie/RandomNumber.pm
```diff
@@ -30,14 +37,17 @@ handle query_lc => sub {
     ($end, $start) = ($start, $end) if ($start > $end);
 
     my $valDiff = $end - $start;
+    
+    #returns if both end and start are equal, and end has been changed by the user
+    return if ($valDiff == 0) && ($end != 0);
```
GuiltyDolphin: Why do we have so many checks? One check of `return if $start == $end` should suffice.

lib/DDG/Goodie/RandomNumber.pm
```diff
@@ -30,14 +37,17 @@ handle query_lc => sub {
...
 
     my $rand = rand;
 
     if ($start && $end || $valDiff > 1) {
         $rand *= ($valDiff + 1);
         $rand = int($rand) + $start;
     }
-
+
```
GuiltyDolphin: Please ensure you don't add any unnecessary whitespace.

lib/DDG/Goodie/RandomNumber.pm
```diff
@@ -13,10 +13,17 @@ handle query_lc => sub {
...
+    # checks if only 'rand' or 'random' is entered
+    return if ($_ =~ /^\!?(?:rand(?:om|))$/i);
+    
     return unless ($_ =~ /^\!?(?:rand(?:om|)(?: num(?:ber|)|)(?: between|))( [\d\.]+|)(?: and|)( [\d\.]+|)$/i);
 
     my $start = $1 || 0;
     my $end   = $2 || 0;
+    
+    #makes sure user inputs two numbers
+    return if ($start != $end) && ($end == 0);
```
GuiltyDolphin: This will be caught by the regex.

lib/DDG/Goodie/RandomNumber.pm
```diff
@@ -13,10 +13,17 @@ handle query_lc => sub {
     srand();
     # Random number.
     # q_check (as opposed to q_internal) Allows for decimals.
+    
+    # checks if only 'rand' or 'random' is entered
+    return if ($_ =~ /^\!?(?:rand(?:om|))$/i);
+    
     return unless ($_ =~ /^\!?(?:rand(?:om|)(?: num(?:ber|)|)(?: between|))( [\d\.]+|)(?: and|)( [\d\.]+|)$/i);
```
GuiltyDolphin: I think this regex is causing some issues, if we clean it up we shouldn't need the checks for additional numbers - having `blah|` in each case it allowing too many invalid forms (for example, if `and` is present, we would *expect* there to be a second number).

jmwoliver:
@GuiltyDolphin I took out the spaces and recommitted and Travis likes it now for some reason. The reason I can't just do 'return if ($start == $end)' is because both are initialized to 0 if not given a value. So if you only search "random number" which is a valid search, then $start == $end and that return statement would not trigger this IA. You are probably right that there is a better way to do it, so I can change it if you'd like. But the way I have it right now fulfills all the parameters for this IA.

t/RandomNumber.t
```diff
@@ -36,6 +36,13 @@ ddg_goodie_test(
     'random number between 0 and 1'   => build_test(0, 1, qr/$RE{num}{real}/),
     'random number between 0 and 10'  => build_test(0, 10, qr/\d{1,2}/),
     'random number between 0 and 100' => build_test(0, 100, qr/\d{1,3}/),
+
+    #cases that don't have any numbers, only one number, or both numbers the same should fail
+    'random number between 1 and '    => undef,
+    'random 10'                       => undef,
+    'random 7 7'                      => undef,
+    'random'                          => undef,
```
GuiltyDolphin: Lets add a test for `random 0 0`.

jmwoliver:
@GuiltyDolphin What is the desired behavior for that? Right now it's finding a random between 0 and 1 but outputting 0 - 0. Obviously that is very wonky and not what we want.
aoeu

GuiltyDolphin:
@jmwoliver It falls into the case of upper and lower bounds being the same (though it seems like another bug has unearthed itself); so it should not trigger.

jmwoliver:
@GuiltyDolphin  I just added `return if ($start == $end) && ($end == 0);` and it fixed the issue. I will commit it with the new test case.

jmwoliver:
@GuiltyDolphin Okay I think it is finally good to go. My solution was really make-shift though by adding `return if ($start == $end) && ($end == 0) && !(($_ =~ /^\!?(?:rand(?:om|)(?: num(?:ber|)|))$/i));` so it would fail if both numbers entered are 0 but not if you just type 'random number'.

lib/DDG/Goodie/RandomNumber.pm
```diff
@@ -13,10 +13,18 @@ handle query_lc => sub {
...
 
     my $start = $1 || 0;
     my $end   = $2 || 0;
+    
+    #makes sure user inputs two numbers, and both numbers are not zero
+    return if ($start != $end) && ($end == 0);
+    return if ($start == $end) && ($end == 0) && !(($_ =~ /^\!?(?:rand(?:om|)(?: num(?:ber|)|))$/i));
```
GuiltyDolphin: I think it might be best just to use the `remainder` handler - then we can just check `matched_trigger` and `remainder` to see what sort of query should be handled.

lib/DDG/Goodie/RandomNumber.pm
jonk1993: Ah, I'm not the creator just became the maintainer a week or so ago. Changing to `remainder_lc` seems to makes sense.

jmwoliver:
@GuiltyDolphin Could you explain further on how to use `remainder_lc` or point me to an IA that uses it how you are suggesting?

GuiltyDolphin:
@jmwoliver I'd reccommend the [docs](http://docs.duckduckhack.com/backend-reference/handle-functions.html).

jmwoliver:
@GuiltyDolphin Sorry, I still haven't figured out how to implement `remainder_lc`. I'll keep looking at it but I couldn't figure it out from the docs, is there anything else you could provide?

GuiltyDolphin:
@jmwoliver There's not much to it - just have a look at some existing Goodies and how they use the handlers in different ways; the docs do a fairly thorough job of explaining though.

zekiel:
@jmwoliver  still need some help?

jmwoliver:
@zekiel Yes please. I still couldn't figure it out by looking at the documentation and examples.

gaulrobe:
@jmwoliver 

Right now, the handler being used is `query_lc` which will give you the complete query that was used to trigger the IA. So if the query is `number between 0 and 10` the incoming value you have to work with (`$_`) will be `number between 0 and 10`.

If you switch the handler to `remainder_lc` you will get the query minus the trigger. So in the above example, `$_` would be `between 0 and 10`. What this means is that you don't have to do additional trigger validation. Just validation on the rest of the query. This _should_ simplify the regex being used.

Hope this helps!

mintsoft:
@jmwoliver Hiya, just checking in. Are you OK continuing with this? If not we can open it up to the community :+1:

moollaza:
This PR is being closed due to inactivity. If you're still interested in finishing, please let me know and I'll gladly re-open it for you!

PR #3352, opened 2016-07-05
Add new JSMinify Instant Answer

sahildua2305:

your Pull Request.

- [x] New Instant Answer
    - [x] Other: **`New JS Minify Instant Answer`**
- [ ] Improvement
- [ ] Non–Instant Answer

###### Description of changes
Developing an interactive JS Minifier using [this library](http://prettydiff.com/prettydiff.js). *Work in progress*

@moollaza @MariagraziaAlastra 

---

Instant Answer Page: https://duck.co/ia/view/js_minify

### DIFF
--- a/lib/DDG/Goodie/JsMinify.pm
+++ b/lib/DDG/Goodie/JsMinify.pm
@@ -1,0 +1,31 @@
+package DDG::Goodie::JsMinify;
+# ABSTRACT: An Interactive JavaScript Minifier Tool
+
+use DDG::Goodie;
+use strict;
+
+zci answer_type => 'jsminify';
+zci is_cached => 1;
+
+triggers any => 'js minify', 'js minifier';
+
+handle query_lc => sub {
+
+    return unless $_;
+
+    return '',
+        structured_answer => {
+
+            data => {},
+
+            templates => {
+                group => 'text',
+                item => 0,
+                options => {
+                    content => 'DDH.js_minify.content'
+                }
+            }
+        };
+};
+
+1;
--- a/lib/DDG/Goodie/JsMinify.pm
+++ b/lib/DDG/Goodie/JsMinify.pm
@@ -18,8 +18,6 @@ handle query_lc => sub {
 
             id => "js_minify",
 
-            name => "Minifier",
-
             data => {},
 
             templates => {
--- a/lib/DDG/Goodie/JsMinify.pm
+++ b/lib/DDG/Goodie/JsMinify.pm
@@ -16,6 +16,10 @@ handle query_lc => sub {
     return '',
         structured_answer => {
 
+            id => "js_minify",
+
+            name => "Minifier",
+
             data => {},
 
             templates => {
--- a/lib/DDG/Goodie/JsMinify.pm
+++ b/lib/DDG/Goodie/JsMinify.pm
@@ -7,7 +7,7 @@ use strict;
 zci answer_type => 'jsminify';
 zci is_cached => 1;
 
-triggers any => 'js minify', 'js minifier';
+triggers any => 'js minify', 'js minifier', 'javascript minify', 'javascript minifier', 'minify js', 'minify javascript', 'minifier js', 'minifier javascript';
 
 handle query_lc => sub {
 
--- a/share/goodie/js_minify/content.handlebars
+++ b/share/goodie/js_minify/content.handlebars
@@ -1,1 +1,3 @@
-<h1>JS Minifier</h1>
+<div>
+    <span>JS Minifier</span>
+</div>
--- a/share/goodie/js_minify/content.handlebars
+++ b/share/goodie/js_minify/content.handlebars
@@ -1,7 +1,7 @@
 <h5>Minify JavaScript Code on the go!</h5>
 <div class="js_minify__io_box">
     <textarea name="js_minify__input" id="js_minify__input" rows="20" cols="60"></textarea>
-    <textarea name="js_minify__output" id="js_minify__output" rows="20" cols="60"></textarea><br>
+    <textarea name="js_minify__output" id="js_minify__output" rows="20" cols="60" readonly="readonly" onclick="this.focus();this.select()"></textarea><br>
 </div>
 <div class="js_minify__action_box">
     <button class="js_minify__action">Minify</button>
--- a/share/goodie/js_minify/content.handlebars
+++ b/share/goodie/js_minify/content.handlebars
@@ -1,3 +1,8 @@
-<div>
-    <span>JS Minifier</span>
+<h5>Minify JavaScript Code on the go!</h5>
+<div class="js_minify__io_box">
+    <textarea name="js_minify__input" id="js_input" rows="20" cols="60"></textarea>
+    <textarea name="js_minify__output" id="js_output" rows="20" cols="60"></textarea><br>
+</div>
+<div class="js_minify__action_box">
+    <button class="js_minify__action">Minify</button>
 </div>
--- a/share/goodie/js_minify/content.handlebars
+++ b/share/goodie/js_minify/content.handlebars
@@ -1,0 +1,1 @@
+<h1>JS Minifier</h1>
--- a/share/goodie/js_minify/content.handlebars
+++ b/share/goodie/js_minify/content.handlebars
@@ -1,7 +1,7 @@
 <h5>Minify JavaScript Code on the go!</h5>
 <div class="js_minify__io_box">
-    <textarea name="js_minify__input" id="js_input" rows="20" cols="60"></textarea>
-    <textarea name="js_minify__output" id="js_output" rows="20" cols="60"></textarea><br>
+    <textarea name="js_minify__input" id="js_minify__input" rows="20" cols="60"></textarea>
+    <textarea name="js_minify__output" id="js_minify__output" rows="20" cols="60"></textarea><br>
 </div>
 <div class="js_minify__action_box">
     <button class="js_minify__action">Minify</button>
--- a/share/goodie/js_minify/js_minify.css
+++ b/share/goodie/js_minify/js_minify.css
@@ -1,0 +1,21 @@
+.zci--js_minify .js_minify__action_box {
+    margin-top: 3.25em;
+}
+
+.zci--js_minify .js_minify__action {
+    background: #5DA4DC;
+    color: #fff;
+    text-decoration: none;
+    font-size: 1.25em;
+    padding: 10px 15px;
+    border: none;
+    border-radius: 4px;
+    position: absolute;
+    bottom: 0;
+    margin-bottom: .25em;
+    cursor: pointer;
+}
+
+.zci--js_minify .js_minify__action:active {
+    background: #4793D0;
+}
--- a/share/goodie/js_minify/js_minify.css
+++ b/share/goodie/js_minify/js_minify.css
@@ -19,3 +19,8 @@
 .zci--js_minify .js_minify__action:active {
     background: #4793D0;
 }
+
+.zci--js_minify #js_minify__input,
+.zci--js_minify #js_minify__output {
+    width: 45%;
+}
--- a/share/goodie/js_minify/js_minify.js
+++ b/share/goodie/js_minify/js_minify.js
@@ -8,7 +8,6 @@ DDH.js_minify.build = function(ops) {
 
     return {
         onShow: function() {
-            console.log("onShow called");
             // make sure this function is run only once, the first time
             // the IA is shown otherwise things will get initialized more than once
             if (shown)
@@ -17,10 +16,19 @@ DDH.js_minify.build = function(ops) {
             // set the flag to true so it doesn't get run again
             shown = true;
 
+            $.getScript('')
+            
             var $dom = $('#zci-js_minify'),
-                $main = $dom.find('.zci__main');
+                $main = $dom.find('.zci__main'),
+                $minifyButton = $dom.find('.js_minify__action');
 
             $main.toggleClass('c-base');
+
+            var minifyCode = function() {
+                console.log("Call for minifyCode");
+            };
+
+            $minifyButton.click(minifyCode);
 
         }
     };
--- a/share/goodie/js_minify/js_minify.js
+++ b/share/goodie/js_minify/js_minify.js
@@ -1,14 +1,27 @@
 DDH.js_minify = DDH.js_minify || {};
 
-(function(DDH) {
-    "use strict";
+"use strict";
 
-    DDH.js_minify.build = function(ops) {
+DDH.js_minify.build = function(ops) {
 
-        return {
-            onShow: function() {
+    var shown = false;
 
-            }
-        };
-    }
-});
+    return {
+        onShow: function() {
+            console.log("onShow called");
+            // make sure this function is run only once, the first time
+            // the IA is shown otherwise things will get initialized more than once
+            if (shown)
+                return;
+
+            // set the flag to true so it doesn't get run again
+            shown = true;
+
+            var $dom = $('#zci-js_minify'),
+                $main = $dom.find('.zci__main');
+
+            $main.toggleClass('c-base');
+
+        }
+    };
+};
--- a/share/goodie/js_minify/js_minify.js
+++ b/share/goodie/js_minify/js_minify.js
@@ -18,13 +18,27 @@ DDH.js_minify.build = function(ops) {
 
             var $dom = $('#zci-js_minify'),
                 $main = $dom.find('.zci__main'),
-                $minifyButton = $dom.find('.js_minify__action');
+                $minifyButton = $dom.find('.js_minify__action'),
+                $input = $dom.find('#js_minify__input'),
+                $output = $dom.find('#js_minify__output');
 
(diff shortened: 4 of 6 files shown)


daxtheduck:
Metadata placeholder post
            
---
            
**How to find your IA page**
- New Instant Answer
              
- Existing Instant Answer

lib/DDG/Goodie/JsMinify.pm
```diff
@@ -0,0 +1,31 @@
+package DDG::Goodie::JsMinify;
+# ABSTRACT: An Interactive JavaScript Minifier Tool
+
+use DDG::Goodie;
+use strict;
+
+zci answer_type => 'jsminify';
+zci is_cached => 1;
+
+triggers any => 'js minify', 'js minifier';
```
MariagraziaAlastra: @sahildua2305 I think we'd also want this to trigger for queries like "javascript minify, javascript minifier, minify js, minify javascript"

MariagraziaAlastra:
@sahildua2305 Thanks for working on this! Please let me know if you need any help :smile:

lib/DDG/Goodie/JsMinify.pm
sahildua2305: Definitely! This is just the beginning. I have added these trigger words for testing. I'll add more. Thanks for suggesting these words. :smile:

sahildua2305:
@MariagraziaAlastra Actually I need help in getting started with this. With this initial setup, I am not able to show anything on search page. The IA gets triggered but doesn't show anything. Not even an IA tab.

Can you please look into this? I may be missing something naive.

MariagraziaAlastra:
@sahildua2305 It actually triggers for me: 
screenshot-devour-sertum codio io 5000 2016-07-05 22-55-58

sahildua2305:
Weird! I'll check again. Thanks a lot.

sahildua2305:
@MariagraziaAlastra I have pushed a working interactive js minifier (with very basic UI) and it looks like this - 
selection_154

Please deploy this on Beta. :smile:

share/goodie/js_minify/js_minify.js
```diff
@@ -0,0 +1,46 @@
...
+
+    return {
+        onShow: function() {
+            // make sure this function is run only once, the first time
+            // the IA is shown otherwise things will get initialized more than once
+            if (shown)
+                return;
+
+            // set the flag to true so it doesn't get run again
+            shown = true;
```
gaulrobe: This `shown` logic seems like it is indicative of a different, underlying problem... does `onShow` get called multiple times?

sahildua2305: `onShow` get called every time the `Answer` tab is shown.

For example, if user searches for `js minify` by default `Answer` tab will be opened. If user clicks on `Web` tab and then comes back to `Answer` tab, `onShow` will be called again.

gaulrobe: I see this is done in other javascript, but it just seems like something that should be built-in at a lower level. Carry on!

sahildua2305: Agreed! :+1:

MariagraziaAlastra:
@sahildua2305 Hey, nice job here, thanks! 
Your PR is deployed to beta, but it doesn't trigger yet. Will let you know when we have the library installed on our end, so you can use DDG.require() instead - it shouldn't take long anyway. 😄 

I tested it on DuckPAN in the meantime and it works really well. As discussed on Slack, why don't you make the tab name be "Minifier" instead? Maybe "JavaScript Minifier", even.

sahildua2305:
@MariagraziaAlastra Thanks :innocent: 

I have changed the tab name to "Minifier" so that even CSS minifier can come under the same name.

Also, @moollaza the CSS minifier doesn't need to be different. I can change this IA's name to `Minifier` and hence it can be used to minify CSS or JS depending upon the query. There can be an option to choose between CSS and JS for the user. What do you think?

sahildua2305:
@GuiltyDolphin Please have a look. Something seems to be wrong with tests for `ChineseToPinyin`.

lib/DDG/Goodie/JsMinify.pm
```diff
@@ -0,0 +1,35 @@
...
+handle query_lc => sub {
+
+    return unless $_;
+
+    return '',
+        structured_answer => {
+
+            id => "js_minify",
+
+            name => "Minifier",
```
MariagraziaAlastra: @sahildua2305 you can remove this, since we decided to stick with the "Answer" tab name for now

share/goodie/js_minify/js_minify.js
```diff
@@ -0,0 +1,52 @@
...
+                    // Operate using the prettydiff function provided by the library
+                    var output = prettydiff(args);
+
+                    // hide output textarea by default
+                    $output.css('display', 'inline');
+
+                    // Add the output to output textarea field
+                    $output.val(output);
+                });
+            })
```
MariagraziaAlastra: semicolon

sahildua2305:
@MariagraziaAlastra please deploy this on Beta.

share/goodie/js_minify/js_minify.js
```diff
@@ -0,0 +1,52 @@
...
+                $minifyButton = $dom.find('.js_minify__action'),
+                $input = $dom.find('#js_minify__input'),
+                $output = $dom.find('#js_minify__output');
+
+            $main.toggleClass('c-base');
+
+            // hide output textarea by default
+            $output.css('display', 'none');
+
+            DDG.require(‘prettydiff’, function() {
```
MariagraziaAlastra: @sahildua2305 Could you use regular quote marks ('' or "" instead of ‘‘)? This is making the minifying process break when installing on beta:
<pre>
Parse error at js_minify.js:30,24
Unexpected character '‘'
Error 
</pre>

share/goodie/js_minify/js_minify.js
```diff
@@ -0,0 +1,52 @@
+DDH.js_minify = DDH.js_minify || {};
+
+"use strict";
```
MariagraziaAlastra: Could you move this inside the function? Like in [here](https://github.com/duckduckgo/zeroclickinfo-goodies/blob/10ed5c13f4a9bcae039cb40d4d1fcd1882a08e2c/share/goodie/game2048/game2048.js#L4)

MariagraziaAlastra:
@sahildua2305 It's deployed to [beta](https://beta.duckduckgo.com/?q=js+minify&ia=answer)!
Now, the very last thing to do here is make sure the library loads on textarea change, and then we're good to go 😄

sahildua2305:
@MariagraziaAlastra Awesome! :smiley: 
How to handle the case when the user clicks on the `minify` button before the library has loaded completely? Should I disable the button till the time library is loaded completely?

MariagraziaAlastra:
Yeah, disabling the button until the library is loaded  sounds like a good idea. Thanks @sahildua2305!

sahildua2305:
@MariagraziaAlastra please deploy this on beta so that we can test it. I have made the library load on change in textarea contents. I have also disabled the button till the time library is loaded.

share/goodie/js_minify/js_minify.js
```diff
@@ -0,0 +1,82 @@
...
+
+            // Hide output textarea by default
+            $output.css('display', 'none');
+
+            // Disable 'Minify' button by default and change cursor to default
+            $minifyButton.prop('disabled', true);
+            $minifyButton.css('cursor', 'default');
+
+            // Add event handler for change in input of textarea
+            $input.on('input', function() {
```
MariagraziaAlastra: @sahildua2305 maybe try on keyup instead?

share/goodie/js_minify/js_minify.js
```diff
@@ -0,0 +1,82 @@
...
+                if (!libLoaded) {
+                    // Set the flag to make sure the library isn't loaded
+                    // again and again
+                    libLoaded = true;
```
MariagraziaAlastra: :+1: good one

MariagraziaAlastra:
Hey @sahildua2305 it's deployed again! It works, but for some reason the minified code shows underneath the original code, instead of showing on the right:
screenshot-beta duckduckgo com 2016-07-18 17-21-59

I'm using Firefox 47, 1920 × 1080 screen resolution

share/goodie/js_minify/js_minify.js
sahildua2305: I am afraid `keyup` won't be triggered if user pastes the code using context menu. `input` will be triggered whenever the input changes, in any way. @MariagraziaAlastra

sahildua2305:
@MariagraziaAlastra this is weird! It works for me. at a lower resolution. Can you please check the dimensions of `div` with class `zci__main` inside `#zci-js_minify`?

share/goodie/js_minify/js_minify.js
MariagraziaAlastra: Ok cool, I was worried about IE9, but turns out this has basic support there, except for deleting, pressing backspace and cutting, and I think we don't care about either of those :smile:

MariagraziaAlastra:
@sahildua2305 Thanks, deployed again! Looks good on my notebook, let me just test it on my desktop in a few hours, before the IAs release.
Hopefully it should be ready!

sahildua2305:
@MariagraziaAlastra let me know how it goes 😄

MariagraziaAlastra:
@sahildua2305 It's ready! Merging now...
Thanks again for the great work here! :smile:

daxtheduck:
@sahildua2305 this is now merged! It should be live on DuckDuckGo within one week (or during the next release) and we'll be sure to notify you when it happens.

share/goodie/js_minify/content.handlebars
```diff
@@ -0,0 +1,8 @@
+<h5>Minify JavaScript Code on the go!</h5>
```
moollaza: This should be a `title` property from `data` -- don't need the `content` template to render this 👍

moollaza: Let's change the text to "JavaScript Minifier"

share/goodie/js_minify/js_minify.js
```diff
@@ -0,0 +1,82 @@
...
+            var $dom = $('#zci-js_minify'),
+                $main = $dom.find('.zci__main'),
+                $minifyButton = $dom.find('.js_minify__action'),
+                $input = $dom.find('#js_minify__input'),
+                $output = $dom.find('#js_minify__output');
+
+            $main.toggleClass('c-base');
+
+            // Hide output textarea by default
+            $output.css('display', 'none');
```
moollaza: We should hide this by adding the `is-hidden` class to the template instead no need to create it then hide it in the DOM. It should be created as hidden.

share/goodie/js_minify/js_minify.js
```diff
@@ -0,0 +1,82 @@
...
+            // Set the flag to true so it doesn't get run again
+            shown = true;
+
+            var $dom = $('#zci-js_minify'),
+                $main = $dom.find('.zci__main'),
+                $minifyButton = $dom.find('.js_minify__action'),
+                $input = $dom.find('#js_minify__input'),
+                $output = $dom.find('#js_minify__output');
+
+            $main.toggleClass('c-base');
```
moollaza: why are we doing this?

share/goodie/js_minify/js_minify.js
```diff
@@ -0,0 +1,82 @@
...
+
+            // Hide output textarea by default
+            $output.css('display', 'none');
+
+            // Disable 'Minify' button by default and change cursor to default
+            $minifyButton.prop('disabled', true);
```
moollaza: should be disabled in template, and enabled in JS when necessary

share/goodie/js_minify/js_minify.js
```diff
@@ -0,0 +1,82 @@
...
+            $minifyButton.css('cursor', 'default');
```
moollaza: Use CSS for this too

share/goodie/js_minify/js_minify.js
```diff
@@ -0,0 +1,82 @@
...
+                    // again and again
+                    libLoaded = true;
+
+                    // Change text of button to show the loading action
+                    // to make sure users aren't confused to see
+                    // the disabled button
+                    $minifyButton.text('Loading..');
+
+                    // load the library
+                    DDG.require('prettydiff', function() {
```
moollaza: We should use DDG.async_build to load the library, and only once loaded, display the Instant Answer.

This prevents the need to wait, and also prevents us from having to deal with inputs and button presses before the lib is loaded.

Look at the WhereAmI Goodie to see how to use DDG.build_async.

The differenence is that you use a callback function to return your DDH object, instead of `return`ing an object inside your `build` function

share/goodie/js_minify/js_minify.js
```diff
@@ -0,0 +1,82 @@
...
+                    mode: "minify",
+                    lang: "javascript",
+                    source: $input.val()
+                };
+
+                // Operate using the prettydiff function provided by the library
+                var output = prettydiff(args);
+
+                // hide output textarea by default
+                $output.css('display', 'inline');
```
moollaza: We should remove the `is-hidden` class instead

lib/DDG/Goodie/JsMinify.pm
```diff
@@ -0,0 +1,33 @@
+package DDG::Goodie::JsMinify;
+# ABSTRACT: An Interactive JavaScript Minifier Tool
+
+use DDG::Goodie;
+use strict;
+
+zci answer_type => 'jsminify';
+zci is_cached => 1;
+
+triggers any => 'js minify', 'js minifier', 'javascript minify', 'javascript minifier', 'minify js', 'minify javascript', 'minifier js', 'minifier javascript';
```
moollaza: this needs to be `startend` -- any is too broad unless we're evaluating the query in `handle` for leading and trailing words

lib/DDG/Goodie/JsMinify.pm
```diff
@@ -0,0 +1,33 @@
...
+
+handle query_lc => sub {
```
moollaza: if we want the queries that trigger this to only be our triggers, then we need to `handle remainder` and check that it's empty

share/goodie/js_minify/js_minify.css
```diff
@@ -0,0 +1,26 @@
+.zci--js_minify .js_minify__action_box {
+    margin-top: 3.25em;
+}
+
+.zci--js_minify .js_minify__action {
+    background: #5DA4DC;
+    color: #fff;
```
moollaza: We shouldn't hardcode colors here, otherwise this will break for DDG themes. Use the `tx-clr--` and `bg-clr--` classes from https://ddg.gg/styleguide to handle the text and background colours. That way they'll adapt to DDG themes

share/goodie/js_minify/js_minify.css
```diff
@@ -0,0 +1,26 @@
...
+    text-decoration: none;
+    font-size: 1.25em;
```
moollaza: Use our `tx-##` class to handle font size. We have a `tx--12` I think or `t-xs`

share/goodie/js_minify/js_minify.css
```diff
@@ -0,0 +1,26 @@
+.zci--js_minify .js_minify__action_box {
+    margin-top: 3.25em;
+}
+
+.zci--js_minify .js_minify__action {
```
moollaza: Actually, all of this can be handled by out `btn` class

share/goodie/js_minify/js_minify.css
```diff
@@ -0,0 +1,26 @@
...
+    cursor: pointer;
+}
+
+.zci--js_minify .js_minify__action:active {
+    background: #4793D0;
+}
+
+.zci--js_minify #js_minify__input,
+.zci--js_minify #js_minify__output {
+    width: 45%;
```
moollaza: We will probably need a CSS rule here to handle mobile UI.

I think `width: 100%` is best so they're always stacked.

Also, we should set `font-face: 'monospace'`

share/goodie/js_minify/content.handlebars
```diff
@@ -0,0 +1,8 @@
...
+<div class="js_minify__io_box">
+    <textarea name="js_minify__input" id="js_minify__input" rows="20" cols="60"></textarea>
+    <textarea name="js_minify__output" id="js_minify__output" rows="20" cols="60" readonly="readonly" onclick="this.focus();this.select()"></textarea><br>
+</div>
+<div class="js_minify__action_box">
+    <button class="js_minify__action">Minify</button>
```
moollaza: Our `btn` class can handle the styling here

share/goodie/js_minify/js_minify.js
sahildua2305: Honestly, I don't remember why I removed this class. It was creating some CSS related problems.

I will test whether it's required or not, once I have made the suggested CSS changes.

share/goodie/js_minify/js_minify.js
sahildua2305: @moollaza I should skip this, right?

PR #3424, opened 2016-07-21
CSS Animations Goodie

### LINKED ISSUE #3389: Create CSS Animations Goodie
### Background

CSS Animations are very popular and we can very easily display a tile view of CSS Animation demos, that show what the effect of the animation is. 

Ideally we can link to a live demo of the CSS where the user can copy and modify the CSS we've demonstrated.

For example, CSS-Tricks has a great post with embedded samples: https://css-tricks.com/css-animation-tricks/
### Objective

Build a Goodie that displays a tile view where each tile demonstrates a different CSS animation technique. Each tile should be a link to a CodePen where the user can see and play with the code.
#### Example Queries
- "css animations"
- "css animations demos"
- "css animations examples"
#### Example Animation Demos
- https://robots.thoughtbot.com/css-animation-for-beginners
- http://www.webdesignerdepot.com/2014/05/8-simple-css3-transitions-that-will-wow-your-users/
- http://www.the-art-of-web.com/css/css-animation/
  Reply
### What to do:
1. Comment here if you'd like to handle this task. 
2. Review any Related Discussions, Issues, and Pull Requests (below).
3. [Setup your Dev Environment](http://docs.duckduckhack.com/welcome/setup-dev-environment.html) (first-timers only)
4. Refer to the [Fathead documentation](http://docs.duckduckhack.com/resources/fathead-overview.html) and [other Fatheads](https://duck.co/ia?repo=fathead) as examples
5. [Submit a Pull Request](http://docs.duckduckhack.com/submitting/pull-request.html) with your completed changes
### Who to contact for help:

Please contact @moollaza  or see [other contact info](http://docs.duckduckhack.com/resources/get-in-touch.html).

---

**Related Discussions, Issues, and Pull Requests:** 

Forum Topic: https://forum.duckduckhack.com/t/create-ia-css-animations-goodie/262/2

ManrajGrover:

your Pull Request.
- [x] New Instant Answer
  - [x] Other: **`New {CSS Animations} Instant Answer`**
- [ ] Improvement
- [ ] Non–Instant Answer
###### Description of changes

This pull request adds CSS Animations Goodie
Fixes #3389

@moollaza @zekiel 

---

Instant Answer Page: https://duck.co/ia/view/css_animations

[Maintainer](http://docs.duckduckhack.com/maintaining/guidelines.html): @mention
None


daxtheduck:
Metadata placeholder post
            
---
            
**How to find your IA page**
- New Instant Answer
              
- Existing Instant Answer

edgesince84:
thank for taking this one up, @ManrajGrover!

ManrajGrover:
@moollaza Need some input here. I'm currently storing data in yml file. Is it ok or there are better ways to do this?

edgesince84:
hey @ManrajGrover - Zaahir is AFK for a bit, I'll try and loop in someone else to help you sooner.

ManrajGrover:
@edgesince84 That would help. Thank you! :smile:

tagawa:
Just adding a screenshot here for reference:

css-animations-screenshot

@ManrajGrover I'm wondering if this could be better served in a different way, with an API or something, or with CSS information is more general or wider reaching. If it's OK, I'd like to hear what @moollaza has to say.

ManrajGrover:
@tagawa I would love to make changes accordingly. :smile:

lib/DDG/Goodie/CSSAnimations.pm
```diff
@@ -0,0 +1,40 @@
+package DDG::Goodie::CSSAnimations;
```
moollaza: @ManrajGrover I would suggest changing this to `CssAnimations` so the `id` will be `css_animations`

lib/DDG/Goodie/CSSAnimations.pm
```diff
@@ -0,0 +1,40 @@
...
+triggers start => 'css animations', 'css animations example', 'css animation', 'css animation examples', 
+                  'css animation demo', 'css animations demos', 'css animations demos', 'css animation demos';
+
+my $animations = LoadFile(share('content.yml'));
+
+handle remainder => sub {
+    print Dumper($animations);
+    return 'CSS Animations',
+        structured_answer => {
+            id => 'cssanimations',
```
moollaza: Once the package name is corrected, the `name` and `id` can be removed

moollaza:
@ManrajGrover I think I may not have been clear enough when describing my idea here. I would like to see each tile as an actual demo of the CSS animation. So instead of words, they would see a live animation in each tile, and clicking the animation would perhaps show a detail view with some info, or simply send the user to a codepen where they can see what code was used to create that animation.

Perhaps the animations should only fire on hover, but the idea is to visually convey the animation instead of describing them.

ManrajGrover:
@moollaza We can add the demo in [detail](http://docs.duckduckhack.com/frontend-reference/templates-overview.html) part of the template. So, if user clicks on a tile, a detail box will open and CodePen embedded demo will be shown with full description and link to CodePen demo. I guess this should work. Firing animations on hover over the tiles won't look good as tiles are small, and some of the demos may not be responsive enough or require good width. 

I feel current `yml` file is serving good purpose. Let me know if changes are required in that too. :smile:

moollaza:
> We can add the demo in detail part of the template. So, if user clicks on a tile, a detail box will open...

@ManrajGrover I'd like to put the demos right in the tiles to make this really attractive visually and to avoid hiding the awesome animations behind a detail pane. That way no clicks are needed to see the awesome demos.

> Firing animations on hover over the tiles won't look good as tiles are small, and some of the demos may not be responsive enough or require good width.

We have full control over what demos we show and how tall/wide the tiles are here, so if necessary we can increase the tile widths and heights (evenly) as needed (within reason).

Our Sports IA's have very large tile widths, and out poster-style IA's have tall tiles, I'm comfortable with us creating larger-than-average tiles to accommodate.

>... and CodePen embedded demo will be shown with full description and link to CodePen demo.

We can't embed CodePen demos. I believe they require an iFrame? This would be considered a privacy breach so we'll have to create the CSS/HTML content for each tile, meaning this won't be a standard Instant Answer, it'll be a little more involved than most ;)

ManrajGrover:
> We have full control over what demos we show and how tall/wide the tiles are here, so if necessary we can increase the tile widths and heights (evenly) as needed (within reason).

@moollaza Yes! That makes it more clear. :smile: 

> We can't embed CodePen demos. I believe they require an iFrame? This would be considered a privacy breach so we'll have to create the CSS/HTML content for each tile, meaning this won't be a standard Instant Answer, it'll be a little more involved than most ;)

Yeps, they require iFrames. This looks interesting. So basically, we will convert each CodePen link to a partial HTML file and import them in handlebars. Please correct me if I'm wrong.

moollaza:
> So basically, we will convert each CodePen link to a partial HTML file and import them in handlebars.

I think we'll have to build out the full demos within this repo, then duplicate that code on CodePen so we can link to an environment where the developer can interact with and modify the code.

Though I believe CodePen exposes the individual HTML/CSS/JS assets for a given Pen, accessing them on DDG would be difficult because I'm not sure if we'll be able to get the plaintext HTML/JS/CSS assets through our existing proxy. This would also mean switching to a Spice, and I suspect we'd need to build an iFrame within the tile to display the HTML/CSS/JS we pull down from CodePen.

Would you like to investigate this a little more?

Alternatively, you can get started with this as a Goodie and we can determine if the Spice approach will work later. If it does, then we could possible also build a general CodePen Spice

ManrajGrover:
@moollaza Something that I feel can work is this:

1. We store CodePen link in YML file. This way we can access HTML and CSS from the demos, for example `http://codepen.io/manrajsingh/pen/wWaQed.css`. Here, there may be security issue because we will be importing code from external source, so we would have to duplicate/fork those CodePens on official DuckDuckGo account. Do let me know how to credit the developer.

2. Then, we send that link to Front end to generate iFrames with those content. I don't know if we can avoid it but injecting CSS directly may effect elements of same name.

3. On hover, we can show the animations accordingly.

Please share your thoughts.

moollaza:
@ManrajGrover I think I found an even easier way!

We can provide a button to CodePen that POST's the necessary JS/HTML/CSS so that when the button is clicked the data is passed along.

This means we don't even need to deal with iFrames or APIs. We could implement this as a Goodie and leave the code in our repo. 

I believe this is the way MDN works as well 👍

ManrajGrover:
@moollaza This is awesome! :smile: But I guess we would still need iFrames to show demos on hover. Please correct me if I'm missing anything.

moollaza:
> But I guess we would still need iFrames to show demos on hover. Please correct me if I'm missing anything.

I don't see why?

ManrajGrover:
@moollaza Because we would need to show animations on hover, as discussed above. I get the usage of buttons for generating CodePen. Am I missing anything?

ManrajGrover:
@moollaza As per our discussion, I tried to embed HTML and CSS directly into tiles. It didn't work out as expected. I tried using iFrames, just to see what happens. They didn't work. Now, I have learned about 'scoped css' which basically allow having css locally for a particular section. Unfortunately, its browser support was [removed](http://caniuse.com/#feat=style-scoped) but there are some polyfills available. Let me know if it works out, otherwise we would need to recreate/make changes in every demo we show, like we discussed before.

ManrajGrover:
Update:

@moollaza I've been busy lately and will be pushing changes as per our discussion on slack soon. :smile:

moollaza:
@ManrajGrover Ping! I'm just checking in to see if you're still interested in finishing this one up?

ManrajGrover:
@moollaza Yeps. I've been a lot busy lately due to shifting of work place. I'll be picking this one up hopefully this weekend and pushing the changes.

ManrajGrover:
@moollaza I guess this can be reviewed now. I have incorporated all of the discussions we had into this. 😄

moollaza:
@ManrajGrover this is now live on Beta but it's not working: https://beta.duckduckgo.com/?q=CSS+Animations&ia=answer

Care to take a look around and see?

ManrajGrover:
@moollaza I see that old code has been deployed to beta. On checking the JS, I see [this](http://pastebin.com/cq83bXnE) which is not current state of PR. Please check.

moollaza:
@ManrajGrover after a redeploy, it looks to be working! 

https://beta.duckduckgo.com/?q=CSS+Animations&ia=answer

tagawa:
I like this a lot! Is it possible to use some other text in the first animation apart from "Error 404"? It might confuse some users.

tagawa:
I just updated the beta server with the latest code change: https://beta.duckduckgo.com/?q=CSS+Animations&ia=answer

ManrajGrover:
@tagawa I would love to make the change as per requirements. One thing I'm not understanding is why build failed with test file not related to this Goodie. Can we retry the build?

tagawa:
Thanks @ManrajGrover. The failing builds are because of an issue with NameDays. There's now a fix in the pipeline ( https://github.com/duckduckgo/zeroclickinfo-goodies/pull/3785 ) so it should be fine again once that's merged. In the meantime, it's OK to ignore the build failure messages. Sorry for the confusion.

moollaza:
@ManrajGrover sorry about the delay, I've fixed the NameDays IA. Would you be able to update and merge master into your branch?

ManrajGrover:
@tagawa How about having "CLOSED" as the text instead of "ERROR 404"?
@moollaza No issues. 😄 Build passed now.

tagawa:
@ManrajGrover Yes, I think that's better. It's supposed to represent a neon sign, right? I like the way you have two lines with two different colours, so a similar format could be:

    OPEN
    
    COME IN!

I'll leave it to you - just avoiding the word "error" is safe I think.

ManrajGrover:
@tagawa Made the change 😄 This can be reviewed now.

tagawa:
@ManrajGrover Yay, looks good - thank you!
https://beta.duckduckgo.com/?q=CSS%20Animations

I think it's ready to merge. @hchienjo - any final comments?

ManrajGrover:
@tagawa Looks like old code has been deployed to beta. Please look into it.

hchienjo:
@tagawa I don't see any results [from](https://beta.duckduckgo.com/?q=CSS+Animations&ia=answer). Could it be caused by what @ManrajGrover is pointing out?

tagawa:
@ManrajGrover @hchienjo Strange, it was working at first but then disappeared. I've just redeployed it and it seems OK now: https://beta.duckduckgo.com/?q=CSS+Animations&ia=answer

hchienjo:
This looks good to me. Thanks @ManrajGrover!

moollaza:
@ManrajGrover Thanks! This LGTM :+1:

Excellent work 💯 🚀

moollaza:
@ManrajGrover whoops, just noticed that the triggering is too loose right now -- any query that starts with "css animations" triggers this:

https://beta.duckduckgo.com/?q=CSS+Animations+yolo&ia=answer

You need to check that the remainder is empty 👍

share/goodie/css_animations/demo_3/style.css
```diff
@@ -0,0 +1,19 @@
+.tile--css_animations__demo4 div {
+  width:115px; height:115px;
```
moollaza: nit: height should be on a new line

share/goodie/css_animations/demo_4/style.css
```diff
@@ -0,0 +1,14 @@
+.tile--css_animations__demo5 div { width: 50px; height: 50px; transform-origin: 100px 70px; position:relative; animation:single_rotate 3s linear infinite; }
```
moollaza: let's make sure this is properly formatted/indented. each css rule should be on a new line

t/CssAnimations.t
```diff
@@ -0,0 +1,40 @@
...
+            }
+        };
+}
+
+sub build_test { test_zci(build_structured_answer()) }
+
+ddg_goodie_test(
+    [qw( DDG::Goodie::CssAnimations )],
+    'css animations' => build_test(),
+    'help css animations' => undef,
```
moollaza: let's add something like "css animation generator" -- that shouldn't trigger this, but currently does

lib/DDG/Goodie/CssAnimations.pm
```diff
@@ -0,0 +1,59 @@
...
+
+zci is_cached => 1;
+
+triggers start => 'css animations', 'css animations example', 'css animation', 'css animation examples', 
+                  'css animation demo', 'css animations demos', 'css animations demos', 'css animation demos';
+
+my $animations = LoadFile(share('data.yml'));
+
+handle remainder => sub {
+    my $demo_count = keys $animations;
```
moollaza: `return unless $_ eq '';`

lib/DDG/Goodie/CssAnimations.pm
```diff
@@ -0,0 +1,59 @@
...
+    
+    for (my $i=0; $i < $demo_count; $i++) {
+        my $demo = "demo_$i";
+        my $title = $animations->{$demo}->{'title'};
+        my $html = share("$demo/demo.html")->slurp if -e share("$demo/demo.html");
+        my $css = share("$demo/style.css")->slurp if -e share("$demo/style.css");
+        my $links = share("$demo/links.html")->slurp if -e share("$demo/links.html");
+        my %value = ('title' => $title, 'html' => $html, 'css' => $css, 'head' => $links);
+        
+        $animations->{$demo}->{'html'} = $html;
```
moollaza: we need to handle the case when any of these are undef, they're not guaranteed to be defined because they're conditionally created;

e.g. `$animations->{$demo}->{'html'} = $html || ''`

lib/DDG/Goodie/CssAnimations.pm
```diff
@@ -0,0 +1,59 @@
...
+                  'css animation demo', 'css animations demos', 'css animations demos', 'css animation demos';
+
+my $animations = LoadFile(share('data.yml'));
+
+handle remainder => sub {
+    my $demo_count = keys $animations;
+    
+    my @result = ();
+    
+    for (my $i=0; $i < $demo_count; $i++) {
```
moollaza: This can all be built outside of the `handle` -- no need to run this each time the IA triggers. We can store the demo results in memory 👍

share/goodie/css_animations/demo_0/links.html
```diff
@@ -0,0 +1 @@
+<link href='https://fonts.googleapis.com/css?family=Monoton' rel='stylesheet' type='text/css'>
```
moollaza: Woah, let's skip the custom font -- **this is a privacy leak**. The client will make a direct request to Google's server, revealing their IP

ManrajGrover: @moollaza Should I make it local?

lib/DDG/Goodie/CssAnimations.pm
ManrajGrover: I'm not sure about this. Does this mean having a YML file with demos already built?

moollaza: @ManrajGrover no, I just mean we can move this function above, outside of `handle`.

If it's inside the `handle` function, it'll get called every time the IA triggers (when the cache expires). Instead, we can build our data once and leave it in memory. 

The `handle` function can then setup the data and return the result as it does

ManrajGrover:
@moollaza This can be reviewed again.

lib/DDG/Goodie/CssAnimations.pm
```diff
@@ -3,7 +3,6 @@ package DDG::Goodie::CssAnimations;
 
 use DDG::Goodie;
 use YAML::XS 'LoadFile';
-use Data::Dumper;
```
moollaza: fwiw this is already loaded into all IAs via the DuckDuckGo lib 👍  -- you can just use `p()`

moollaza:
@ManrajGrover thanks for the updates! 👍 

Deploying to Beta...

moollaza:
@ManrajGrover Thanks! This LGTM :+1:

Excellent work 👍 💯 🚀

daxtheduck:
@ManrajGrover this is now merged! It should be live on DuckDuckGo within one week (or during the next release) and we'll be sure to notify you when it happens.

PR #3635, opened 2016-09-27
New SASS to CSS IA

akanshgulati:

your Pull Request.

- [x] New Instant Answer
    - [ ] Other: **`New SASS to CSS Instant Answer`**
###### Description of changes
This is the new IA for converting SASS to CSS

What changes does this PR introduce? 

###### People to notify

@moollaza  @hchienjo 

---
Instant Answer Page: https://duck.co/ia/view/sass_to_css
Maintainer: @moollaza


daxtheduck:
Metadata placeholder post
            
---
            
**How to find your IA page**
- New Instant Answer
              
- Existing Instant Answer

moollaza:
@akanshgulati which of the files need to be loaded for all this to work? I assume we can get rid of the un-minified files?

I will have to add them internally so that they can be loaded via `DDG.require( 'libraryName', function { ... })` will load them.

akanshgulati:
https://cdnjs.cloudflare.com/ajax/libs/sass.js/0.9.12/sass.min.js
https://cdnjs.cloudflare.com/ajax/libs/sass.js/0.9.12/sass.worker.min.js

These are the two files which need to be added in same folder.

t/SassToCss.t
```diff
@@ -0,0 +1,56 @@
...
+sub build_structured_answer {
+    my @test_params = @_;
+
+    return "",
+        structured_answer => {
+
+            id => "sass_to_css",
+
+            data => {
+                goodie_version => 999,
```
moollaza: @akanshgulati please remove this so the tests are passing again

share/goodie/sass_to_css/sass.min.js
```diff
@@ -0,0 +1,21 @@
+/*! sass.js - v0.9.12 (e1a0112) - built 2016-08-14
```
moollaza: Please remove this and the other libraries from the PR as they're no longer needed

lib/DDG/Goodie/SassToCss.pm
```diff
@@ -0,0 +1,45 @@
...
+use warnings;
+
+zci answer_type => 'sass_to_css';
+
+zci is_cached => 1;
+
+
+
+# Triggers - http://docs.duckduckhack.com/walkthroughs/calculation.html#triggers
+triggers any => share('triggers.txt')->slurp;
```
moollaza: I think we should use `startend` here -- do you have any example queries where the trigger phrase is in the middle of a larger query?

lib/DDG/Goodie/SassToCss.pm
```diff
@@ -0,0 +1,45 @@
...
+
+handle remainder => sub {
+
+    # Return unless the remainder is empty or contains online or tool
+    return unless ( $_ =~ /(^$|online|tool)/i );
+
+    return '',
+        structured_answer => {
+
+            id => "sass_to_css",
```
moollaza: Goodies don't need to specify the `ID` -- it comes from the IA Page 👍

share/goodie/sass_to_css/sass_to_css.js
```diff
@@ -28,7 +27,6 @@ DDH.sass_to_css.build = function(ops) {
             // Load library when the IA is shown for the first time
 
             DDG.require('sass.js', function(){
-                sass = new window.Sass();
```
moollaza: @akanshgulati you still need this, or you'll need to move the `$validateButton.click` function definition inside the `DDG.require()` call.

the `Sass` variable is only guaranteed to be defined inside there, as that callback function executes after sass.js has loaded.

On a slow connection for example, `DDG.require` could take a while to load sass.js, but the click function will still be defined and `Sass` isn't guaranteed to exist yet.

akanshgulati: @moollaza you are right on this, I even thought on this and making changes to add click event only when sass.js is loaded.
Moreover, loading sass.js initially is also not efficient, I am changing the library to load when input has finally some content.

moollaza:
@akanshgulati this is now live on Beta 👍 

I noticed that it only seems to accept SCSS and not SASS input? We should ensure it supports both syntaxes.

I borrowed the snippet of code from the [SASS Lang website](http://sass-lang.com/guide)

sass_to_css_at_duckduckgo

daxtheduck:
@akanshgulati this is now merged! It should be live on DuckDuckGo within one week (or during the next release) and we'll be sure to notify you when it happens.

PR #3714, opened 2016-10-23
JSONValidator: Add option to choose indentation in formatted output

### LINKED ISSUE #3713: JSONValidator Suggestion: Provide options to select preferred indentation in formatted json
## Type of Issue

- [ ] Bug (incorrect behavior)
  - [ ] Triggering
  - [ ] Relevancy
  - [ ] Internal (issue with code)
- [X] Feature Request
  - [ ] Triggers (new triggers)
  - [X] Design Change
## Description

There should be an option so that users can choose from 2 or 4 spaces indentation options that they want to have in their final formatted json.
**Hypothesis** - Ideally, our IAs should be as flexible as possible and shouldn't be posing any limitations on the result for our users.

---

Instant Answer Page: https://duck.co/ia/view/json_validator

sahildua2305:
## Type of Change

- [ ] New Instant Answer
  - [ ] Cheat Sheet
- [X] Improvement
  - [ ] Bug fix
  - [X] Enhancement
- [ ] Non–Instant Answer
  - [ ] Other (Role, Template, Test, Documentation, etc.)

Fixes #3713 
## People to notify

@MariagraziaAlastra @moollaza 

---

Forum Topic:

IA Page: https://duck.co/ia/view/json_validator

### DIFF
--- a/share/goodie/jsonvalidator/content.handlebars
+++ b/share/goodie/jsonvalidator/content.handlebars
@@ -1,5 +1,9 @@
 <textarea name="json_validator--input" class="json_validator--input whole tx--12 frm__text" rows="{{rows}}" spellcheck="false" cols="60" placeholder="Paste code here..."></textarea>
 
+<div class="json_validator--indentation-options">
+	<input type="radio" name="indentation" value="2"> 2 spaces
+	<input type="radio" name="indentation" value="4" checked> 4 spaces
+</div>
 <button class="btn btn--skeleton whole--screen-s json_validator--validate_button" disabled>Loading..</button>
 <button class="btn btn--secondary whole--screen-s json_validator--clear_button">Clear</button>
 
--- a/share/goodie/jsonvalidator/content.handlebars
+++ b/share/goodie/jsonvalidator/content.handlebars
@@ -1,6 +1,6 @@
 <textarea name="json_validator--input" class="json_validator--input whole tx--12 frm__text" rows="{{rows}}" spellcheck="false" cols="60" placeholder="Paste code here..."></textarea>
 
-<div class="json_validator--indentation-options">
+<div class="json_validator--indentation_options">
 	<input type="radio" name="indentation" value="2"> 2 spaces
 	<input type="radio" name="indentation" value="4" checked> 4 spaces
 </div>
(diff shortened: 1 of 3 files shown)


daxtheduck:
Metadata placeholder post
            
---
            
**How to find your IA page**
- New Instant Answer
              
- Existing Instant Answer

share/goodie/jsonvalidator/content.handlebars
```diff
@@ -1,5 +1,9 @@
 <textarea name="json_validator--input" class="json_validator--input whole tx--12 frm__text" rows="{{rows}}" cols="60" placeholder="Paste code here..."></textarea>
 
+<div class="json_validator--indentation_options">
```
moollaza: @sahildua2305 We should probably use a `<select>` instead of radios -- less visual clutter.

Also we have form classes that should be used for styling. Checkout https://duckduckgo.com/styleguide

sahildua2305: @moollaza I used `radio` buttons so that user is able to see the both options in first glance and there's no need to put any label to describe what this option is for. Should I use still `select`?

sahildua2305: Also, it's not clear from the styleguide which classes to use in this case, because the example there is too narrow. :/

moollaza: @sahildua2305 yes, let's use a `select` I think it looks much nicer. Also allows for more options without using more space.

```
<div class="json_validator--indentation_options">
  <form class="frm frm--vrt">
<div class="frm__select">
<select name="tabsize" id="tabsize">
                    <option value="1">Indent with a tab character</option>
                    <option value="2">Indent with 2 spaces</option>
                    <option value="3">Indent with 3 spaces</option>
                    <option value="4">Indent with 4 spaces</option>
                    <option value="8">Indent with 8 spaces</option>
                </select>
  </div>
  </form>
</div>
```

Produces:
json_validator_at_duckduckgo

I think we'll want the select to be on the same like as the other buttons though, or perhaps above the `textarea`.

If we have more options, then perhaps having them on their own line is best.

FWIW I often use the DDG Settings page as a good example of our form classes/markup 👍

sahildua2305: Thanks for detailed reply @moollaza 😻 

> FWIW I often use the DDG Settings page as a good example of our form classes/markup 👍

Thanks for sharing the secret! 👍
Don't you think we should rather add these as examples in the [styleguide](https://duckduckgo.com/styleguide) itself?

sahildua2305:
Good to go, @moollaza

sahildua2305:
@moollaza can you please review again?

moollaza:
@sahildua2305 this looks amazing! Nicely done 💯  👍 

One thought: changing the indentation should trigger a re-build, so they don't have to click the button again

PR #3805, opened 2016-12-09
New Molar Mass Calculator

Iambecomeroot:
@acampb311 has given me permission to continue work on their molar mass calculator IA. Since that pr was closed, I thought I would open a new one.

Calculate the molar mass of compounds. For example: "Molar mass of h2o" yields 18.025

## People to notify

@acampb311

---

Instant Answer Page: https://duck.co/ia/view/molarmass


daxtheduck:
@Iambecomeroot Thanks for the pull request! Unfortunately, we're only accepting contributions and improvements to a limited portion of the search space.
            
The staff and community are currently focused on making DuckDuckGo the [best search engine for programmers](https://forum.duckduckhack.com/t/duckduckhack-programming-mission-overview/53). To ensure we're making as much progress as possible, **we are only accepting Instant Answers and updates related to the programming mission**.
    
If this is not related to the Programming Mission, it will likely be closed but you can:
    
1. Help us prevent further confusion by telling us where you found out about DuckDuckHack or where we could've alerted you earlier.
2. Keep this code handy! We'll be expanding into other search topics in the future!
3. Join us at our new home: https://forum.duckduckhack.com/

If this is related to the Programming Mission, it will be reviewed by staff and the community as soon as possible!

daxtheduck:
Metadata placeholder post
            
---
            
**How to find your IA page**
- New Instant Answer
              
- Existing Instant Answer

Iambecomeroot:
I got it to work. Now all I have to do is figure out html output, handle some edge cases, and write tests. Anyone know how to do html output?

Iambecomeroot:
What needs to happen before we can get this merged?

lib/DDG/Goodie/MolarMass.pm
```diff
@@ -0,0 +1,225 @@
...
+use List::Util qw(any);
+use YAML::XS 'LoadFile';
+zci answer_type => 'molar_mass';
+zci is_cached => 1;
+
+# Get elements from yml file
+my @elements = @{ LoadFile(share('elements.yml')) };
+my @symbols = map { lc @$_[3] } @elements;
+
+triggers query_lc => qr/(?:what is the |)(molar|atomic) (mass|weight) (?:of|)/;
```
mintsoft: We generally tend to prefer word triggers where possible. Judging by the regex `triggers any => qw(molar atomic mass weight);` should trigger in the same conditions and the guard regex in handle should ensure that it doesn't trigger too much

lib/DDG/Goodie/MolarMass.pm
```diff
@@ -0,0 +1,225 @@
...
+
+# Handle statement
+handle query_lc => sub {
+    my $compound = $_;
+    
+    # Parse input
+    $compound =~ s/((?:what is the |)(molar|atomic) (mass|weight) (?:of|)| )//g;
```
mintsoft: Rather than replacing the bits we won't want, I find it easier to match out the bits we do want. Something like this:
```
my $input =  $compound =~ /^(?:what is the |)(molar|atomic) (mass|weight) (?:of)? ([a-z]+)$/;
```
At that point we have $input which contains the bits of the query we want. Another option is to use named regex matches (`(?<name>molar|atomic)`; then `$+{'name'}` contains `molar` or `atomic`)

lib/DDG/Goodie/MolarMass.pm
```diff
@@ -0,0 +1,225 @@
...
+    my $mass = calculateMass(@result);
+    
+    # Guard against invalid input
+    return if $mass eq 0;
+    
+    # Rebuild string from hash with proper capitilization to ensure that the right molecule was calculated
+    $compound = rebuildString(@result);
+    
+    return "The molar mass of $compound is $mass g/mol.",
+    html => "The molar mass of <strong>" . subInts($compound) . "</strong> is <strong>$mass g/mol</strong>.";
```
mintsoft: Rather than using raw html output, we've deprecated that and started using prebuilt templates.
This means that if you switch this out to something a bit like: https://github.com/duckduckgo/zeroclickinfo-goodies/blob/master/t/DaysBetween.t#L16  then it should just work!

t/MolarMass.t
```diff
@@ -0,0 +1,36 @@
...
+use Test::More;
+use DDG::Test::Goodie;
+
+zci answer_type => "molar_mass";
+zci is_cached   => 1;
+
+ddg_goodie_test(
+    [qw( DDG::Goodie::MolarMass )],
+    
+    'molar mass of h2o' => test_zci(
```
mintsoft: We've started wrapping the output into a function so we can change it easily like: https://github.com/duckduckgo/zeroclickinfo-goodies/blob/master/t/DaysBetween.t#L13

mintsoft:
@Iambecomeroot Hi, apologies this took so long to get to. I've put some feedback against the diff, any questions feel free to ping me

lib/DDG/Goodie/MolarMass.pm
Iambecomeroot: The problem I have with that is it forces the compound to go at then end. For example, `h20 molar mass` wouldn't work.

Iambecomeroot:
I guess @joshliu beat me to it #4219.

PR #3878, opened 2017-01-21
calculator: trigger/safe

zachthompson:
* Update trigger
* Run expression with Safe with minimum ops
* Add tests

Have fun @moollaza ;)

---
IA Page: https://duck.co/ia/view/calculator

### DIFF
--- a/lib/DDG/Goodie/Calculator.pm
+++ b/lib/DDG/Goodie/Calculator.pm
@@ -60,10 +60,10 @@ my $network   = qr#^$ip4_regex\s*/\s*(?:$up_to_32|$ip4_regex)\s*$#;         # Lo
 my $safe = new Safe;
 $safe->permit_only(qw'
     :base_core :base_loop :base_mem :base_math
-    require caller padany
+    rv2gv require caller padany
 ');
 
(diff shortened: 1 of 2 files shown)


daxtheduck:
Metadata placeholder post
            
---
            
**How to find your IA page**
- New Instant Answer
              
- Existing Instant Answer

moollaza:
@zachthompson this should be live on Beta now. One issue I can see is a string like `********23` -- it still throws errors:

```
Argument "*main::*" isn't numeric in exponentiation (**) at (eval 4040)
```

Similarly, `log23/log3` produces:

```
Warning: Use of "log" without parentheses is ambiguous at (eval 4038)
```

I suppose we can bail early if we have more than 1 consecutive operators without a number between them?

zachthompson:
I believe both of these are present on master.  I don't see the error with `********23` on this branch but do with the other.  

Checking for, say, three or more ops in a row is probably a quick guard.

Made an update to the opsets.

moollaza:
> Checking for, say, three or more ops in a row is probably a quick guard.

I'll make a task for this. 

The rest LGTM :+1:

PR #3485, opened 2016-08-03
Countdown Goodie v2

rasikapohankar:
###### What does your Pull Request do (check all that apply)?
- [x] Improvement
  - [x] Enhancement: **`{Countdown}: {Update javascript and implement design changes}`**
###### Description of changes

Changed the javascript code to replace updating of countdown values using `html()` with `meta.rerender`, used `DDH_async_add` to load moment.js, made changes to implement the design suggested by the design team.
In continuation of #2808 
Fix #3500 
###### People to notify (@mention interested parties)

@moollaza 

---

Instant Answer Page: https://duck.co/ia/view/countdown

[Maintainer](http://docs.duckduckhack.com/maintaining/guidelines.html): @rasikapohankar

PR #3936, opened 2017-02-13
Countdown: Continue PR #3485

### LINKED ISSUE #3500: Countdown: Needs to load alarm sound using versioned asset path
In production, all paths for `/share` are versioned based on the current repo release number.

For Goodies, we need to manually build the `/share` path by passing the current Goodie Repo version to the JS and construction the path.

An example of grabbing the Perl version number can be seen in the [SolarSystem](https://github.com/duckduckgo/zeroclickinfo-goodies/blob/master/lib/DDG/Goodie/SolarSystem.pm#L14) Goodie.

We need to use that same approach to ensure the audio asset is successfully loaded.

---

IA Page:  http://duck.co/ia/view/countdown 
[Maintainer](http://docs.duckduckhack.com/maintaining/guidelines.html): @rasikapohankar

rasikapohankar:

Implements changed design, fixes #3500

Design reference:
countdown v2 design

#3500 

## People to notify

@moollaza 

---

Instant Answer Page: https://duck.co/ia/view/countdown


daxtheduck:
Metadata placeholder post
            
---
            
**How to find your IA page**
- New Instant Answer
              
- Existing Instant Answer

lib/DDG/Goodie/Countdown.pm
```diff
@@ -13,10 +13,12 @@ zci answer_type => 'countdown';
 
 zci is_cached => 1;
 
+my $goodieVersion = $DDG::GoodieBundle::OpenSourceDuckDuckGo::VERSION // 999;
```
mintsoft: ?

t/Countdown.t
```diff
@@ -34,9 +40,9 @@ set_fixed_time("2016-04-15T15:31:02Z");
 ddg_goodie_test(
     [qw( DDG::Goodie::Countdown )],
     
-    'time until 1st June 2016'            => build_test("1st June 2016", 4019338 , "01 Jun 2016 00:00:00 EDT"),
-    'how long until 31st December 2016'   => build_test("31st December 2016", 22426138, "31 Dec 2016 00:00:00 EST"),
-    'countdown to tomorrow'               => build_test("tomorrow", 86400, "16 Apr 2016 11:31:02 EDT"),
+    'time until 1st June 2016'            => build_test("1st June 2016", 4019338 , "June 01, 2016, 12:00:00 AM"),
```
mintsoft: Usually when we test something that uses the DateRole we tend to use ISO8601 for the date format; there's no reason to test the Role here, as it's already tested independently

share/goodie/countdown/countdown.js
```diff
@@ -106,48 +72,52 @@ DDH.countdown = DDH.countdown || {};
...
+                        return;
+                    }
+                    hasShown = true;
+
+                    $display = $(".zci--countdown").find(".countdown_container").find('.number');
                     setInterval(function() {
                         duration = getCountdown(duration);
+                        item.set({ year: duration.years(), month: duration.months(), day: duration.days(), hour: duration.hours(), minute: duration.minutes(), second: duration.seconds() });
+                        $(".zci--countdown").find('.c-base__sub').addClass("tx-clr--grey");
                     }, 1000);
```
mintsoft: With an interval of 1000 we might see skips of 2 seconds as the timer firing isn't going to be exact, we can flip this out for a shorter interval if it's something you're concerned about

share/goodie/countdown/countdown.js
```diff
@@ -5,46 +5,13 @@ DDH.countdown = DDH.countdown || {};
...
-        $time_display,
-        $displayYears, $displayMonths, $displayDays,
-        $displayHrs, $displayMins, $displaySecs,
-        $year,$month,$display,
+        $display,
         stopped = false,
         cachedPlayer, soundIsPlaying = false,
         SOUND_NAME = "alarm-sound",
-        soundUrl = 'share/goodie/countdown/alarm.mp3',
+        soundUrl,  // 'share/goodie/countdown/goodie-repo-version/alarm.mp3',
```
mintsoft: Might as well get shot of the comment as it's defined below

t/Countdown.t
```diff
@@ -11,18 +11,24 @@ use DDG::Test::Location;
 zci answer_type => "countdown";
 zci is_cached   => 1;
 
+my $goodieVersion = $DDG::GoodieBundle::OpenSourceDuckDuckGo::VERSION // 999;
```
mintsoft: Again, I'm not sure why this is here/needed ?

t/Countdown.t
```diff
@@ -11,18 +11,24 @@ use DDG::Test::Location;
...
-                remainder => $remainder,
-                difference => $initialDifference,
-                countdown_to => $output_string
+                remainder      => $remainder,
+                difference     => $initialDifference,
+                countdown_to   => $output_string,
+                goodie_version => $goodieVersion
             },
             templates => {
                 group => "text",
```
mintsoft: If you're using a custom template like this `group => 'base'` is preferred

lib/DDG/Goodie/Countdown.pm
rasikapohankar: @mintsoft It is used to form the path of the alarm sound in production ( #3500 )

t/Countdown.t
rasikapohankar: @mintsoft I'm unclear about this. It checks the difference between 2016-04-15T15:31:02Z and 1st June 2016 for the countdown. It's not testing the Role..

t/Countdown.t
rasikapohankar: @mintsoft I had added this to include the `goodie_version` returned from Countdown.pm..

share/goodie/countdown/countdown.js
rasikapohankar: If we make the interval shorter, the countdown will run faster as this renders the template after every second :sweat_smile: I don't think the interval should be changed. Do you think this will be unreliable?

t/Countdown.t
moollaza: `text` is actually necessary here because we're defining the `subtitle` which is handled by the template. If we switch to base we'd need to add more markup for that to our template 👍

rasikapohankar:
@mintsoft @moollaza Can this be deployed to beta?

share/goodie/countdown/countdown.js
mintsoft: I don't think it's a major concern, we'll see how it looks. If we see a 2 second jump then it'll need decreasing, but it's probably fine

t/Countdown.t
mintsoft: The Dates Role parses "1st june 2016", so normally we'd use like `time until 2016-04-15` as the test case

rasikapohankar: @mintsoft I understand now! Updated the test file :+1:

moollaza:
Deploying to beta 👍

share/goodie/countdown/countdown.js
```diff
@@ -106,48 +72,52 @@ DDH.countdown = DDH.countdown || {};
...
+                    if(hasShown) {
+                        return;
+                    }
+                    hasShown = true;
+
+                    $display = $(".zci--countdown").find(".countdown_container").find('.number');
                     setInterval(function() {
                         duration = getCountdown(duration);
+                        item.set({ year: duration.years(), month: duration.months(), day: duration.days(), hour: duration.hours(), minute: duration.minutes(), second: duration.seconds() });
+                        $(".zci--countdown").find('.c-base__sub').addClass("tx-clr--grey");
```
moollaza: @rasikapohankar we should use a CSS rule instead of repeatedly applying a class every second

moollaza:
@rasikapohankar aside from my one comment, this LGTM :+1:

moollaza:
Oh one bug I just noticed. It seems to not account for the daylight savings shift in my timezone. Can you confirm if this is a moment.js bug?

countdown_to_april_11th_at_duckduckgo

Midnight is 5 minutes away, not 1 hour and 5 minutes.

share/goodie/countdown/countdown.js
rasikapohankar: @moollaza This was to apply the class to the `subtitle`. I'll include the subtitle in the handlebars and apply to it the same class as the already defined `subtitle` (`.c-base__sub`). It is working on duckpan. Just wanted to confirm that it will not break anything else?

rasikapohankar:
> Can you confirm if this is a moment.js bug?

@moollaza We have only parsed the input string using the Dates Role and found the initial difference between that date and `now`. :slightly_frowning_face:

I had tried to calculate the difference using moment.js at the very start but removed that part later. I will check if it gives the correct answer.

mintsoft:
@rasikapohankar @moollaza I think that problem comes from the fact that `$diff` has been calculated by subtracting the `epoch` from each other. IIRC `$dt->delta_ms( $datetime )` should be used for DST aware differences; however it may not have enough information to accurately determine the DST settings for the user. If that's the case, it probably makes sense to use the role to normalise the date into ISO8601 and then pass that through to the frontend, use moment to calculate the diff so that it can handle the timezone based on the browser settings

rasikapohankar:
@moollaza @mintsoft I've changed the code to calculate the difference in JS using momentjs and used it to show the countdown, also kept the parsing of the date from query and difference calculation in Perl as it is to avoid going ahead on incorrect input. Please take a look.

@moollaza After using template type option `content`, the subtitle was shown at the top. So removed the class applied to `subtitle` completely, it is now #888.
Also corrected a mistake at the end of the countdown.

@moollaza Can you please deploy to beta again?

moollaza:
@rasikapohankar Thanks! This LGTM :+1:

Great work 💯

rasikapohankar:
@moollaza  Thanks for merging! I was not aware that this had been deployed to beta. It must have shown the correct answer now?

Issue #3760, opened 2016-11-17
President of the United States (POTUS): shouldn't trigger on president-elect


bsstoner:
<img width="697" alt="screen shot 2016-11-17 at 7 48 28 am" src="https://cloud.githubusercontent.com/assets/126358/20389950/511af736-ac9a-11e6-807a-5871190c6d19.png">
   

------
IA Page:  http://duck.co/ia/view/potus 
[Maintainer](http://docs.duckduckhack.com/maintaining/guidelines.html): @numbertheory

jasonporta:
I'm new at contributing on GitHub (I have my own projects, I use GitHub desktop, and I've completed the set up process for Codio and everything). I found potus.pm and this seems like an easy fix (is it even still an issue after 25 days?) Do I discuss it in Slack or add my edits and submit a pull request? Through Codio or GitHub Desktop? 

Sorry. Like I said, I'm new at this and I'm trying to get involved so I can contribute to more projects and in grow my portfolio to get a programming job in the process.

mintsoft:
This is actually really strange; with the query `who is the president-elect of the us` the value of `$rem` on https://github.com/duckduckgo/zeroclickinfo-goodies/blob/master/lib/DDG/Goodie/POTUS.pm#L22 is `who is the` whereas I wouldn't expect it to trigger at all because the string `president of the us` isn't present in the query.

Feels like it might be broader than just this IA's interesting triggering mechanism misfiring

nalinbhardwaj:
@bsstoner @mintsoft I can't seem to be able to trigger the IA. Was it shut down due to the aforementioned issues?

moollaza:
> I can't seem to be able to trigger the IA. Was it shut down or something due to the aforementioned issues?

Yes it was taken offline

> whereas I wouldn't expect it to trigger at all because the string president of the us isn't present in the query.

@mintsoft that's really odd...possibly a bug in DuckDuckGo Repo.

Either way, one big issues is that the regex inside the `handle` is not anchored. We need to ensure that the query **only** matches the phrases we expect and nothing else.

For example it currently triggers on "who is the worst president of the us" which it should not!

mintsoft:
@moollaza is it worth fixing this or has it been replaced by the wikipedia fatheat thingymabob?

bsstoner:
It will likely be made irrelevant in the near future.

moollaza:
[...]

PR #3938, opened 2017-02-15
POTUS: Triggering is way too keen

mintsoft:
Fixes https://github.com/duckduckgo/zeroclickinfo-goodies/issues/3760

Unfortunately, the regex isn't actually matching the query, it's replacing bits that we don't need so we can pass the remainder on to `words2nums` to convert. This was defaulting to the current president if it couldn't parse, which basically means it would show the current president for anything that isn't "first" "second" etc i.e. "elect" or "vice".

Whether this is worth fixing or not I'll leave up to the community/staff. If this doesn't work/isn't resliant enough, I'd suggest starting from scratch from a guard regex instead.

## People to notify
@moollaza @bsstoner 

---

Instant Answer Page: https://duck.co/ia/view/potus

### DIFF
--- a/lib/DDG/Goodie/POTUS.pm
+++ b/lib/DDG/Goodie/POTUS.pm
@@ -24,11 +24,11 @@ handle remainder => sub {
       (^who\s+(?<iswas>is|was)\s*(?:the\s*)?(?<num>.*?)$)
       |(^(?<num>.*)$)
     /gix;
-   use Data::Dump qw(dump); 
-    my $num = $+{num}; 
-    warn($num);
-    $num = $prez_count if $num eq "";
-    $num = words2nums($num) if words2nums($num);
+    
+    my $num;
+    $num = $prez_count if $+{num} eq "";
+    $num = $prez_count -1 if $+{num} eq "" and $+{iswas} eq "was";
+    $num = words2nums($+{num}) if words2nums($+{num});
     return unless $num;
 
     my $index = $num - 1;
--- a/lib/DDG/Goodie/POTUS.pm
+++ b/lib/DDG/Goodie/POTUS.pm
@@ -16,15 +16,19 @@ zci is_cached   => 1;
 
 my @presidents = @{LoadFile(share('presidents.yml'))};
 my $prez_count = scalar @presidents;
+my $potus_or_president = qr/(potus|president of the (us|united states))/i;
 
-handle remainder => sub {
+handle query_lc => sub {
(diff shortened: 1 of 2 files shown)


daxtheduck:
Metadata placeholder post
            
---
            
**How to find your IA page**
- New Instant Answer
              
- Existing Instant Answer

moollaza:
@mintsoft seems like it doesn't trigger on `who is the president of the united states` ?

I think a guard regex is the best plan. Also "Who was" should probably show the previous president?

mintsoft:
@moollaza I've implemented a guard regex here, and I think I've exposed a bug in duckpan/zci . 

I've got a failing test at the moment for `who is the president-elect of the us`, which shouldn't even trigger based on the triggers: https://github.com/duckduckgo/zeroclickinfo-goodies/pull/3938/files#diff-66449435cf1d3b4e87ded23d450483eaR12  

https://travis-ci.org/duckduckgo/zeroclickinfo-goodies/builds/205568133#L1010

I'm guessing that it's not matching that as an array of strings verbatim, but an array of words or something. I've not had a chance to dig into duckpan yet.

mintsoft:
Also just confirmed it's just tests, it happens in `duckpan server` and `duckpan query`

moollaza:
@mintsoft thanks for confirming, I'm investigating. I think you're correct that it's a bug in `DDG::Block::Words`

You can see how it breaks `president-elect` into `president` and `elect` and then checks the following words which match the trigger phrase.

I think we should probably not break dash separated words into sub-words. That change could have a potentially very big effect on triggering though...

moollaza:
@mintsoft can you add another guard to ensure we're dealing with `president\s` and not `president-.+`? That's the easiest fix for now.

mintsoft:
@moollaza I'll add that guard in

mintsoft:
Also I think the bug is more subtle than that; if it was just as you described then "who is the president elect of the united states" should trigger, which it doesn't. Unless I've misinterpreted?

mintsoft:
@moollaza I've flipped out to `query_lc` and then re-implemented the replacements of the triggers with empty sting before the guard; that way, if we fix up duckpan/`DDG::Block::Words` then we can easily flip back to 'remainder' and delete https://github.com/duckduckgo/zeroclickinfo-goodies/pull/3938/files#diff-66449435cf1d3b4e87ded23d450483eaR25

moollaza:
Thanks! This LGTM :+1:

Issue #3984, opened 2017-03-07
Conversions: Round to appropriate decimal place


pjhampton:
### Description

User feedback suggests that [rounding to 3 decimal places is confusing](https://duckduckgo.com/?q=cup+to+gallon&ia=answer). 

A more appropriate response for this search would be **1 Cup = 0.0625 Gallon**. We could do with a solution to display the appropriate number of decimal places.

## Get Started
- [ ] 1) Claim this issue by commenting below
- [ ] 2) Review our [Contributing Guide](https://github.com/duckduckgo/zeroclickinfo-goodies/blob/master/CONTRIBUTING.md)
- [ ] 3) [Set up your development environment](https://docs.duckduckhack.com/welcome/setup-dev-environment.html), and fork this repository
- [ ] 4) Create a Pull Request

## Resources
- Join [DuckDuckHack Slack](https://quackslack.herokuapp.com/) to ask questions
- Join the [DuckDuckHack Forum](https://forum.duckduckhack.com/) to discuss project planning and Instant Answer metrics
- Read the [DuckDuckHack Documentation](https://docs.duckduckhack.com/) for technical help

---

Instant Answer Page: https://duck.co/ia/view/conversions

mintsoft:
@pjhampton This might be fixed here already: https://github.com/duckduckgo/zeroclickinfo-goodies/pull/3917

PR #3988, opened 2017-03-07
Conversions: Improving result aesthetics 

mintsoft:
Previously, unless the result was large/small enough to warrant scientific notation, it was always given to 3 DP. That's basically fine for large-ish numbers, like 20.345 as the result, however it's crap for 1e-6 < x < 1 as this is where fractions like 1/8 live and it looks crap. 

So now, if the result is >1 and <10,000,000 then we use 3dp (for nice integer handling of things like 2^20 mainly), the same as before. However if it's < 1 then we use 3 significant figures. Everywhere else it's scientific notation, again unchanged.

The `Math::SigFigs` module I already have on my ddh box, so this may be pulled in already by an existing package, if not then it'll need flagging as `needs new perl dependency`.

If it would require the dependency adding to production, I'll probably remove it and implement the functionality in a role instead rather than wait. 

Fixes: https://github.com/duckduckgo/zeroclickinfo-goodies/issues/3984

## People to notify
@pjhampton 

---

### DIFF
--- a/lib/DDG/Goodie/Conversions.pm
+++ b/lib/DDG/Goodie/Conversions.pm
@@ -6,6 +6,7 @@ use DDG::Goodie;
 with 'DDG::GoodieRole::NumberStyler';
 
 use Math::Round qw/nearest/;
+use Math::SigFigs qw/:all/;
 use utf8;
 use YAML::XS 'LoadFile';
 use List::Util qw(any);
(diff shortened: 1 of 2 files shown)


daxtheduck:
Metadata placeholder post
            
---
            
**How to find your IA page**
- New Instant Answer
              
- Existing Instant Answer

moollaza:
Deploying to beta...

moollaza:
Deployed and working. Thanks! This LGTM :+1:

@mintsoft I checked and we have `Math::SigFigs` installed already 👍

Issue #4004, opened 2017-03-10
Time Zone: Support "Time in <time_zone>" queries 


pjhampton:
### Description

Based on user feedback and anonymized search data, DuckDuckGo isn't supporting queries such as "Time in <time_zone>". To further complicate things, the API that we're using for the [TimeZone Converter](https://duck.co/ia/view/timezone_converter), [Location Sunrise/Sunset](https://duck.co/ia/view/sun_rise_set), and [Holiday](https://duck.co/ia/view/holiday) doesn't support time zone based lookups 😞 

![screen shot 2017-03-10 at 8 07 20 pm](https://cloud.githubusercontent.com/assets/8960296/23811277/36a9b040-05cd-11e7-9373-f4598b0b90d7.png)
### Solution 

There is no perfect solution here, however, the most brainstormed idea could use a hash that maps time zones to countries and uses the time for a given country to show the time for the specified timezone. 

https://en.wikipedia.org/wiki/List_of_time_zones_by_country

We could start off with the most popular as a proof of concept, and scale from there. The most popular time zones include:

- PST
- CST
- EST
- UTC 
- GMT (probably BST too!)

**Triggers should include:**
time in <time_zone>
Time now <time_zone>
<time_zone> now time

## People to notify

@duckduckgo/duckduckhack-contributors 

## Get Started
- [ ] 1) Claim this issue by commenting below
- [ ] 2) Review our [Contributing Guide](https://github.com/duckduckgo/zeroclickinfo-goodies/blob/master/CONTRIBUTING.md)
- [ ] 3) [Set up your development environment](https://docs.duckduckhack.com/welcome/setup-dev-environment.html), and fork this repository
- [ ] 4) Create a Pull Request

## Resources
- Join [DuckDuckHack Slack](https://quackslack.herokuapp.com/) to ask questions
- Join the [DuckDuckHack Forum](https://forum.duckduckhack.com/) to discuss project planning and Instant Answer metrics
- Read the [DuckDuckHack Documentation](https://docs.duckduckhack.com/) for technical help

---

Instant Answer Page: **To Be Created** 🚀

BillyBarbaro:
I've already been looking into this one a little in issue [3181](https://github.com/duckduckgo/zeroclickinfo-spice/issues/3181) in the Spice repo.  It seems to me like the work on that should cover this issue.

BillyBarbaro:
[...]

PR #4242, opened 2017-06-03
Timezonetime: Instant Answer POC 

shivam99aa:
New timezonetime Goodie

## Description
This IA will give time in a particular timezone as entered by the user. Currently this is a POC so supporting only EST, UTC, GMT, PST, BST, CST and IST timezones

## Discussion
Related to discussion in #4004 

## People to notify
@pjhampton 

Instant Answer Page: https://duck.co/ia/view/timezonetime

### DIFF
--- a/lib/DDG/Goodie/Timezonetime.pm
+++ b/lib/DDG/Goodie/Timezonetime.pm
@@ -1,0 +1,85 @@
+package DDG::Goodie::Timezonetime;
+# ABSTRACT: Write an abstract here
+
+# Start at http://docs.duckduckhack.com/walkthroughs/calculation.html if
+# you are new to instant answer development
+
+use DDG::Goodie;
+use strict;
+use warnings;
+
+use DateTime;
+use DateTime::TimeZone;
+
+zci answer_type => 'timezonetime';
+
+# Caching - http://docs.duckduckhack.com/backend-reference/api-reference.html#caching`
+zci is_cached => 1;
+
+# Triggers at the start of the query
+my @triggerStart = ("time in", "time now");
+# Triggers at the end of the query
+my @triggerEnd   = ("now time");
+# Final trigger list
+my @triggersFinal = (@triggerStart, @triggerEnd);
+
+# Triggers - http://docs.duckduckhack.com/walkthroughs/calculation.html#triggers
+triggers any => @triggersFinal;
+
+# Mapping short timezone names to one used by DateTime:Timezone module
+my $timezoneMapping = {"IST" => "Asia/Kolkata", "EST" => "EST", "UTC" => "UTC", "GMT" => "GMT",
+                       "PST" => "PST8PDT", "BST" => "Europe/London", "CST" => "CST6CDT"
+                      };
+
+# Handle statement
+handle query_lc => sub {
+
+    my $query_lc = $_;
+    
+    my $timezones = join('|', keys(%$timezoneMapping));
+    my $daylightStatus = "";
+    
+    # Parsing timezone out of query string
+    $query_lc =~ s/.*\b($timezones)\b.*/$1/ig;
+    my $timezone = uc($query_lc);
+    
+    # If timezone is not in hash then return null
+    unless (exists($timezoneMapping->{$timezone})) {
+	return;
+    }
+
+    # Getting corresponding timezone value from hash
+    my $mappedTimezone = $timezoneMapping->{$timezone};
+
+    # Get time for desired timezone
+    my $tz = DateTime::TimeZone->new( name => $mappedTimezone );
+    my $dt = DateTime->now();
+    my $offset = $tz->offset_for_datetime($dt);
+    $dt->add(seconds => $offset);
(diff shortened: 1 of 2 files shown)


daxtheduck:
Metadata placeholder post
            
---
            
**How to find your IA page**
- New Instant Answer
              
- Existing Instant Answer

lib/DDG/Goodie/Timezonetime.pm
```diff
@@ -0,0 +1,85 @@
+package DDG::Goodie::Timezonetime;
+# ABSTRACT: Write an abstract here
```
moollaza: Please remove the comments from the boilerplate 👍

lib/DDG/Goodie/Timezonetime.pm
```diff
@@ -0,0 +1,85 @@
...
+
+# Triggers at the start of the query
+my @triggerStart = ("time in", "time now");
+# Triggers at the end of the query
+my @triggerEnd   = ("now time");
+# Final trigger list
+my @triggersFinal = (@triggerStart, @triggerEnd);
+
+# Triggers - http://docs.duckduckhack.com/walkthroughs/calculation.html#triggers
+triggers any => @triggersFinal;
```
moollaza: @shivam99aa this is the wrong approach. `any` means this will trigger when those phrases appear anywhere in the query.

You can define both start and end triggers for a single IA so that's what we want here:

```perl
triggers start => 'time in', 'time now';
triggers end => 'now time';
```

t/Timezonetime.t
```diff
@@ -0,0 +1,55 @@
...
+use warnings;
+use Test::More;
+use Test::MockTime qw(:all);
+use Test::Deep;
+use DDG::Test::Goodie;
+
+zci answer_type => 'timezonetime';
+zci is_cached   => 1;
+
+my $goodieVersion = $DDG::GoodieBundle::OpenSourceDuckDuckGo::VERSION // 999;
```
moollaza: You can remove this 👍

t/Timezonetime.t
```diff
@@ -0,0 +1,55 @@
...
+                #
+                # }
+            }
+        };
+}
+
+# Use this to build expected results for your tests.
+sub build_test { test_zci(build_structured_answer(@_)) }
+
+set_fixed_time('2017-06-03T09:36:53Z');
```
moollaza: Let's add a more tests to ensure the DST logic works as expected 👍

lib/DDG/Goodie/Timezonetime.pm
moollaza: I think you'll want "time" as an `end` trigger as well to handle queries like "pst time"

moollaza: @shivam99aa actually, it may be best to do the reverse, and use the time zones as the triggers. Then we can use a regular expression in the `handle` to determine if the query matches an expected format.

lib/DDG/Goodie/Timezonetime.pm
```diff
@@ -0,0 +1,85 @@
...
+my $timezoneMapping = {"IST" => "Asia/Kolkata", "EST" => "EST", "UTC" => "UTC", "GMT" => "GMT",
+                       "PST" => "PST8PDT", "BST" => "Europe/London", "CST" => "CST6CDT"
+                      };
+
+# Handle statement
+handle query_lc => sub {
+
+    my $query_lc = $_;
+    
+    my $timezones = join('|', keys(%$timezoneMapping));
```
moollaza: No need to do this every time the `handle` runs. We can create this once outside the `handle` and use it inside so the hash is stored in memory.

lib/DDG/Goodie/Timezonetime.pm
```diff
@@ -0,0 +1,85 @@
...
+# Triggers at the end of the query
+my @triggerEnd   = ("now time");
+# Final trigger list
+my @triggersFinal = (@triggerStart, @triggerEnd);
+
+# Triggers - http://docs.duckduckhack.com/walkthroughs/calculation.html#triggers
+triggers any => @triggersFinal;
+
+# Mapping short timezone names to one used by DateTime:Timezone module
+my $timezoneMapping = {"IST" => "Asia/Kolkata", "EST" => "EST", "UTC" => "UTC", "GMT" => "GMT",
```
moollaza: We also need the names for these time zones like "eastern time" and "pacific time" so those can be used as triggers also. It may be best to use a store this data in a YAML file so we can easily load it into the Perl and deal with it as needed.

moollaza:
Ping! Would you be able to finish this one up? We'd love to get this live!

pjhampton:
Thanks @shivam99aa and @moollaza 
This is a high priority so it was pretty urgent to get finished up. I finished all the code in https://github.com/duckduckgo/zeroclickinfo-goodies/pull/4393. I have added your attribution to the duck.co/ia page @shivam99aa - thanks for all your hard work!

PR #4261, opened 2017-06-07
Conversions: Support single unit input, case-insensitive fallback check

moollaza:
# Description of new Instant Answer, or changes
* Allow for a single unit input with an amount to be converted.
  E.g. "35 cm", "1 acre", "200 lbs"

  **Note**: Because we use word triggers, only single-unit queries where the number and unit are separated by a space (or dash) will work. A query like "35m" will not trigger currently. We may enable that later with internal triggering improvements.

* Lowercased additional units in the YAML that don't require case-sensitive comparisons. And added case-insensitive check comparing lowercased match against symbol. This will allow all the energy units to use lowercased units in queries. 
  E.g. "covert w to kw", "35kw to w"

* Corrected the symbol for kilojoules so that queries like `3 kj to j` will work.

## People to notify
@pjhampton 

## Testing & Review

**Pull Request**
- [ ] Description contains a valid Instant Answer Page Link (e.g. https://duck.co/ia/view/my_ia)

**Code**
- [ ] Tests are passing (run `$ duckpan test <goodie_id>`)

**Ready to merge?**

---

### DIFF
--- a/lib/DDG/Goodie/Conversions.pm
+++ b/lib/DDG/Goodie/Conversions.pm
@@ -117,13 +117,14 @@ handle query => sub {
     my $left_num = $+{'left_num'};
     my $right_unit = $+{'right_unit'} // "";
     my $right_num = $+{'right_num'} // "";
+    my $question = $+{'question'} // "";
+    my $connecting_word = $+{'connecting_word'} // "";
 
     my $factor = $left_num;
     my @matches = ($left_unit, $right_unit);
 
-
     # ignore conversion when both units have a number
-    return if $left_num && $right_num;
+    return if ($left_num && $right_num) || $left_unit && !($left_num || $right_unit);
 
     # Compare factors of both units to ensure proper order when ambiguous
     # also, check the <connecting_word> of regex for possible user intentions
@@ -168,8 +169,8 @@ handle query => sub {
         "" ne $right_num
         || (   "" eq $left_num
             && "" eq $right_num
-            && $+{'question'} !~ qr/convert/i
-            && $+{'connecting_word'} !~ qr/to/i ))
+            && $question !~ qr/convert/i
+            && $connecting_word !~ qr/to/i ))
     {
         $factor = $right_num;
         @matches = reverse @matches;
--- a/lib/DDG/Goodie/Conversions.pm
+++ b/lib/DDG/Goodie/Conversions.pm
@@ -218,7 +218,7 @@ sub get_matches {
 
     foreach my $match (@input_matches) {
         foreach my $type (@types) {
-            if (($type->{'symbols'} && grep { $_ eq $match } @{$type->{'symbols'}})
+            if (($type->{'symbols'} && grep { lc $_ eq lc $match } @{$type->{'symbols'}})
              || lc $match eq lc $type->{'unit'}
              || grep { $_ eq lc $match } @{$type->{'aliases'}} ) {
                 push(@output_matches,{
--- a/lib/DDG/Goodie/Conversions.pm
+++ b/lib/DDG/Goodie/Conversions.pm
@@ -237,18 +237,18 @@ sub get_matches {
 
 sub convert {
     my ($conversion) = @_;
-
-    my @matches = get_matches($conversion->{'from_unit'}, $conversion->{'to_unit'});
-	return if scalar(@matches) != 2;
+    my @inputs = ($conversion->{'from_unit'});
+    push @inputs, $conversion->{'to_unit'} if defined $conversion->{'to_unit'};
+    my @matches = get_matches(@inputs);
+	return if scalar(@matches) < 1;
     return if $conversion->{'factor'} < 0 && !($matches[0]->{'can_be_negative'});
 
     # matches must be of the same type (e.g., can't convert mass to length):
-    return if ($matches[0]->{'type'} ne $matches[1]->{'type'});
+    return if (scalar(@matches) > 1 && $matches[0]->{'type'} ne $matches[1]->{'type'});
 
     return {
-        "result" => "",
         "from_unit" => $matches[0]->{'unit'},
-        "to_unit" => $matches[1]->{'unit'},
+        "to_unit" => $matches[1]->{'unit'} // "",
         "type"  => $matches[0]->{'type'}
     };
 }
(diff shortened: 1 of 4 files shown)


daxtheduck:
Metadata placeholder post
            
---
            
**How to find your IA page**
- New Instant Answer
              
- Existing Instant Answer

lib/DDG/Goodie/Conversions.pm
```diff
@@ -96,17 +105,26 @@ handle query => sub {
...
     # hack support for "degrees" prefix on temperatures
-    $_ =~ s/ degree[s]? (centigrade|celsius|fahrenheit|rankine)/ $1/i;
+    $_ =~ s/ degree[s]? (centigrade|cel(?:s|c)ius|fah?renheit|rankine)/ $1/i;
 
     # hack - convert "oz" to "fl oz" if "ml" contained in query
     s/(oz|ounces)/fl oz/i if(/(ml|cup[s]?|litre|liter|gallon|pint)/i && not /fl oz/i);
 
     # guard the query from spurious matches
-    return unless $_ =~ /$guard/;
+    return unless $_ =~ m/$guard/;
```
mintsoft: ? What significance does that have here?

lib/DDG/Goodie/Conversions.pm
```diff
@@ -211,18 +236,18 @@ sub get_matches {
 
 sub convert {
     my ($conversion) = @_;
-
-    my @matches = get_matches($conversion->{'from_unit'}, $conversion->{'to_unit'});
-	return if scalar(@matches) != 2;
+    my @inputs = ($conversion->{'from_unit'});
+    push @inputs, $conversion->{'to_unit'} if defined $conversion->{'to_unit'};
+    my @matches = get_matches(@inputs);
+	return if scalar(@matches) < 1;
```
mintsoft: Hard tab here

moollaza:
Deploying to Beta 👍

share/goodie/conversions/triggers.yml
```diff
@@ -950,7 +950,7 @@ aliases:
   - joules
 type: energy
 unit: joule
-symbols: [J]
+symbols: [j]
```
mintsoft: Why do these need to be lowercase now? Isn't the desired behaviour to be case sensitive unless it's unambiguous? I.e. MB and Mb are treated correctly whilst kj and KJ both work?

share/goodie/conversions/triggers.yml
```diff
@@ -1195,49 +1195,49 @@ aliases:
...
-symbols: [kW]
+symbols: [kw]
 ---
 aliases:
   - megawatt
   - megawatts
 type: power
 unit: megawatt
-symbols: [MW]
+symbols: [mw]
```
mintsoft: Hmm, mw/mW is milliwatt, MW is megawatt

lib/DDG/Goodie/Conversions.pm
moollaza: whoops, removing.

share/goodie/conversions/triggers.yml
moollaza: Yes that's still the desired behaviour, but it seems somehow we lost that. Right now we only compare the original query against symbols, so we're missing a lot of queries:
https://duckduckgo.com/?q=1+w+to+gw&t=brave&ia=web (same applies for j and kj)

IMO we need to re-organize this data file and clean it up as it serves a slightly different purpose now. The aliases and symbols are all mixed up and they don't really serve much of a purpose anymore as all we're using it for is to determine if an acceptable unit exists, and pass along its type.

I think if we separate out the SI and Binary prefixes from the units that would significantly reduce the size of this file. As well, we should do more upfront normalization, like stripping spaces, periods, etc so the aliases don't have to be so redundant. It would also be good to handle plurals (like remove `/s$/`) and check for a match against a unit name/symbol.

I've reverted the lowercasing of the symbols, but I've added an additional check that finally compares the lowercased symbols against the lowercased query.

How's that sound?

moollaza: > but I've added an additional check that finally compares the lowercased symbols against the lowercased query.

Just realized that isn't going to work...I think our matching logic needs to be re-considered.

moollaza: I think my approach is best for now. Anything that is case insensitive should be lowercased in the YAML. That seems to work.

share/goodie/conversions/triggers.yml
moollaza: I think for now we'll just have to accept mw as megawatt. Google doesn't even handle watts, and Bing only handles mega, so `mw` is MegaWatt. Most users aren't doing cased searches and I think if they really need to the can adjust our interpretation in the UI. 

We don't currently support milliwatt so I think this is acceptable for now

share/goodie/conversions/triggers.yml
mintsoft: > How's that sound?

Yeah sounds about right; there is a *lot* of unnecessary code in the perl now. Although If we make the the backend completely case insensitive again there will be a difficulty in situations where case is needed to determine type (S - Siemens, s - seconds for example)

lib/DDG/Goodie/Conversions.pm
moollaza: Fixed

share/goodie/conversions/triggers.yml
mintsoft: Yeah, if the handling of the actual interpretation is handed off the Math.js, it should get it right anyway

pjhampton:
This all looks fine to me. I wasn't particularly sure that defaulting to the same unit was a good idea (See Screenshot 1 + 2) so I adjusted some logic in https://github.com/duckduckgo/zeroclickinfo-goodies/pull/4261/commits/baef3d6e723af27851a9ed858863707c760f1d94 (See Screenshot 3 + 4)

### Screenshot 1
mhz_mhz

### Screenshot 2
gram_gram

### Screenshot 3
screen shot 2017-06-15 at 3 26 29 pm

### Screenshot 4
screen shot 2017-06-15 at 3 11 31 pm

On a foot note, supporting queries such as **5m** might not be such a great idea due to a number of ambiguities that we would have to handle. For example, `5s`, `3m`. If we do go down this root I can imagine it would be done on the backend or just setting the signal to low. 

Regardless, nice work @moollaza 💯  🤠  💯

bsstoner:
This was released to production 🚀

PR #4407, opened 2017-08-08
New Router Gateway Cheat Sheet

benabrahamson:
This cheat sheet contains a list of links for popular router brands' gateway IP address.

---

Instant Answer Page: https://duck.co/ia/view/router_gateway_reference


daxtheduck:
Metadata placeholder post
            
---
            
**How to find your IA page**
- New Instant Answer
              
- Existing Instant Answer

benabrahamson:
Hmmm... looks like it doesn't like the IA page I made. What can I do to solve this?

share/goodie/cheat_sheets/json/routergateway.json
```diff
@@ -0,0 +1,1009 @@
+{
+    "id": "router_gateway_reference",
```
kirkins: As noted in Travis build log

File name must end with _cheat_sheet

kirkins: **Correction** ID must be file name + _cheat_sheet

So in this case your json file will be called: `router-gateway-reference`

The ID will be: `router_gateway_reference_cheat_sheet`

We try to keep the style the same across all cheat sheets

If in doubt reference the cheat sheet folder https://github.com/duckduckgo/zeroclickinfo-goodies/tree/master/share/goodie/cheat_sheets/json

kirkins:
@benabrahamson also could run the json file through this beautifier with indent set to 4.

benabrahamson:
Hi @kirkins, thanks for your help. I have successfully beautified the file, as you said, however I am still having issues with the ID. 

According to the Travis Build log
`Cheat sheet with ID "router_gateway_reference_cheat_sheet" not found in metadata.`

I think the issue might lie with the IA page itself not having the right ID.

(In addition, optimally I think that maybe the actual name of the cheat sheet should drop the "reference" part all together - it is not a necessary part of the name.)

I seem unable to edit the IA page to fix the renaming and ID issue. How should I go about doing that?

Thanks again!

share/goodie/cheat_sheets/json/router-gateway-reference.json
```diff
@@ -0,0 +1,836 @@
...
+        "Planet",
+        "Repotec",
+        "Senao",
+        "Siemens",
+        "Sitecom",
+        "SMC Networks",
+        "Sonicwall",
+        "SpeedTouch",
+        "Sweex",
+        "Thomson",
```
kirkins: You have a Tenda section but it's not in the section order array. Can you add it above Thomson.

`Failed test 'FAIL: The following sections were used in 'sections' but not defined in 'section_order:' ('Tenda')'`

kirkins:
@benabrahamson I will get back to you on the ID issue if someone else doesn't chime in first.

benabrahamson:
@kirkins I've added "Tenda" to to section order. Must have missed that one! Thanks :+1:

kirkins:
@benabrahamson great it looks like it's passing the tests now. I'm just here to help people get their cheat sheets into the expected format. 

Someone from duckduckgo will be able to give you more feedback or accept it as is and merge it.

mintsoft:
You could add `Fortigate`; they're usually on `192.168.1.99` or `192.168.1.2`

benabrahamson:
@sahildua2305 I have actually already made the IP addresses into links :+1:

benabrahamson:
@mintsoft Thanks, have added that suggestion. I have also added a "Generic" section for routers that are not listed.

sahildua2305:
@benabrahamson Sorry, I wasn't clear probably. I meant it doesn't really make sense to have those IPs as links. Why do you think the otherwise?

mintsoft:
@benabrahamson ping! Just checking in, are you OK with the feedback?

benabrahamson:
@sahildua2305 My reasoning behind having the IPs as links was to make it easier to access the router gateway from the browser. i.e if someone wanted to change the SSID of their network, for example, they could click on the link to access the router config page, and do that, or whatever.

sahildua2305:
@benabrahamson May be I am understanding it wrong. Please correct me if that's the case.

According to my understanding, people will need to copy these router gateway IPs if they want to use any of these. So making them as links doesn't make sense to me. Plain text will be better. What do you think about this?

benabrahamson:
@sahildua2305 People would need to copy them if they wanted to use the IP in the command line or something like that, but if they wanted to access their router config page, they could just click the link. I'm talking about the config page accessed via your browser.

sahildua2305:
Okay, looks good then. Thanks for explaining @benabrahamson 😄

Issue #4344, opened 2017-07-10
Roman Numerals Suggestion: Improve IA


pjhampton:
### Description

**This will require a developer comfortable working with js**. We aim to get High Impact tasks completed and merged in a week.

There is quite a lot of queries per week ( 1000 > x < 2000 ) looking for [Roman Numeral](https://en.wikipedia.org/wiki/Roman_numerals) Conversions. **We should improve the triggering and interactivity.** A random dump of queries includes:

```
- roman numerals
- roman numeral
- roman numeral chart
- roman numeral convert
- roman numeral converter
- roman numbers
- xv roman numerals
- <number> in roman numerals
- <number> in roman numbers
- convert to roman numerals
- convert <number> to roman
etc
```

The developer could use the UI from one of the following IAs:

- [Conversions](https://duckduckgo.com/?q=1337+m+to+cm&atb=v1&ia=answer)
- [js minify](https://duckduckgo.com/?q=minify+js&atb=v1&ia=answer)

## People to notify

@duckduckgo/duckduckhack-contributors 

---

Instant Answer Page: https://duck.co/ia/view/roman

gargaml:
I'd like to work on this :-)

pjhampton:
Thanks @gargaml 👍

fenilgandhi:
The IA does not work for "4999 in roman [numerals"``](https://duckduckgo.com/?q=4999+in+roman+numerals&ia=web)

gargaml:
@fenilgandhi This is true. Currently the upper bound is 3999 as it is the greatest number you can write without using non-ascii characters. The current algorithm is extensible so if this is an issue, we can go a bit further. 
I don't know how many queries involve spelled numbers. There are some rules for numbers smaller than 10. If it's useful, I'll add it (or you can actually get involved if you want to :-))

I'm going to add a warning when the user uses number greater than the upper bound to avoid confusion.

PR #4413, opened 2017-08-09
Roman numerals

gargaml:
Some new features added to the roman numerals IA,

A new converter has been added to the client side. When the IA is triggered by the query, a converter appears with the possibility to change values dynamically.
For instance, if the query is roman XV, the server side pre-computes the arabic number and send the answer to the client. But then, the user can actually modify the value to do more. If the query requests a conversion from arabic to roman, the interface swaps input fields automatically.

#4344

The interface can be meliorated but it is already functional. Maybe we could add a button in the middle to replace the '=' sign and allows the user to swap fields dynamically.

## People to notify
@pjhampton

## Testing & Review

**Pull Request**
- [ ] Description contains a valid Instant Answer Page Link (e.g. https://duck.co/ia/view/roman)

**Instant Answer Page** (for new Instant Answers)
- [ ] Instant Answer page is correctly filled out and contains:
    - The **Title** is appropriately named and formatted
    - The **IA topics** are present and appropriate
    - The **Description** is clear and coherent
    - **Source Name** exists if applicable
    - All **Example Queries** trigger on [Beta](https://beta.duckduckgo.com/)

**Code**
- [ ] Tests are passing (run `$ duckpan test roman`)

**Ready to merge?**

---

Instant Answer Page: https://duck.co/ia/view/roman


daxtheduck:
Metadata placeholder post
            
---
            
**How to find your IA page**
- New Instant Answer
              
- Existing Instant Answer

share/goodie/roman/roman.handlebars
```diff
@@ -0,0 +1,11 @@
+<form class="converter">
```
moollaza: nit: this should be named `content.handlebars`

share/goodie/roman/roman.js
```diff
@@ -0,0 +1,218 @@
...
+    }
+    
+    DDH.roman.build = function(ops) {
+        var input = ops.data.input;
+        var output = ops.data.output;
+        var inputValue = ops.data.input_value;
+        var outputValue = ops.data.output_value;
+        
+        return {
+            onShow: function() {
```
moollaza: You need to add a flag the ensures the init logic only runs once. `onShow` is fired each time the tab is clicked, so this has the potential to run several times.

e.g.

```js
var initiated = false;
...
onShow: function(){
  if (!initiated) {
     // setup event handlers, classes, etc
     initiated = true;
  }
}
```

share/goodie/roman/roman.css
```diff
@@ -0,0 +1,29 @@
+.converter {
```
moollaza: @gargaml please prefix all these with `.zci--roman` to ensure there are no collisions with other classes on the page.

We wrap the IA in a div with a unique class to ensure this.

lib/DDG/Goodie/Roman.pm
```diff
@@ -5,35 +5,82 @@ use strict;
...
-        $in  = uc($in);
-        $out = arabic($in);
+    # By default, we convert from roman to arabic.
+    my $input = 'roman';
+    my $input_value = '';
+    my $output = 'arabic';
+    my $output_value = '';
+ 
+    # These two lists are used to load the converter without any answer.
+    my @roman_to_arabic = (
```
gargaml: I almost forgot that handle is a function. Done.

share/goodie/roman/roman.js
gargaml: Done.

share/goodie/roman/roman.css
gargaml: I didn't know this unique class. Done :-)

pjhampton:
Hey @gargaml ! Thanks for this PR 👍  It's awesome so far. 

Two things:

1) Passing empty strings creates for odd UI behaviour. I would consider this over triggering. See my comments in the *.t file.
screen shot 2017-08-16 at 7 29 58 pm

2. The portrait mobile UI is slightly off. Can you please look into improving this, maybe taking inspiration from the [Tips UI](https://duckduckgo.com/?q=tip+calculator&atb=v1&ia=answer)

screen shot 2017-08-16 at 7 26 17 pm

t/Roman.t
```diff
@@ -10,32 +10,46 @@ zci answer_type => 'roman_numeral_conversion';
...
-    '20 to roman numerals' => build_test('XX (roman numeral conversion)', '20', 'XX'),
+    'roman 15' => build_test('arabic', '15', 'roman', 'XV'),
+    "roman xii" => build_test('roman', 'XII', 'arabic', '12'),
+    "roman mmcml" => build_test('roman', 'MMCML', 'arabic', '2950'),
+    "roman 2344" => build_test('arabic', '2344', 'roman', 'MMCCCXLIV'),
+    "arabic cccxlvi" => build_test('roman', 'CCCXLVI', 'arabic', '346'),
+    'roman MCCCXXXVII' => build_test('roman', 'MCCCXXXVII', 'arabic','1337'),
+    'roman numeral 10' => build_test('arabic', '10', 'roman', 'X'),
+    'roman numeral XVI' => build_test('roman', 'XVI', 'arabic', '16'),
+    'roman 1337' => build_test('arabic', '1337', 'roman','MCCCXXXVII'),
```
pjhampton: The term `arabic man` returns a result. I think you need to do a bit more aggressive validation:

screen shot 2017-08-16 at 7 37 48 pm

share/goodie/roman/roman.js
```diff
@@ -0,0 +1,227 @@
...
+            }
+            
+            previousComponent = component;
+            arabicValue -= component * leftDigit;
+        });
+        
+        return romanDigits;
+    }
+    
+    function upperCaseFirstLetter (string) {
```
pjhampton: You can use `DDG.capitalize` instead of that. 
See the js utilities available to you 😄  https://github.com/duckduckgo/duckduckgo-utils/blob/master/util.js#L42

share/goodie/roman/roman.js
```diff
@@ -0,0 +1,227 @@
...
+     * component in the roman notation (V, L, D).
+     * The remaining cases are straightforward.
+     */
+    
+    function arabicToRoman(arabic) {
+        var arabicValue = parseInt(arabic);
+        var previousComponent;
+        var romanDigits = '';
+        
+        [1000, 100, 10, 1].forEach(function (component) {
```
pjhampton: I don't think we can use this just yet because it is es6? I might be wrong but not all browsers will support this.

kirkins: ES5 to be exact, but it doesn't work on IE8.

t/Roman.t
```diff
@@ -10,32 +10,46 @@ zci answer_type => 'roman_numeral_conversion';
...
+    '10 into roman' => build_test('arabic', '10', 'roman', 'X'),
+    '3999 in roman numeral' => build_test('arabic', '3999', 'roman', 'MMMCMXCIX'),
+    'xiii to arabic' => build_test('roman', 'XIII', 'arabic', '13'),
+    'CXIII to arabic' => build_test('roman', 'CXIII', 'arabic', '113'),
+    'convert XXX into arabic' => build_test('roman', 'XXX', 'arabic', '30'),
+    'convert MMCC to arabic' => build_test('roman', 'MMCC', 'arabic', '2200'),
+    'convert 15 to roman' => build_test('arabic', '15', 'roman', 'XV'),
+    'convert 30 into roman' => build_test('arabic', '30', 'roman', 'XXX'),
+    'convert to arabic' => build_test('roman', '', 'arabic', ''),
+    'foo to arabic numerals' => build_test('roman', '', 'arabic', '')
```
gargaml: I hesitated with this one, fixed :-)

t/Roman.t
gargaml: Indeed this can lead to weird behavior… Now, I check that the input is a valid roman number.

share/goodie/roman/roman.js
gargaml: Thank you, I'll dig into this!

share/goodie/roman/roman.js
gargaml: I guess we want to support IE8, I'll use a for in loop instead (I'll just make sure that this is compatible).

gargaml:
I removed the equal sign because with the vertical alignment on narrow screens, it doesn't make sense.

I will add a warning if the user input goes out of range later.

t/Roman.t
```diff
@@ -10,32 +10,46 @@ zci answer_type => 'roman_numeral_conversion';
...
+    '10 into roman' => build_test('arabic', '10', 'roman', 'X'),
+    '3999 in roman numeral' => build_test('arabic', '3999', 'roman', 'MMMCMXCIX'),
+    'xiii to arabic' => build_test('roman', 'XIII', 'arabic', '13'),
+    'CXIII to arabic' => build_test('roman', 'CXIII', 'arabic', '113'),
+    'convert XXX into arabic' => build_test('roman', 'XXX', 'arabic', '30'),
+    'convert MMCC to arabic' => build_test('roman', 'MMCC', 'arabic', '2200'),
+    'convert 15 to roman' => build_test('arabic', '15', 'roman', 'XV'),
+    'convert 30 into roman' => build_test('arabic', '30', 'roman', 'XXX'),
+    'convert to arabic' => build_test('roman', '', 'arabic', ''),
+    'foo to arabic numerals' => undef
```
mintsoft: can you add a test for the case that pj pointed out? `"arabic man" => undef`

gargaml:
@mintsoft New test added.

mintsoft:
@gargaml Currently this returns the calculator on the query `roman` and `arabic` I don't think it should show for that. It probably shouldn't display unless the user has specified some input values in either roman numerals or arabic numbers.

gargaml:
Some information about inputs and warnings have been added.

lib/DDG/Goodie/Roman.pm
```diff
@@ -5,35 +5,81 @@ use strict;
...
-    } elsif ($in =~ /^[mdclxvi]+$/i) {
-        $in  = uc($in);
-        $out = arabic($in);
+    # By default, we convert from roman to arabic.
+    my $input = 'roman';
+    my $input_value = '';
+    my $output = 'arabic';
+    my $output_value = '';
+    
+    if (any { $query =~ $_ } @roman_to_arabic) {
```
moollaza: I would remove this, and quality the `return` in the else with this:

```perl
else {
 return unless any { $query =~ $_ } @roman_to_arabic;
}
```

IMO the empty `if` block looks really...odd

moollaza:
@gargaml Let's replace `converter_message` with static text in the `subtitle` (set in the Perl). Also, change `subtitle_content` to `title_content`.

Other changes:

- Remove margin-left for left hand side (to align content with left edge of search box)
  - On mobile there should be no margin-left
- Replace margin-top with margin-bottom, to add spacing between content and new subtitle
  - On mobile you'll need margin-top for the right hand side
- Add 1em margin-right to left side for desktop to space out the inputs more

Final result should look more like this:
roman_numeral_50_at_duckduckgo

Subtitle could also be, `Accepts inputs from 1 - 3999, I - MMMCMXCIX`

share/goodie/roman/content.handlebars
```diff
@@ -0,0 +1,16 @@
+<form class="converter">
+  <div class="converter__input">
+    <label class="converter__input__label">Roman</label>
+    <input type="text" class="converter__input__field" />
+  </div>
+  <div class="converter__output">
+    <label class="converter__output__label">Arabic</label>
+    <input type="text" class="converter__output__field" />
```
moollaza: What about `type="number"`?

share/goodie/roman/content.handlebars
```diff
@@ -0,0 +1,16 @@
...
+  </div>
+  <div class="converter__messages">
```
moollaza: Please remove, and put into a `subtitle`

t/Roman.t
```diff
@@ -10,32 +10,47 @@ zci answer_type => 'roman_numeral_conversion';
...
-    '11 in roman numerals' => build_test('XI (roman numeral conversion)', '11', 'XI'),
-    'xiii to arabic' => build_test('13 (roman numeral conversion)', 'XIII', '13'),
-    '20 to roman numerals' => build_test('XX (roman numeral conversion)', '20', 'XX'),
+    'roman 15' => build_test('arabic', '15', 'roman', 'XV'),
+    "roman xii" => build_test('roman', 'XII', 'arabic', '12'),
+    "roman mmcml" => build_test('roman', 'MMCML', 'arabic', '2950'),
+    "roman 2344" => build_test('arabic', '2344', 'roman', 'MMCCCXLIV'),
+    "arabic cccxlvi" => build_test('roman', 'CCCXLVI', 'arabic', '346'),
+    'roman MCCCXXXVII' => build_test('roman', 'MCCCXXXVII', 'arabic','1337'),
+    'roman numeral 10' => build_test('arabic', '10', 'roman', 'X'),
```
moollaza: how about `roman numeral for 10`?

lib/DDG/Goodie/Roman.pm
```diff
@@ -5,35 +5,81 @@ use strict;
...
+            input => $input,
+            input_value => $input_value,
+            output => $output,
+            output_value => $output_value
         },
         templates => {
-            group => 'text'
+            group => 'text',
+            options => {
+                subtitle_content => 'DDH.roman.content'
```
moollaza: Let's change this to `title_content`

share/goodie/roman/roman.css
```diff
@@ -0,0 +1,25 @@
+.zci--roman .converter {
+    width: 100%;
+}
+
+.zci--roman .converter__input, .zci--roman .converter__output {
```
moollaza: nit: we prefer to put each selector on a newline

```css
zci--roman .converter__input,
.zci--roman .converter__output {
   ...
}
```

share/goodie/roman/content.handlebars
```diff
@@ -0,0 +1,16 @@
+<form class="converter">
+  <div class="converter__input">
```
moollaza: please indent with 4 spaces

share/goodie/roman/content.handlebars
gargaml: If I add this restriction, I have to do it in the Javascript because the input content can vary depending on the order of the conversion.

moollaza: Fair enough. You could easily pass along some variables to build this UI as needed.

For example:

```js
var inputLabel = 'Roman'
var outputLabel = 'Arabic'
var inputType = 'text'
var outputType = 'number'
if (convertingFromArabic) {
    inputLabel = 'Arabic';
    outputLabel = 'Arabic';
    inputType = 'number';
    outputType = 'text'
}
```

```html
<div class="converter__output">
    <label class="converter__output__label">{{outputLabel}}</label>
    <input type="{{outputType}}" class="converter__output__field" />
    ...
```

moollaza: The other benefit here is that we generated the UI correctly the first time, instead of modifying the DOM after we've created it, in the case that it's an Arabic -> Roman conversion

moollaza:
@gargaml Thanks! This LGTM :+1:

I tweaked the UI slightly.

PR #4428, opened 2017-08-12
Update cheat sheets of tmux

TiagoDanin:

Added:
- Close current window
- Previous/Next window
- Delete session

---

Instant Answer Page: https://duck.co/ia/view/tmux_cheat_sheet

### DIFF
--- a/share/goodie/cheat_sheets/json/tmux.json
+++ b/share/goodie/cheat_sheets/json/tmux.json
@@ -70,9 +70,6 @@
         }, {
             "val": "Next window",
             "key": "[Ctrl] [b], [n]"
-        }, {
-            "val": "Search for text in windows",
-            "key": "[Ctrl] [b], [f]"
         }],
         "Pane Control": [{
             "val": "Split pane horizontally",
--- a/share/goodie/cheat_sheets/json/tmux.json
+++ b/share/goodie/cheat_sheets/json/tmux.json
@@ -66,10 +66,10 @@
             "key": "[Ctrl] [b], [w]"
         }, {
             "val": "Previous window",
-            "key": "[Ctrl] [b], [{]"
+            "key": "[Ctrl] [b], [\\{]"
         }, {
             "val": "Next window",
-            "key": "[Ctrl] [b], [}]"
+            "key": "[Ctrl] [b], [\\}]"
         }],
         "Pane Control": [{
             "val": "Split pane horizontally",
--- a/share/goodie/cheat_sheets/json/tmux.json
+++ b/share/goodie/cheat_sheets/json/tmux.json
@@ -66,10 +66,13 @@
             "key": "[Ctrl] [b], [w]"
         }, {
             "val": "Previous window",
-            "key": "[Ctrl] [b], [\\{]"
+            "key": "[Ctrl] [b], [p]"
         }, {
             "val": "Next window",
-            "key": "[Ctrl] [b], [\\}]"
+            "key": "[Ctrl] [b], [n]"
+        }, {
+            "val": "Find window",
+            "key": "[Ctrl] [f]"
         }],
         "Pane Control": [{
             "val": "Split pane horizontally",
@@ -84,11 +87,17 @@
             "val": "Previous pane",
             "key": "[Ctrl] [b], [;]"
         }, {
+            "val": "Show pane numbers",
+            "key": "[Ctrl] [b], [q]"
+        }, {
+            "val": "Toggle pane zoom",
+            "key": "[Ctrl] [b], [z]"
+        }, {
+            "val": "Convert pane into a window",
+            "key": "[Ctrl] [b], [!]"
+        }, {
             "val": "Kill current pane",
             "key": "[Ctrl] [b], [x]"
-        }, {
-            "val": "Kill all panes but the current one",
-            "key": "[Ctrl] [b], [!]"
         }, {
             "val": "Swap panes",
             "key": "[Ctrl] [b], [Ctrl] [O]"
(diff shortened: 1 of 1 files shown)


daxtheduck:
Metadata placeholder post
            
---
            
**How to find your IA page**
- New Instant Answer
              
- Existing Instant Answer

share/goodie/cheat_sheets/json/tmux.json
```diff
@@ -56,11 +56,20 @@
...
             "key": "[Ctrl] [b], [,]"
         }, {
+            "val": "Close current window",
+            "key": "[Ctrl] [b], [&]"
+        }, {
             "val": "List windows",
             "key": "[Ctrl] [b], [w]"
+        }, {
+            "val": "Previous window",
+            "key": "[Ctrl] [b], [\\{]"
```
charles-l: The man page for tmux 2.5 (which I'm currently running) has the brackets mapped to different comments:

           {           Swap the current pane with the previous pane.
           }           Swap the current pane with the next pane.

`<leader>n` and `<leader>p` are the bindings I've seen before. Did this change in a different version of tmux?

TiagoDanin: Oh, Work only in my config Tmux. Updated

share/goodie/cheat_sheets/json/tmux.json
```diff
@@ -56,11 +56,23 @@
...
             "key": "[Ctrl] [b], [w]"
+        }, {
+            "val": "Previous window",
+            "key": "[Ctrl] [b], [p]"
+        }, {
+            "val": "Next window",
+            "key": "[Ctrl] [b], [n]"
+        }, {
+            "val": "Find window",
+            "key": "[Ctrl] [f]"
```
sahildua2305: @TiagoDanin If I understand it correctly, `[Ctrl] [f]` is for searching a text in open windows. Is that right? If so, we can change the text to reflect that better. 🙂

share/goodie/cheat_sheets/json/tmux.json
```diff
@@ -139,6 +166,9 @@
         }, {
             "val": "Re-attach a detached session (and detach it from elsewhere)",
             "key": "tmux attach -d"
+        }, {
+            "val": "Delete session",
+            "key": "tmux kill-session -t"
```
sahildua2305: According to the source of this cheat sheet, `tmux kill-session -t` should be `tmux kill-session -t <target-session>`. Can you please check it once? 🙂

kirkins: @sahildua2305 if he adds `<target-session>` here I think he would also need to add it in some places that aren't in the original to keep consistency, for example `tmux attach -d <target-session>`

sahildua2305: @kirkins Good point again! 😄 

Personally, I'd prefer having `<target-session>` in all of the commands where required. My point is - that will make the cheat sheet more useful for the beginners as well. What do you think?

kirkins: @sahildua2305 yes, that would be ideal.

sahildua2305: @TiagoDanin would you like to make the corresponding change in this PR or a different one? Both will be fine with me.

share/goodie/cheat_sheets/json/tmux.json
charles-l: Hmm... I'm not seeing this binding in the default tmux config either. The man page (http://man.openbsd.org/cgi-bin/man.cgi/OpenBSD-current/man1/tmux.1) doesn't mention a default C-f binding and when I tested a default tmux config C-f didn't do anything.

kirkins: It's defentily a shortcut as mentioned in the man file

```
     find-window [-CNT] [-F format] [-t target-window] match-string
                   (alias: findw)
             Search for the fnmatch(3) pattern match-string in window names, titles, and visible content (but not history).  The flags control
             matching behavior: -C matches only visible window contents, -N matches only the window name and -T matches only the window title.
             The default is -CNT.  If only one window is matched, it'll be automatically selected, otherwise a choice list is shown.  For the
             meaning of the -F flag, see the FORMATS section.  This command works only if at least one client is attached.

```

But it is listed incorrectly here. You need to use:

    Ctrl-b f <search term>

https://wiki.archlinux.org/index.php/Tmux#Key_bindings

kirkins: _P.S. Normally window names are assigned automatically but you can set custom window names._ 

_I think this is where using the search window tool would come into play._

kirkins: @TiagoDanin can we also add the `Ctrl-b w` command listed on the [archlinux link](https://wiki.archlinux.org/index.php/Tmux#Key_bindings)?

This seems like it would be more useful than `Ctrl-b f` though it doesn't seem to work as I would expect. 

I open 4 panes and then use `Ctrl-b w` and I only see one option not 4.

_Maybe I'm confusing tmux panes and tmux windows(?)_

kirkins: Correction `Ctrl-b w` is already listed. My mistake.

TiagoDanin: @kirkins panel and window are different

charles-l: > But it is listed incorrectly here. You need to use:
> `Ctrl-b f <search term>`

Ah, good call - I missed that.

> Maybe I'm confusing tmux panes and tmux windows(?)

Yeah, tmux's naming convention is a bit odd. The hierarchy is:

    sessions > windows (which are like tabs) > panes (which are splits in a window)

Sessions and windows get automatic names by default but can be manually named later.

`Ctrl-b f` will search through windows, though I've never had enough windows that I've had to search for them (`Ctrl-b w` has always been good enough for me). I think it's more useful for scripting tmux?

TiagoDanin:
wasn't ready. (deleting commit 601049b)

bsstoner:
This was released to production 🚀

PR #4502, opened 2017-09-23
conversions(added connecting words)

hektr:
## Description of new Instant Answer, or changes

Now connecting words can start with "are/is".

## Related Issues and Discussions

Fixes #4501 

## People to notify

---

Instant Answer Page: https://duck.co/ia/view/conversions

PR #4506, opened 2017-09-23
Conversions fix

### LINKED ISSUE #4501: Conversions: 'how many x are in y' doesn't trigger, bit 'how many x in y' does
_Reported via User Feedback_

Example: https://duckduckgo.com/?q=how+many+tsp+are+in+1+tbsp

------
IA Page:  http://duck.co/ia/view/conversions 
[Maintainer](http://docs.duckduckhack.com/maintaining/guidelines.html): @pjhampton

hektr:
**Conversions**

now conversions IA can answer at questions, composed not only like this "how many metres [are] in a mile?" but also like this "how many metres are there in a mile?" or "how many metres are in 2 miles?".

Fixes #4502 Fixed #4501 

## People to notify

---

### DIFF
--- a/t/Conversions.t
+++ b/t/Conversions.t
@@ -695,6 +695,22 @@ ddg_goodie_test(
             physical_quantity => 'length'
         })
     ),
+    'how many metres are in 10 yards?' => test_zci(
+        '', structured_answer => make_answer({
+            raw_input => '10',
+            from_unit => 'yard',
+            to_unit => 'meter',
+            physical_quantity => 'length'
+        })
+    ),
+    'how many metres are there in a mile?' => test_zci(
+        '', structured_answer => make_answer({
+            raw_input => '1',
+            from_unit => 'mile',
+            to_unit => 'meter',
+            physical_quantity => 'length'
+        })
+    ),
     'how many cm in a metre?' => test_zci(
         '', structured_answer => make_answer({
             raw_input => '1',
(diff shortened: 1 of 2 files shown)


daxtheduck:
Metadata placeholder post
            
---
            
**How to find your IA page**
- New Instant Answer
              
- Existing Instant Answer

t/Conversions.t
```diff
@@ -695,6 +695,22 @@ ddg_goodie_test(
             physical_quantity => 'length'
         })
     ),
+    'how many metres are in 10 yards?' => test_zci(
```
pjhampton: I think the original problem is that they do already trigger on `are in`, but not `in`. This test should be `how many metres in 10 yards?` 
You can get rid of the test below and a couple more like this that aren't metres / yards.

hektr: Is this question good ? `how many kilobytes are in 2Tb?` That regexp doesn't match it,  because prefix-left_num `are in` is not followed by `a`. You can only write something like `how many bytes **are in a** kilobyte?` but not `how many bytes are in kilobyte?` If you omit "a"  after `are in` in your request, regexp doesn't match and nothing triggers. But you can write this `how many tsp are in a 1 tbsp` and it will work. So, you can just add ?-quatifier after the corresponding `a` letter at regexp and it will give chance to people with bad english to know, how many bytes are in kilobyte or in 2 kilobytes.

PR #4526, opened 2017-10-25
Text Converter: Fix binary conversion bug.

### LINKED ISSUE #4513: Text Converter:  Ignore Binary to Decimal conversions greater than 11111111
I noticed that binary to decimal converter works fine only till 11111111, maybe it's a normal behavior, but it would be great to warn users instead of printing some strange values.

------

IA Page:  http://duck.co/ia/view/text_converter 
[Maintainer](http://docs.duckduckhack.com/maintaining/guidelines.html): @pjhampton

### LINKED ISSUE #4566: Text Converter:  Incorrect results for decimal to binary
I entered 1056760 and it gave me 1111111111000.  That is INCORRECT!  

The correct result is 100000001111111111000.

Using the incorrect result caused me some headaches, until I finally figured it out. 

I think this is an extremely serious error that could impact others as well.

I suggest you remove the bogus converter until it is repaired.   

------
IA Page:  http://duck.co/ia/view/text_converter 
[Maintainer](http://docs.duckduckhack.com/maintaining/guidelines.html): @pjhampton

### LINKED ISSUE #4567: Text Converter: Incorrect results for binary to hexadecimal
`101101` is converted to hex as `b1` (which is most likely because it is grouped as `1011` `01`). The numeric equivalent of that number is `2d` in hex (where the grouping should be done from right to left `10` `1101`, or left pad with two zeros before grouping from left).

------
IA Page:  http://duck.co/ia/view/text_converter 
[Maintainer](http://docs.duckduckhack.com/maintaining/guidelines.html): @pjhampton

### LINKED ISSUE #4570: Text Converter: incorrect conversion binary to decimal
OK (dec to bin):
 -2       ---->       1111111111111110 

UNEXPECTED (bin to dec):

1111111111111110      --->         255 254

------
IA Page:  http://duck.co/ia/view/text_converter 
[Maintainer](http://docs.duckduckhack.com/maintaining/guidelines.html): @pjhampton

vonaka:

Fixes #4513, fixes #4567, fixes #4570 and fixes #4566. For binary-to-decimal conversion we don't need to convert each byte separately. For binary-to-hexadecimal we need to go from the last byte to the first one.

## People to notify

---

Instant Answer Page: https://duck.co/ia/view/text_converter

### DIFF
--- a/share/goodie/text_converter/text_converter.js
+++ b/share/goodie/text_converter/text_converter.js
@@ -10,7 +10,7 @@ DDH.text_converter = DDH.text_converter || {};
         $convert_selects,
         $convert_from_select,
         $convert_to_select,
-        $convert_from_textarea, 
+        $convert_from_textarea,
         $convert_to_textarea;
 
     // TextConverter: A singleton object that performs all the text conversions
@@ -29,7 +29,10 @@ DDH.text_converter = DDH.text_converter || {};
             var to_type = $convert_to_select.val();
             var from = $convert_from_textarea.val();
 
-            if(from_type === "binary" && to_type === "hexadecimal") {
+            if(from_type === "decimal" && to_type === "binary") {
+                return parseInt(from, 10).toString(2);
+
+            } else if(from_type === "binary" && to_type === "hexadecimal") {
                 return TextConverter.binaryToHex(from);
 
             } else if(from_type === "binary" && to_type === "base64") {
--- a/share/goodie/text_converter/text_converter.js
+++ b/share/goodie/text_converter/text_converter.js
@@ -129,14 +129,8 @@ DDH.text_converter = DDH.text_converter || {};
         },
 
         binaryToDecimal: function(binary) {
-            var octet = binary.replace(/\s/g, "").match(/.{1,8}/g);
-            var dec_cache = [];
-
-            for(var i = 0; i < octet.length; i++) {
-                dec_cache.push(parseInt(octet[i], 2));
-            }
-
-            return dec_cache.join(" ");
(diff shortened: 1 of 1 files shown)


daxtheduck:
Metadata placeholder post
            
---
            
**How to find your IA page**
- New Instant Answer
              
- Existing Instant Answer

moollaza:
@pjhampton care to review?

@vonaka can you please provide some test cases?

vonaka:
@moollaza, sorry, but I'm not sure I understand, there is no such query for Text Converter that will bring some real value as result, I mean, one can't write _"100 binary to hex translator"_ for example ( _"100 binary to hex"_ brings us to Base Conversion ), but we can write _"binary to hex translator"_ and manually put 100 as 'argument', so I can't add any useful test, I guess. But once again, maybe I just don't understand something.

lolo101:
Should also fix #4570

vonaka:
@pjhampton, @moollaza, I added (lazy) fix for #4566. But I think it's a good idea to completely remove this strange and buggy Text Converter.

pjhampton:
It's [offline](https://duck.co/ia/view/text_converter/json). I don't think it needs to be purged (from recollection, it was quite popular and what makes DDG great is being able to serve niche queries). But I don't think I would be comfortable it going live again unless there was a full unit test suite back-fitted. 

Hopefully I will get the time required at some stage to nurse this IA back to health.

moollaza:
Deploying to Beta...

moollaza:
@vonaka Thanks! This LGTM :+1:

PR #4578, opened 2018-01-28
text "json" added in triggers for json validator

Rafi993:
jsonvalidator:  text "json" added to trigger

It adds json to triggers so that users get json validator even when they type json

## People to notify
@sahildua2305 , @tgckpg , @gautamkrishnar

---

Instant Answer Page: https://duck.co/ia/view/{ID}


moollaza:
Thanks for the PR. DuckDuckHack is in maintenance mode and during this time we are only accepting PRs for critical bug fixes.

This will be marked as tolerated and closed.

